PR #66, opened 2020-05-22
Refactor Image::MaxColor to reduce computation time

### LINKED ISSUE #65: Image::MaxColor scales very poorly with image size
While debugging why large Image heightmaps are so slow to load in osrf/gazebo, I noticed that the loader was spending lots of time in `Image::MaxColor` (via Heightmap::MaxElevation). I've added a performance test in [branch scpeters:image_max_color_performance](https://github.com/ignitionrobotics/ign-common/compare/ign-common3...scpeters:image_max_color_performance?expand=1), with square images of increasing size 2^N+1 x  2^N+1 (129x129, 257x257, 513x513, 1025x1025) added in https://github.com/ignitionrobotics/ign-common/commit/02ffa909169b34fda32115e041afe3bb4b7ed9b2 and a performance test added in d3eb55b. The images increase in size by roughly a factor of 4, but the time spent calling `Image::MaxColor` in the performance test increases by factors greater than 10 in my first tests on my mac laptop:

~~~
[----------] 4 tests from FlatHeightmaps/ImagePerformanceTest
[ RUN      ] FlatHeightmaps/ImagePerformanceTest.MaxColorFlatHeightmap/0
Error opening log file: /Users/scpeters/tmp-ign//test_logs/FlatHeightmaps/ImagePerformanceTest_MaxColorFlatHeightmap/0.log
[       OK ] FlatHeightmaps/ImagePerformanceTest.MaxColorFlatHeightmap/0 (261 ms)
[ RUN      ] FlatHeightmaps/ImagePerformanceTest.MaxColorFlatHeightmap/1
Error opening log file: /Users/scpeters/tmp-ign//test_logs/FlatHeightmaps/ImagePerformanceTest_MaxColorFlatHeightmap/1.log
[       OK ] FlatHeightmaps/ImagePerformanceTest.MaxColorFlatHeightmap/1 (3551 ms)
[ RUN      ] FlatHeightmaps/ImagePerformanceTest.MaxColorFlatHeightmap/2
Error opening log file: /Users/scpeters/tmp-ign//test_logs/FlatHeightmaps/ImagePerformanceTest_MaxColorFlatHeightmap/2.log
[       OK ] FlatHeightmaps/ImagePerformanceTest.MaxColorFlatHeightmap/2 (49867 ms)
[ RUN      ] FlatHeightmaps/ImagePerformanceTest.MaxColorFlatHeightmap/3
Error opening log file: /Users/scpeters/tmp-ign//test_logs/FlatHeightmaps/ImagePerformanceTest_MaxColorFlatHeightmap/3.log
[       OK ] FlatHeightmaps/ImagePerformanceTest.MaxColorFlatHeightmap/3 (764936 ms)
[----------] 4 tests from FlatHeightmaps/ImagePerformanceTest (818616 ms total)
~~~

* 129x129: 261 ms
* 257x257: 3551 ms
* 513x513: 49867 ms
* 1025x1025: 764936 ms

We should be able to speed this up significantly, which will be helpful for reducing the time spent loading Heightmaps.

scpeters:
This refactors `Image::MaxColor` to massively improve performance for large images, which is a significant bottleneck when loading ImageHeightmaps, as documented in #65. Several all-black images of different sizes (that happen to be suitable for heightmaps) were added in 02ffa90 and a performance test was added to `UNIT_Image_TEST` in d3eb55b. In b8e0a51, I copied the implementation of `Image::Pixel(...)` into `Image::MaxColor` to reduce the allocation and deallocation of FreeImage structs, which @mjcarroll had suggested may be the issue in https://github.com/ignitionrobotics/ign-common/issues/65#issuecomment-632082205. This refactor completely resolves the performance issues for me, though it could perhaps be done in a more elegant way with less duplicated code.

Also, it wasn't my intent to put a performance test inside the unit test, but I couldn't quite figure out how to make a performance test that depends on a component library (`graphics` in this case). If I put the test in `graphics/test/performance`, then it's tricky to know what is the target name of the component library. If I put it in `test/performance`, then I need some logic to know if that component is being built, so I just put it in the unit test.

Fixes #65.


scpeters:
I noticed a gcc compiler warning, and I feel the cleanest fix is to update to gtest 1.10.0: https://github.com/ignitionrobotics/ign-common/pull/67

mjcarroll:
Since the test images are completely blank, do you think it would make sense to just have a generation function for them?

scpeters:
> Since the test images are completely blank, do you think it would make sense to just have a generation function for them?

That's a good point. The issue is that `Image` has only [one constructor that takes a file name as input](https://github.com/ignitionrobotics/ign-common/blob/ign-common3/graphics/include/ignition/common/Image.hh#L98). Are you suggesting that we generate `png` files of different sizes dynamically in the test? I'm not sure how to do that.

mjcarroll:
>  The issue is that Image has only one constructor that takes a file name as input.

Ah, so I thought that we had the capability to construct one in-memory without touching the filesystem.  In that case, I think that what you have now is suitable.

scpeters:
I just rebased and added a changelog entry for #71

mjcarroll:
In the future, I think that our preference is to do squash-merge (which isn't the default for some reason)

PR #67, opened 2020-05-22
Update to gtest 1.10.0 with patch for gcc warnings

scpeters:
I noticed a gcc warning about `ISO C++11 requires at least one argument for the "..." in a variadic macro` in #66 related to the use of the gtest macro `INSTANTIATE_TEST_CASE_P`:

* https://build.osrfoundation.org/job/ignition_common-ci-pr_any-ubuntu_auto-amd64/321/warnings22Result/

We've faced this before and have added an extra `,` and `// NOLINT`:

* https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/pull-requests/3168/page/1#comment-130457920

That's not entirely satisfying, and I noticed that this has been fixed in gtest 1.10.0, which renames the `INSTANTIATE_TEST_CASE_P` macro to `INSTANTIATE_TEST_SUITE_P`:

* https://github.com/google/googletest/pull/2063
* https://github.com/google/googletest/releases/tag/release-1.10.0

So this PR updates the fork to gtest 1.10.0. It includes a patch in 61c4cb0 to suppress a separate gcc warning.


mjcarroll:
Retroactive approval, but just a note that we actually package `gtest` in almost all of the ignition repositories, do they need to be updated as well?

scpeters:
> Retroactive approval, but just a note that we actually package gtest in almost all of the ignition repositories, do they need to be updated as well?

this was driven by the use of `INSTANTIATE_TEST_CASE_P`, which is used in ign-physics and osrf/gazebo for sure, but I'm not sure about other ignition packages

so they certainly could all be upgraded, but I was just going to prioritize the ones using that macro

PR #70, opened 2020-06-03
More file path APIs

chapulina:
Add a couple of functions that I'm using to tackle https://github.com/ignitionrobotics/ign-gazebo/issues/123 : 

* `common::SystemPaths::FilePathEnv` - not to make assumptions about the current value, since we can't be sure `common::SystemPaths::SetFilePathEnv` hasn't been called.
* `common::systemPaths` - we've been exposing function by function of the global instance of `SystemPaths` though `Utils`. Instead, just expose the pointer so the functions can be called directly.


include/ignition/common/SystemPaths.hh
```diff
@@ -109,8 +109,17 @@ namespace ignition
...
       /// delimited paths. These paths will be used with the FindFile function.
       /// \param [in] _env name of the environment variable
+      /// \sa FilePathEnv
       public: void SetFilePathEnv(const std::string &_env);
 
+      /// \brief Get the file path environment variable in use.
+      /// The environment variable contains a set of colon (semicolon on
+      /// windows) delimited paths. These paths are used with the FindFile
+      /// functions.
+      /// \retuen Name of the environment variable
```
mjcarroll, line 119:
```suggestion
      /// \return Name of the environment variable
```

include/ignition/common/Util.hh
```diff
@@ -160,6 +161,12 @@ namespace ignition
     void IGNITION_COMMON_VISIBLE addFindFileURICallback(
         std::function<std::string(const URI &)> _cb);
 
+    /// \brief Get a pointer to the global system paths that's used by all
```
mjcarroll, line 164:
```suggestion
    /// \brief Get a pointer to the global system paths that is used by all
```

include/ignition/common/Util.hh
```diff
@@ -160,6 +161,12 @@ namespace ignition
...
+    /// the findFile functions.
+    /// Caller should not assume ownership.
```
mjcarroll, line 166: Are there any other warnings that should go along with this?  Does this mean that the caller can mutate the global system paths and impact other users?

chapulina: The idea is to give the caller the ability to do that. The impact will depend on who's using the global system paths though. The `MeshManager` happens to use it. Whoever calls `common::findFile` will use it.But there are many places where custom `SystemPaths` objects are used, such as when `ign-gazebo` loads worlds,

mjcarroll: I agree, that's a bit difficult to document.  I think just enough to make sure that if people start messing with paths orders that they would have a place to look. Maybe something like:

```
The returned instance has global shared state for a given process.  Care should be taken when manipulating global system paths.
```

chapulina: Thanks, added it here: b4aa1bce44de23b63d54b3186407eb9d81ab4c4d

Issue #77, opened 2020-06-16
Support hierarchical submeshes in ColladaLoader


nkoenig:
It's possible to structure submeshes in a hierarchy. For example:

* BodyFront
    * WheelFR
    * WheelFL
* BodyBack
    * WheelBL
    * WheelBR

If the `BodyFront` submesh is used, then nothing appears. If the `WheelFR` submesh is used, then both the `WheelFR` and `BodyFront` meshes are shown.

Inside a collada file I see something similar to:

```
 <library_visual_scenes>
    <visual_scene id="k2" name="k2">
      <node name="BodyBack" id="BodyBack" sid="BodyBack">
        <matrix sid="matrix">1.000000 0.000000 0.000000 0.000000 0.000000 0.000000 -1.000000 0.158456 0.000000 1.000000 0.000000 6.812362 0.000000 0.000000 0.000000 1.000000</matrix>
        <instance_geometry url="#BodyBack-lib">
          <bind_material>
            <technique_common>
              <instance_material symbol="K2" target="#K2"/>
            </technique_common>
          </bind_material>
        </instance_geometry>
        <extra>
          <technique profile="FCOLLADA">
            <visibility>1.000000</visibility>
          </technique>
        </extra>
        <node name="WheelBL" id="WheelBL" sid="WheelBL">
          <matrix sid="matrix">1.000000 0.000000 0.000000 9.865527 0.000000 1.000000 0.000000 -2.609944 0.000000 0.000000 1.000000 -6.854113 0.000000 0.000000 0.000000 1.000000</matrix>
          <instance_geometry url="#WheelBL-lib">
            <bind_material>
              <technique_common>
                <instance_material symbol="K2" target="#K2"/>
              </technique_common>
            </bind_material>
          </instance_geometry>
          <extra>
            <technique profile="FCOLLADA">
              <visibility>1.000000</visibility>
            </technique>
          </extra>
        </node>
        <node name="WheelBR" id="WheelBR" sid="WheelBR">
          <matrix sid="matrix">1.000000 0.000000 0.000000 9.865527 0.000000 1.000000 0.000000 -2.609943 0.000000 0.000000 1.000000 7.564952 0.000000 0.000000 0.000000 1.000000</matrix>
          <instance_geometry url="#WheelBR-lib">
            <bind_material>
              <technique_common>
                <instance_material symbol="K2" target="#K2"/>
              </technique_common>
            </bind_material>
          </instance_geometry>
          <extra>
[...]

PR #78, opened 2020-06-16
Fix hierarchical submesh processing in the ColladaLoader.

nkoenig:
Fix for issue #77.

The problem is that the `this->currentNodeName` would have the child node name when process the parent's mesh information.

Signed-off-by: Nate Koenig <nate@openrobotics.org>


Changelog.md
```diff
@@ -2,6 +2,9 @@
 
 ## Ignition Common 3.X.X
 
+1. Fix hierarchical submesh processing in the ColladaLoader.
```
mjcarroll, line 5: Can you rename the PR to be consistent with this? That way the squash merge will be consistent.

nkoenig: Done.

iche033:
we just noticed that the changes here broke the [Polaris Ranger EV](https://app.ignitionrobotics.org/OpenRobotics/fuel/models/Polaris%20Ranger%20EV) model on fuel. @nkoenig  Do you have an example hierarchical model that we can test with?

PR #81, opened 2020-07-08
Fixed the AudioDecoder class

adlarkin:
I've updated the [unit tests](https://github.com/ignitionrobotics/ign-common/blob/audioDecoder_fix_adlarkin/av/src/AudioDecoder_TEST.cc) for the `AudioDecoder` class so that they compile and pass.

I also added a check in `AudioDecoder::SampleRate` to ensure that a null pointer is not de-referenced in case users call `AudioDecoder::SampleRate` before calling `AudioDecoder::SetFile`.

One question: In the case where users call `AudioDecoder::SampleRate` before calling `AudioDecoder::SetFile`, what should `AudioDecoder::SampleRate` return? I made `AudioDecoder::SampleRate` return `-1` in this case, because I figured that a negative sample rate implies that the decoder's current state is invalid. I anyone thinks there is a better value to return (for example, `0`), let me know, and I will update `AudioDecoder::SampleRate` accordingly.

### DIFF
--- a/av/include/ignition/common/AudioDecoder.hh
+++ b/av/include/ignition/common/AudioDecoder.hh
@@ -60,6 +60,7 @@ namespace ignition
 
       /// \brief Get the sample rate from the latest decoded file.
       /// \return Integer sample rate, such as 44100.
+      /// If no file is decoded, -1 is returned.
       public: int SampleRate();
 
       /// \brief Free audio object, close files, streams.
--- a/av/src/AudioDecoder.cc
+++ b/av/src/AudioDecoder.cc
@@ -172,7 +172,10 @@ bool AudioDecoder::Decode(uint8_t **_outBuffer, unsigned int *_outBufferSize)
 /////////////////////////////////////////////////
 int AudioDecoder::SampleRate()
 {
-  return this->data->codecCtx->sample_rate;
+  if (this->data->codecCtx)
+    return this->data->codecCtx->sample_rate;
+
+  return -1;
 }
 
 /////////////////////////////////////////////////
--- a/av/src/AudioDecoder_TEST.cc
+++ b/av/src/AudioDecoder_TEST.cc
@@ -59,7 +59,7 @@ TEST(AudioDecoder, DataBuffer)
 {
   common::AudioDecoder audio;
 
-  std::string path = PROJECT_SOURCE_PATH;
+  std::string path = TEST_PATH;
   path += "/data/cheer.wav";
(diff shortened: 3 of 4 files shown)


mjcarroll:
Rebase on ign-common3 and signoff for DCO please.

scpeters:
> One question: In the case where users call `AudioDecoder::SampleRate` before calling `AudioDecoder::SetFile`, what should `AudioDecoder::SampleRate` return? I made `AudioDecoder::SampleRate` return `-1` in this case, because I figured that a negative sample rate implies that the decoder's current state is invalid. I anyone thinks there is a better value to return (for example, `0`), let me know, and I will update `AudioDecoder::SampleRate` accordingly.

`-1` sounds reasonable to me, and you documented it in the header file, so it looks fine to me

iche033:
changes look good to me. Just need to rebase for abi-checker build to pass

mjcarroll:
@scpeters and @iche033  I'm trying to help Ashton understand the output of the ABI report here: 
https://build.osrfoundation.org/job/ignition_common-abichecker-any_to_any-ubuntu_auto-amd64/251/API_5fABI_20report/

Based on the changes in the PR, I wouldn't anticipate any API or ABI problems, because nothing along those lines has changed.

As far as I can tell the checker believes that essential the same symbol (mod template specialization) is being removed and added.  What is interesting is that this `cc` file wasn't part of the build until this pull request.  

If I look at my local copies of `libignition-common3`, that symbol doesn't exist:

```
$ nm -D /usr/lib/x86_64-linux-gnu/libignition-common3-av.so.3 | grep AudioDecoder
$ nm -D /usr/lib/x86_64-linux-gnu/libignition-common3.so.3 | grep AudioDecoder
```

scpeters:
> I'm trying to help Ashton understand the output of the ABI report here:
> https://build.osrfoundation.org/job/ignition_common-abichecker-any_to_any-ubuntu_auto-amd64/251/API_5fABI_20report/

that looks like a false positive, which I've seen before: https://github.com/ignition-tooling/release-tools/issues/120

no code in that header file is changed, so I would just ignore it. an administrator should override the blocked merge in this case

iche033:
oh I didn't know about the abi-checker bug. Not sure if it's an issue with the abi-checker when the base version contains only the header files and the .cc files were not being built.

adlarkin:
Thanks for following up on this, @scpeters and @mjcarroll. Is everything okay then, or are there any other changes I need to make?

PR #86, opened 2020-08-12
Add more detail to profiler tutorial

claireyywang:
Signed-off-by: claireyywang <22240514+claireyywang@users.noreply.github.com>


tutorials/profiler.md
```diff
@@ -73,6 +72,12 @@ cmake ..
 make profiler_example
 ```
 
+Or if you are building with ``colcon``, enable profiler by passing in ``cmake`` args.
```
chapulina, line 75: Is this necessary on this example even though `IGN_PROFILER_ENABLE` is already set on the CMake file above?

I think that your suggestion is important when compiling the Ignition libraries, because unlike this example, they have the profiler [disabled by default](https://github.com/ignitionrobotics/ign-gazebo/blob/8ec8d66cc8a6b661a2db6ccae153a785287fa315/CMakeLists.txt#L23-L29). How about explaining that instead? Something like:

```
If you want to use the profiler of an Ignition library, be sure to enable
the profiler at compile time with the `ENABLE_PROFILER` CMake
argument.

When using plain CMake:

`cmake .. -DENABLE_PROFILER=1`

When using colcon:

`colcon build --cmake-args -DENABLE_PROFILER=1`
```
I think this explanation could be outside of this example though.

tutorials/profiler.md
```diff
@@ -99,7 +104,18 @@ xdg-open /usr/share/ignition/ignition-common3/profiler_vis/index.html
 open /usr/share/ignition/ignition-common3/profiler_vis/index.html
 ```
 
-If you are successful, you should see the profiler output in a browser.
+If you see ``connection error``, there are a couple of things to double check
+1. Is profiler enabled?
```
chapulina, line 108: How about:

"Was the profiler enabled when the project you're trying to run was compiled? Note that this isn't the case if you installed Ignition libraries from binaries, for example. You need to compile the project from source with the `ENABLE_PROFILER` variable set."

ahcorde: I suggest to add a 4th point. You should configure the environment variable `RMT_PORT` with the right port number.

tutorials/profiler.md
```diff
@@ -99,7 +100,39 @@ xdg-open /usr/share/ignition/ignition-common3/profiler_vis/index.html
...
+
+If the profiler is run successfully, you should see output in a browser. Similar to this 
+
+![profiler viewer exmaple](imgs/profiler_tutorial_example.png)
+
+### TroubleShoot the web viewer
+
+If you see ``connection error``, there are a couple of things to double check
+1. Was the profiler enabled when the project you're trying to run was compiled? Note that this isn't the case if you installed Ignition libraries from binaries, for example. You need to compile the project from source with the ENABLE_PROFILER variable set.
+2. Are you using the correct port number in the upper left corner ``Connection Addresss: ws://127.0.0.1:1500/rmt``? Run ``ign gazebo -v 4`` or ``echo $RMT_PORT`` to show the port number in use. The port number will be printed out if the profiler is enabled.
```
iche033, line 124: minor note here that ```echo $RMT_PORT``` will not print out anything if the user has not set this env variable. I think an easy change is to just leave out `or ``echo $RMT_PORT``` text. Point 4 should be sufficient to cover this port env usage.

tutorials/profiler.md
```diff
@@ -99,7 +100,39 @@ xdg-open /usr/share/ignition/ignition-common3/profiler_vis/index.html
...
+When compiling with ``colcon``:
+```{.sh}
+colcon build --cmake-args -DENABLE_PROFILER=1
+```
+
+If the profiler is run successfully, you should see output in a browser. Similar to this 
+
+![profiler viewer exmaple](imgs/profiler_tutorial_example.png)
+
+### TroubleShoot the web viewer
```
iche033, line 120: `Troubleshoot`

tutorials/profiler.md
mjcarroll: It should be almost always unset.

PR #90, opened 2020-08-25
Deprecated common:Time

ahcorde:
This PR is part of this issue https://github.com/ignitionrobotics/ign-common/issues/61. The idea it's to deprecate the class `common::Time` in favor of `std::chrono`.

This two functions convert from `time_point` to seconds and nanoseconds and viceversa.

### DIFF
--- a/include/ignition/common/Time.hh
+++ b/include/ignition/common/Time.hh
@@ -76,7 +76,7 @@ namespace ignition
 
       /// \brief Get the wall time
       /// \return the current time
-      public: static const Time &SystemTime();
+      public: static const IGN_DEPRECATED(3) Time &SystemTime();
 
       /// \brief Set to sec and nsec
       /// \param[in] _sec Seconds
@@ -103,7 +103,7 @@ namespace ignition
       /// actual call to the system's sleep function.
       ///
       /// On Windows the return value is always common::Time::Zero.
-      public: static Time Sleep(const common::Time &_time);
+      public: static Time IGN_DEPRECATED(3) Sleep(const common::Time &_time);
 
       /// \brief Get the time as a string formatted as "DD hh:mm:ss.mmm", with
       /// the option to choose the start/end.
--- a/include/ignition/common/Time.hh
+++ b/include/ignition/common/Time.hh
@@ -52,31 +52,31 @@ namespace ignition
       };
 
       /// \brief Constructors
-      public: IGN_DEPRECATED(3) Time();
+      public: IGN_DEPRECATED(4) Time();
 
       /// \brief Copy constructor
       /// \param[in] time Time to copy
-      public: IGN_DEPRECATED(3) Time(const Time &_time);
+      public: IGN_DEPRECATED(4) Time(const Time &_time);
 
       /// \brief Constructor
       /// \param[in] _tv Time to initialize to
-      public: explicit IGN_DEPRECATED(3) Time(const struct timespec &_tv);
+      public: explicit IGN_DEPRECATED(4) Time(const struct timespec &_tv);
 
       /// \brief Constructor
       /// \param[in] _sec Seconds
       /// \param[in] _nsec Nanoseconds
-      public: IGN_DEPRECATED(3) Time(int32_t _sec, int32_t _nsec);
+      public: IGN_DEPRECATED(4) Time(int32_t _sec, int32_t _nsec);
 
       /// \brief Constuctor
       /// \param[in] _time Time in double format sec.nsec
-      public: explicit IGN_DEPRECATED(3) Time(double _time);
+      public: explicit IGN_DEPRECATED(4) Time(double _time);
 
       /// \brief Destructor
       public: virtual ~Time();
 
       /// \brief Get the wall time
       /// \return the current time
-      public: static const IGN_DEPRECATED(3) Time &SystemTime();
+      public: static const IGN_DEPRECATED(4) Time &SystemTime();
 
       /// \brief Set to sec and nsec
       /// \param[in] _sec Seconds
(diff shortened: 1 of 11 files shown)


include/ignition/common/Time.hh
```diff
@@ -52,24 +52,32 @@ namespace ignition
       };
 
       /// \brief Constructors
-      public: Time();
+      public: [[deprecated("Do not use this class, use std::chrono")]] Time();
```
chapulina, line 55: We shouldn't introduce deprecations in a released major. We can add the new functions here, but let's add the deprecations to `master`.

Also, please use `IGN_DEPRECATED(3)` instead of `[[deprecated]]`.

include/ignition/common/Time.hh
```diff
@@ -105,6 +113,15 @@ namespace ignition
       /// On Windows the return value is always common::Time::Zero.
       public: static Time Sleep(const common::Time &_time);
 
+      public: static void GetSecondsAndNanoseconds(
```
chapulina, line 116: This looks similar to [ignition::math::durationToSecNsec](https://github.com/ignitionrobotics/ign-math/blob/6a3c4a32b18e6bd7956fecfe53ccfe14bfcd2907/include/ignition/math/Helpers.hh#L573-L584). How about putting the new functions there instead of `ign-common`?

ahcorde:
- Remove two new methods (moved to ign-math).
 - Deprecated constructor and static functions.
 - Retarget branch to `master`

include/ignition/common/Time.hh
mjcarroll: Would that be an issue for people that were previously depending on ign-common for time-related functions?  Or would both generally be available everywhere

chapulina: Since `ign-math` is a required dependency of `ign-common3` anyway, I don't think current users should need to add any new dependencies.

include/ignition/common/Time.hh
iche033: `master` is version 4, so `IGN_DEPRECATED(4)`?

ahcorde:
@chapulina, This issue is the meta-ticket to follow all the libraries that currently are using `common::Time` and it points to the related PRs to replace the use of `common::Time` with `std::chrono`

ahcorde:
I fixed the `after_make.sh` script in this PR, but I think the right solution is in this other PR https://github.com/ignition-tooling/action-ignition-ci/pull/22

src/WorkerPool_TEST.cc
```diff
@@ -92,7 +92,14 @@ TEST(WorkerPool, WaitWithTimeout)
       {
         workSentinel = 5;
       });
+#ifndef _WIN32
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
   Time time(5);
```
chapulina, line 99: Instead of suppressing warnings, I think we should update this to use chrono. Otherwise, when the `common::Time` class is removed, how will this test work?

This means deprecating the `WaitForResults` function and adding a new one that takes chrono.

src/Timer.cc
```diff
@@ -20,11 +20,18 @@
 using namespace ignition;
 using namespace common;
 
+#ifndef _WIN32
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
 //////////////////////////////////////////////////
 Timer::Timer()
```
chapulina, line 28: I'm not so sure about deprecating this class, does chrono have a good substitute?

src/Timer.cc
```diff
@@ -52,15 +59,18 @@ bool Timer::Running() const
 }
 
 //////////////////////////////////////////////////
-Time Timer::Elapsed() const
+double Timer::Elapsed() const
```
chapulina, line 62: If `Timer` will be deprecated, we shouldn't bother to change this function.

If `Timer` won't be deprecated, I'd suggest returning a chrono type here. And we should deprecate this function before removing it. Maybe the new function can be called `ElapsedTime` or something similar.

chapulina (changes requested):

include/ignition/common/Timer.hh
```diff
@@ -46,21 +46,25 @@ namespace ignition
 
       /// \brief Get the elapsed time
       /// \return The time
-      public: Time Elapsed() const;
+      public: Time IGN_DEPRECATED(4) Elapsed() const;
+
+      /// \brief Get the elapsed time
+      /// \return The time
+      public: double ElapsedTime() const;
```
chapulina, line 53: Can this be a duration?

src/WorkerPool.cc
```diff
@@ -194,3 +194,35 @@ bool WorkerPool::WaitForResults(const Time &_timeout)
   }
   return signaled && !this->dataPtr->done;
 }
+
+//////////////////////////////////////////////////
+bool WorkerPool::WaitForResults(
```
chapulina, line 199: To avoid duplicating the logic, you could have the old function call this one by converting `Time` to `duration`.

ahcorde:
@osrf-jenkins run tests

src/Timer.cc
```diff
@@ -56,11 +56,45 @@ Time Timer::Elapsed() const
...
+#endif
+    return Time(diff.count());
+#ifndef _WIN32
+# pragma GCC diagnostic pop
+#endif
+  }
+}
 
-    return currentTime - this->start;
+//////////////////////////////////////////////////
```
chapulina, line 85: Same here about avoiding duplication by having the old function call the new one. It will also reduce the number of pragmas.

chapulina (approved):
LGTM with happy CI. I just have one minor comment about code duplication.

ahcorde:
@osrf-jenkins run tests

PR #93, opened 2020-09-11
Fix trajectory info memory leak

JShep1:
Simple fix but took awhile to find and I think deserves some explanation as well as some tools I used that can be used in the future for memory profiling (possibly this part deserves its own tutorial or guide).

It was discovered that there was a significant memory leak in any ign-gazebo examples involving actors.  I used a project called [mem-usage-ui](https://github.com/parikls/mem_usage_ui) which was very simple yet effective in being able to visually confirm the existence of memory leaks.  The output from the actor example that was experiencing serious leakage is as follows: 
Citadel-actor-population-no-delete

It can be seen that the memory leakage grew by about 100Mb in just one minute, if left for 15-20 minutes, this could crash the computer.  Using [heaptrack](https://github.com/KDE/heaptrack) or `valgrind` didn't quite prove effective given it monitors the ruby executable, outputting little to no helpful information.  For this reason, I'd recommend we get a standalone custom gui example out, similar to [custom_server](https://github.com/ignitionrobotics/ign-gazebo/tree/master/examples/standalone/custom_server) as this would allow `valgrind` to do its job.  I did, however, find that [memleax](https://github.com/WuBingzheng/memleax) was quite useful in pinpointing memory leaks, the output is a bit complicated to discern and there are false positives, but it proved very helpful in pointing me right to the culprit function that was leaking (in this case, [ActorMeshAnimationAt(...)](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/src/rendering/SceneManager.cc#L883)).

After further investigation, I found that [line 894 of SceneManager.cc](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/src/rendering/SceneManager.cc#L894) `auto trajs = this->dataPtr->actorTrajectories[_id];` was the problem line that was spawning new `TrajectoryInfo`'s, invoking [this constructor](https://github.com/ignitionrobotics/ign-common/blob/ign-common3/graphics/src/Animation.cc#L426-L430) for every currently existing `TrajectoryInfo` in the `this->dataPtr->actorTrajectories[_id]` vector, I suppose when assigning that vector to a new variable, the (deep) copy operation is made and that vector is duplicated, resulting in the original vector remaining as the value of that `this->dataPtr->actorTrajectories` map and a new copy vector set to `trajs`.

So adding a simple deletion of `this->dataPtr` to the destructor of `TrajectoryInfo` fixed up the memory leak, which can be visually confirmed in the following graph
Citadel-actors-population-delete

So, not perfect, there appears to be another smaller leak somewhere, but ~10Mb / minute is much better than the original.  I'll continue to look for the remaining leaks.

Signed-off-by: John Shepherd <john@openrobotics.org>

### DIFF
--- a/graphics/src/Animation.cc
+++ b/graphics/src/Animation.cc
@@ -408,6 +408,7 @@ TrajectoryInfo::TrajectoryInfo(TrajectoryInfo &&_trajInfo) noexcept
 /////////////////////////////////////////////////
 TrajectoryInfo::~TrajectoryInfo()
 {
+  delete this->dataPtr;
 }
 
 //////////////////////////////////////////////////


chapulina:
Marking this as `beta` for Dome since it's a bug fix.

JShep1:
> Marking this as `beta` for Dome since it's a bug fix.

Sounds good, I'm going to test Dome in a bit to see if the memory leak occurs there too, I noticed there was some significant changes of the Actor code so I wanted to confirm that the leak still occurs for Dome.

JShep1:
Looks like Dome has this memory leak as well and this fix will fix it.  It also appears Dome doesn't have the extra memory leak that Citadel has :thinking:

chapulina (approved):

mjcarroll:
Could we make this a smart pointer in `dome`, then you could probably get rid of all of the constructors and destructor here.

This also currently violates rule-of-five principle.

chapulina:
> Could we make this a smart pointer in dome

Dome will use `ign-common3`, so I don't think it can be changed there. But we could change it from `ign-common4`.

iche033:
Thanks for the fix! 

> It also appears Dome doesn't have the extra memory leak that Citadel has

Dome switched from doing manual interpolation of skeletal bone poses (by ign-common) to interpolation done by the rendering engine. So one place I see that may leak is here:

https://github.com/ignitionrobotics/ign-common/blob/master/graphics/src/SkeletonAnimation.cc#L82

There are also a few other possible places for leaks which I don't a matching `delete`:
https://github.com/ignitionrobotics/ign-common/blob/master/graphics/src/Animation.cc#L505
https://github.com/ignitionrobotics/ign-common/blob/master/graphics/src/Animation.cc#L220

yikes, quite a few leaks in our animation code

PR #97, opened 2020-09-18
Fix env behavior to return true on empty vars

### LINKED ISSUE #96: Bug: getenv does not properly detect set-but-empty variable
Discovered when writing some logic that checks for the presence of an environment variable, rather than the value.

You could in theory do something like
```
# In bash/zsh
export IGN_ENV_VAR=""

# and then in C++
ignition::common::env("IGN_ENV_VAR", foo)  # returns FALSE, but should be TRUE
```

This is completely valid (at least on Posix systems, unsupported on Windows)

Ref: https://github.com/ignitionrobotics/ign-common/blob/526c2aceb847dcdbfc2d25fc72457caaece03d3c/src/Util.cc#L328-L333

mjcarroll:
Empty environment variables are valid on Linux (and other posix systems), we should return `true` in those cases.

Closes #96 


include/ignition/common/Util.hh
```diff
@@ -217,9 +217,12 @@ namespace ignition
     /// \brief Find the environment variable '_name' and return its value.
     /// \param[in] _name Name of the environment variable.
     /// \param[out] _value Value if the variable was found.
+    /// \param[in] _allowEmpty Allow set-but-empty variables.
+    ///           (Unsupported on Windows)
     /// \return True if the variable was found or false otherwise.
     bool IGNITION_COMMON_VISIBLE env(
-        const std::string &_name, std::string &_value);
+        const std::string &_name, std::string &_value,
+        bool _allowEmpty=false);
```
chapulina, line 225: I believe this breaks ABI. We could add an overload that takes the extra argument, but I think it won't be possible to give it a default value.

src/Util_TEST.cc
```diff
@@ -172,6 +172,100 @@ TEST(Util_TEST, emptyENV)
   EXPECT_TRUE(var.empty());
 }
 
+/////////////////////////////////////////////////
+TEST(Util_TEST, env_set)
```
chapulina, line 176:
```suggestion
TEST(Util_TEST, envSet)
```

Same for other tests below

chapulina (changes requested):
Due to the different behavior on Windows, it sounds like we shouldn't rely on empty env vars for any cross-platform code :confused: 

Besides that and the ABI issue, the rest looks good to me.

mjcarroll:
> Due to the different behavior on Windows, it sounds like we shouldn't rely on empty env vars for any cross-platform code confused

I believe that is true, but it also breaks expectations on Linux a bit if you are just checking for a value to be _set_ rather than checking the value itself.

include/ignition/common/Util.hh
mjcarroll: You are correct, I have added a new function with a TODO to deprecate the old one (which I will do on forward-port)

src/Util_TEST.cc
mjcarroll: d7821e8

src/Util.cc
```diff
@@ -321,21 +321,41 @@ ignition::common::SystemPaths *ignition::common::systemPaths()
...
   const DWORD buffSize = 32767;
   static char buffer[buffSize];
   if (GetEnvironmentVariable(_name.c_str(), buffer, buffSize))
   {
     v = buffer;
   }
+
+  if(!v.empty())
+    valid = true;
+
```
chapulina, line 344: I'm still thinking about how this may affect Windows unintentionally. How about printing a warning if `_allowEmpty == true` on Windows?

chapulina (commented):
Code checker is angry:

```
   ./src/Util_TEST.cc:181:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
  ./src/Util.cc:342:  Missing space before ( in if(  [whitespace/parens] [5]
  ./src/Util.cc:355:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
  ./include/ignition/common/Util.hh:220:  Line ends in whitespace.  Consider deleting these extra spaces.  
```

And Windows CI failed:

```
D:\Jenkins\workspace\ignition_common-ci-pr_any-windows7-amd64\ws\ign-common\src\Util_TEST.cc(179): error C3861: 'setenv': identifier not found
```

chapulina (approved):

PR #98, opened 2020-09-23
Fix more Animation memory leak

iche033:
Continuing the effort from pull request #93, this PR fixes memory leaks in the animation classes by deleting objects that are new'd and using smart pointer where possible


graphics/src/Animation.cc
```diff
@@ -262,6 +262,8 @@ PoseAnimation::~PoseAnimation()
 {
   delete this->positionSpline;
   delete this->rotationSpline;
+  for (auto kf : this->keyFrames)
+    delete kf;
```
mjcarroll, line 266: When bringing this patch forward to the main branch, let's change these to all use smart pointers.  I don't think it's possible in a minor, because it would break ABI.

graphics/src/SkeletonAnimation.cc
```diff
@@ -45,6 +45,8 @@ SkeletonAnimation::SkeletonAnimation(const std::string &_name)
 //////////////////////////////////////////////////
 SkeletonAnimation::~SkeletonAnimation()
 {
+  for (auto &it : this->data->animations)
+    delete it.second;
```
mjcarroll, line 49: These could also be replaced by smart pointers.

JShep1 (approved):
Looks good, I agree with Michael about the smart pointers.

graphics/src/SkeletonAnimation.cc
iche033: yep done. c4c3a86

iche033:
> I think we should remove raw pointers when porting forward. I can do this separately if you would like.

yeah that'll be great. The remaining one is protected variable so I'm not sure if it's possible to deprecate it or if we should just change the type.

iche033:
ign-common3 is in Ignition Dome so I think we should hold off on merging this until after the release

iche033:
ok common3 is stable. Merging after last round of CI builds

PR #99, opened 2020-09-25
Remove FilesystemBoost

mjcarroll:
As far as I can tell, this is no longer used anywhere.


mjcarroll:
Just kidding :sweat_smile:

PR #101, opened 2020-09-29
Fix loading dae file with hierarchical node that does not have a name

iche033:
pull request #78 fixed loading hierarchical submeshes in ColladaLoader but we ran into an edge case that if a node in the dae file does not have a name, it would be given the name of the previous node. For example,the [Polaris Ranger EV model](https://app.ignitionrobotics.org/OpenRobotics/fuel/models/Polaris%20Ranger%20EV) model has the node structure below and the result is that the nested node inside `node_B` would be incorrectly given the name `node_A` :

```
<node name="node_A">
</node>
<node name="node_B">
    <node>
        ...
    </node>
</node>
```

The fix adds the logic to check that if a node does not have a name, it'll then be appended and become part of the parent mesh. It restores the original behavior before pull request #78 was merged while preserving the hierarchical submesh fix.


luca-della-vedova (approved):
Works great for me, thanks for the fix!

mjcarroll:
Looks like the Ubuntu build is erroring due to low disk space: https://build.osrfoundation.org/job/ignition_common-ci-pr_any-ubuntu_auto-amd64/395/console

CC: @j-rivero and @tfoote (current build farmer)

mjcarroll:
@iche033 Do you think you can add a `dae` file that exercises this behavior?  It should make codecov a bit happier.

iche033:
yep added test in d958def

graphics/src/ColladaLoader.cc
```diff
@@ -477,6 +477,31 @@ void ColladaLoaderPrivate::LoadNode(tinyxml2::XMLElement *_elem, Mesh *_mesh,
...
+      {
+        nodeName = name;
+        break;
+      }
+    }
+    if (nodeName.empty())
+    {
+      // if none of the ancestor node has a name, then create a custom name
+      static int nodeCounter = 0;
+      nodeName = "unnanmed_submesh_" + std::to_string(nodeCounter++);
```
mjcarroll, line 501:
```suggestion
      nodeName = "unnamed_submesh_" + std::to_string(nodeCounter++);
```

mjcarroll (approved):
One spelling suggestion, but LGTM.

graphics/src/ColladaLoader.cc
iche033: fixed. b73e933ecc4d110a437809c6808487fb18a5d295

PR #84, opened 2020-08-01
3 ➡️ 4

chapulina:

PR #104, opened 2020-10-14
[port] 3 ➡️ 4

mjcarroll:
Intended to supercede #84.

I combined a few of the commits to more closely align with how they landed on bitbucket.  This should be the last time that doing a forward port will be complicated by that!


mjcarroll:
Same tests are failing for me locally, I'm looking at them right now.

mjcarroll:

mjcarroll:

graphics/include/ignition/common/Animation.hh
```diff
@@ -34,6 +34,7 @@ namespace ignition
     class KeyFrame;
     class PoseKeyFrame;
     class NumericKeyFrame;
+    class AnimationPrivate;
```
chapulina, line 37: I don't see `AnimationPrivate` used in this file. Bad merge?

graphics/include/ignition/common/Material.hh
```diff
@@ -222,6 +242,14 @@ namespace ignition
       /// \return the lighting enabled state
       public: bool Lighting() const;
 
+      /// \brief Set the Physically Based Rendering (PBR) material
+      /// \return The PBR material to set to.
+      public: void SetPbrMaterial(const Pbr &_pbr);
+
+      /// \brief Get the Physically Based Rendering (PBR) material
+      /// \return Pointer to the PBR material. Null if it does not exist.
+      public: Pbr *PbrMaterial() const;
```
chapulina, line 251: I think this should return a const pointer, because the caller can use `SetPbrMaterial` to set it. This may be a good time to make the change and document on the migration guide.

chapulina (approved):
Nice cleanup! I just have a couple of comments which don't need to be addressed on this PR.

graphics/include/ignition/common/Material.hh
mjcarroll: Agree. I'll follow up on this.

PR #106, opened 2020-10-17
Add Console example

chapulina:
The motivation was to test this issue: https://github.com/ignitionrobotics/ign-common/issues/6

Also added a README to all examples.

I think we should consider putting each example into its own directory so they're self-contained. A new user might not necessarily know how to split them, especially as they get more complex.


ahcorde (approved):
LGTM

mjcarroll (approved):

PR #107, opened 2020-10-19
[ign-common3] Miscellaneous Cleanups

mjcarroll:
A few things that I found when doing the forward port

1. Use variables in the examples `CMakeLists.txt` file to keep things from breaking across ports
2. Add a string "Join" function for convenience.
3. Replace all `getenv`, `setenv`, `putenv`, `unsetenv` with Ignition functionality.


chapulina (approved):
LGTM with happy CI once Jenkins is back

PR #108, opened 2020-10-21
⬆️ 3.7.0

mjcarroll:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.10.2 FATAL_ERROR)
 #============================================================================
 # Initialize the project
 #============================================================================
-project(ignition-common3 VERSION 3.6.1)
+project(ignition-common3 VERSION 3.7.0)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.10.2 FATAL_ERROR)
 #============================================================================
 # Initialize the project
 #============================================================================
-project(ignition-common3 VERSION 3.6.1)
+project(ignition-common3 VERSION 3.6.2)
```
chapulina, line 6: Already mentioned to @mjcarroll , but for completeness I'll leave it here too, I think this should be a minor bump because of the new string join and env functions.

See comparison to 3.6.1:

https://github.com/ignitionrobotics/ign-common/compare/ignition-common3_3.6.1...ign-common3

mjcarroll: Changed to minor bump, thanks!

chapulina (approved):

PR #110, opened 2020-10-28
Video encoder timing test

jennuine:
Test that verifies video encoder timing, which uses [VideoEncoder](https://github.com/ignitionrobotics/ign-common/blob/ign-common3/av/src/VideoEncoder.cc) to encode several frames to a video file and verifies the duration using the [Video](https://github.com/ignitionrobotics/ign-common/blob/encoder_timing_test/av/src/Video.cc) class with a new `Duration()` function.

### DIFF
--- a/av/src/Video.cc
+++ b/av/src/Video.cc
@@ -289,3 +289,9 @@ int Video::Height() const
 {
   return this->dataPtr->codecCtx->height;
 }
+
+/////////////////////////////////////////////////
+int64_t Video::Duration() const
+{
+  return this->dataPtr->formatCtx->duration;
+}
--- a/av/src/Video.cc
+++ b/av/src/Video.cc
@@ -291,7 +291,7 @@ int Video::Height() const
 }
 
 /////////////////////////////////////////////////
-int64_t Video::Duration() const
+std::chrono::duration<int64_t, std::ratio<1, AV_TIME_BASE>> Video::Duration() const
 {
-  return this->dataPtr->formatCtx->duration;
+  return std::chrono::duration<int64_t, std::ratio<1, AV_TIME_BASE>>(this->dataPtr->formatCtx->duration);
 }
--- a/av/src/Video.cc
+++ b/av/src/Video.cc
@@ -291,8 +291,7 @@ int Video::Height() const
 }
 
 /////////////////////////////////////////////////
-std::chrono::steady_clock::duration Video::Duration() const
+Video::Length Video::Duration() const
 {
-  return std::chrono::duration<int64_t, std::ratio<1, AV_TIME_BASE>>(
-      this->dataPtr->formatCtx->duration);
+  return Video::Length(this->dataPtr->formatCtx->duration);
 }
--- a/av/src/Video.cc
+++ b/av/src/Video.cc
@@ -291,7 +291,8 @@ int Video::Height() const
 }
 
 /////////////////////////////////////////////////
-Video::Length Video::Duration() const
+std::chrono::steady_clock::duration Video::Duration() const
 {
-  return Video::Length(this->dataPtr->formatCtx->duration);
+  return std::chrono::duration<int64_t, std::ratio<1, AV_TIME_BASE>>(
+      this->dataPtr->formatCtx->duration);
 }
--- a/av/src/Video.cc
+++ b/av/src/Video.cc
@@ -291,7 +291,7 @@ int Video::Height() const
 }
 
 /////////////////////////////////////////////////
-std::chrono::duration<int64_t, std::ratio<1, AV_TIME_BASE>> Video::Duration() const
+Video::Length Video::Duration() const
 {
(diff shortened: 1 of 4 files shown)


av/src/Video.cc
```diff
@@ -289,3 +289,10 @@ int Video::Height() const
 {
   return this->dataPtr->codecCtx->height;
 }
+
+/////////////////////////////////////////////////
+std::chrono::steady_clock::duration Video::Duration() const
+{
+  return std::chrono::duration<int64_t, std::ratio<1, AV_TIME_BASE>>(
+      this->dataPtr->formatCtx->duration);
```
mjcarroll, line 297: I think it would make sense to typedef this definition and return that type here.

The duration that we are returning isn't a `steady_clock` duration here.

test/integration/encoder_timing.cc
```diff
@@ -0,0 +1,64 @@
...
+
+  int frameCount = 0;
+  while (frameCount != _fps*_seconds)
+  {
+    if (_vidEncoder.AddFrame(kFrame, kSize, kSize))
+      ++frameCount;
+  }
+
+  _vidEncoder.Stop();
+  _video.Load(common::cwd() + "/TMP_RECORDING.mp4");
```
mjcarroll, line 46: Use `common::joinPath` here.

test/integration/encoder_timing.cc
```diff
@@ -0,0 +1,64 @@
...
+
+using namespace ignition;
+using namespace common;
+
+const unsigned int kSize = 10;
+const unsigned char *kFrame = new unsigned char[kSize*kSize];
+
+// set to 720ms because video duration missing additional 18 frames
+//    which may be due to how video encoding works
+const int kTol = 720;
```
mjcarroll, line 31: Can this be a `std::chrono::milliseconds` to be a bit more explicit?

mjcarroll:
I believe that the Windows failure here is somewhat expected, being that `av` wasn't currently being built.  The integration test needs to be dependent on that component being successfully built.

av/src/Video.cc
jennuine: Would this be ok? https://github.com/ignitionrobotics/ign-common/blob/10b186592ccf650af373ba8f4a485713081ab530/av/include/ignition/common/Video.hh#L61-L68

jennuine: I don't mind to switch it to public but isn't just the alias private and what is being returned is `std::chrono::duration`?

mjcarroll: I believe it's the full type that's being returned, but as an example in this case:

```
// Works
auto dt = v.Duration();

// Works
std::chrono::duration<int64_t, std::ratio<1, 1000000>> dt = v.Duration();

// Works
auto millis = std::chrono::duration_cast<std::chrono::milliseconds>(v.Duration());

// Doesn't work
Video::Length dt = v.Duration();

// Doesn't work
std::chrono::milliseconds dt = v.Duration();
```

I think that the 4th case is somewhat convenient in the case that you may want to use auto?

mjcarroll: Also, people who use either `auto` or `Video::Length` shouldn't get burned if the definition of `Length` changes.

jennuine: Ah I see, thanks. I made the changes here: c6f76c8a7f6e1c0355538b95f29c08feb1f24d00

test/integration/encoder_timing.cc
```diff
@@ -0,0 +1,64 @@
...
+TEST(EncoderTimingTest, Duration)
+{
+  VideoEncoder vidEncoder;
+  Video video;
+
+  durationTest(vidEncoder, video, 50, 1);
+  durationTest(vidEncoder, video, 30, 2);
+  durationTest(vidEncoder, video, 25, 5);
+
+  delete [] kFrame;
```
mjcarroll, line 63: I think you either want to:

a) Allocate and deallocate this frame completely within the test, and pass it to duration test.
b) Allocate the deallocate the frame globally and don't do any memory management in the test.

Mixing the two could potentially make it easy to introduce bugs if more tests are ever added here.

Since this is also an array, you may consider using `std::array<unsigned char, kSize*kSize>`, or a `std::vector` to not have to deal with memory allocation at all.

jennuine: Thanks for the feedback. Better? 39ccc6ae3fef37c48c8d791a8d91ba1e03a291f6

mjcarroll: Looks great, just looks like you need to include `<array>` to make MacOS happy.

jennuine: done 67db98b0246542bc0dd6fcddcbae0121e5069bba. Is there a way to check things like this beforehand?

mjcarroll: Sometimes codecheck can catch it.  There are some explicit tools that will do it as well such as: https://include-what-you-use.org/

In this case it's generally good practice to do, but some compilers are a bit more picky about it.  There are many cases where an extra iteration or two may be spent appeasing Windows or Mac...

mjcarroll (approved):

PR #111, opened 2020-10-29
Call ignition::common::load() from Video constructor

peci1:
Just copying from VideoEncoder so that Video class can be used standalone and `ignition::common::load()` doesn't need to be called from downstream code.


chapulina (approved):
Makes sense to me.

mjcarroll (approved):

PR #114, opened 2020-10-30
Video decoder fixes

peci1:
This PR fixes a few things regarding video decoding. Descriptions of the respective fixed problems is in the commit messages.

Basically, without this PR, decoding an MP4 video ends either with a segfault or nonsense output. I wonder if it ever worked for someone before.

The output pixel format fix assumes that the desired format of the output buffer is R8G8B8 (that was deduced from the swscale config). But as the docs were scarce here, I'm not sure whether it wasn't desired to output YUV420.

Finally, I added a test that checks a decode-encode-decode loop for basic functionality. It seems that up to FPS frames get lost at the end of the encoded video, but the same was observed in #110 and accepted, so I also accept it here.


peci1:
I manually tested compatibility of this PR with #105 and it ran ok.

test/integration/video_encoder.cc
```diff
@@ -0,0 +1,108 @@
...
+TEST(EncoderDecoderTest, DecodeEncodeDecode)
+{
+  const unsigned int fps = 25u;
+
+  const char* testVideoInName = "test_video.mp4";
+  const char* testVideoOutName = "test_video_out.mp4";
+  const auto testVideoInPath = common::joinPaths(TEST_PATH, "data",
+                                                 testVideoInName);
+  const auto testVideoOutPath = common::joinPaths(common::cwd(),
+                                                  testVideoOutName);;
```
mjcarroll, line 48: Nit: extra semicolon.

test/integration/video_encoder.cc
```diff
@@ -0,0 +1,108 @@
...
+
+  Video decoder;
+  decoder.Load(testVideoInPath);
+
+  VideoEncoder encoder;
+  encoder.Start("mp4", "", decoder.Width(), decoder.Height(), fps);
+
+  const size_t bufferSize = decoder.Width() * decoder.Height() * 3;
+  auto* buf = new unsigned char[bufferSize];
```
mjcarroll, line 57: I would prefer an stl type to raw `new` calls.

mjcarroll (commented):
Small nits.

test/integration/video_encoder.cc
peci1: which one? is it "legal" to take the buffer of a std::vector?

mjcarroll: I believe a `std::vector` would be appropriate here.  It is guaranteed to be contiguous.  You can get the underlying data via [`std::vector::data`](https://en.cppreference.com/w/cpp/container/vector/data)

test/integration/video_encoder.cc
peci1: Fixed.

mjcarroll:
I believe that you'll also need to update the branch with latest ign-common3 changes as well as do a DCO signoff.

test/integration/video_encoder.cc
peci1: I'm afraid that this is not possible with the current API of VideoEncoder. `NextFrame()` takes `uint**`, however `vector::data()` returns a single pointer and you can't take the address of it as an rvalue. I could store the result of `data()` into a local variable, but that would be, again, `uint*` which we wanted to get rid of... 

I was really thinking that the double-pointer API is horrible, but didn't want to change it to stay compatible. The double-pointer argument makes sense in places where the calee allocates the buffer, but in this case, it has to come pre-allocated. So there's no reason why it couldn't come as a single pointer. Or, even better, as a `std::vector<uint>` reference. Or, the function could take a double pointer and allocate it, but then it should also return the length of the buffer (although it is pretty straight-forward to tell what the length should be thanks to the added documentation).

I think the most logical thing would be if `NextFrame()` would take arg of type `ignition::common::Image&` and used `Image::SetFromData()` to save the decoded data. This way, no allocation would have to happen from the outside, and at the same time, the user would get a nice interface with correctly computed width, height and more. The only drawback is that there are no copy-free ways to get the image into `Image` and out of it - both `SetFromData()` and `Data()` create copies of the internal buffer. But I can imagine we could add something like `BYTE* RawData()` to image which would be a pointer to the internal buffer used by FreeImage. Then, `NextFrame()` could call the standard `SetFromData()` for the first time (if the image is invalid), and reuse the automatically (and correctly) allocated buffer in further calls when given a reference to the same image.

To not break API, I can add a new overload of `NextFrame()` taking the `Image` argument. It just comes with one difficulty for me - it'd add a dependency of the `av` component on the `graphics` component. I'm neither sure if adding such dependency is valid, nor how to correctly express it in terms of ign-cmake. Or it could be a "voluntary dependency" - if both `av` and `graphics` are built, this function would be a part of the API, and otherwise it would be missing. What do you think of such design?

peci1:
Closing in favor of #125 which contains all of the fixes that were introduced in this PR.

PR #115, opened 2020-10-30
Little fixes for video encoder

peci1:
I've played around with the video encoder and found (and fixed) the following little issues.

1. If you try to encode a video and supply a zero timestamp as the time of the first frame, it gets skipped.

1. Reset() did not reset timeStart and timePrev. Reset of timeStart is not technically needed, because Start() would correctly set it correctly based on the reset-ted value of frameCount. However, resetting it doesn't hurt anything. On the other hand, timePrev remained at its last value before Reset() call, which might lead to skipping the first frame of another video recording.

### DIFF
--- a/av/src/VideoEncoder.cc
+++ b/av/src/VideoEncoder.cc
@@ -510,7 +510,8 @@ bool VideoEncoder::AddFrame(const unsigned char *_frame,
 
   // Skip frames that arrive faster than the video's fps
   double period = 1.0/this->dataPtr->fps;
-  if (dt < std::chrono::duration<double>(period))
+  if (this->dataPtr->frameCount > 0u &&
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):
LGTM.

* MacOS failure is a flaky test.
* Windows failure is because the integration test is not skipped when `av` dependencies are not found.  This will either be fixed by a) Add `ffmpeg` and friends to Windows CI or b) correctly skipping the integration test.

peci1:
Can I assume the solution for windows tests will be done in #105?

mjcarroll:
> Can I assume the solution for windows tests will be done in #105?

Yes, that will have to be done before #105 gets merged in.

iche033 (approved):

PR #116, opened 2020-11-01
Do not hardcode USE_EXTERNAL_TINYXML2 to OFF on Windows

traversaro:
Fortunately now in 2020 we have package managers on Windows, so at least giving an option to use an external TinyXML2 also on Windows make sense. : )

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,9 +33,8 @@ endif()
 if(UNIX OR APPLE)
   option(USE_EXTERNAL_TINYXML2 "Use a system-installed version of tinyxml2" ON)
 elseif(WIN32)
-  # This is always false for Windows, so we can avoid trying to search for a
-  # system installation of tinyxml2.
-  set(USE_EXTERNAL_TINYXML2 false)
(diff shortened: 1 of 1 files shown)


traversaro:
The `ignition_common-ci-pr_any-homebrew-amd64` failure seems to be unrelated.

mjcarroll:
Do you think it makes sense to move this logic into `ign-cmake` so that it can be used in other places as well?  I believe TinyXML2 is at least used here, in `sdformat` and a few more places up the stack.

traversaro:
> 
> 
> Do you think it makes sense to move this logic into `ign-cmake` so that it can be used in other places as well? I believe TinyXML2 is at least used here, in `sdformat` and a few more places up the stack.


I don't think so. `USE_EXTERNAL_TINYXML2` is an option that only make sense for projects that vendor tinyxml2, and ignition-cmake does not provide any function to vendor tinyxml2 itself. Furthermore, given that package managers such as conan, vcpkg and conda are becoming to be more used on Windows, I hope that for future projects the need to vendor dependencies such as tinyxml2 will disappear.

mjcarroll (approved):

PR #117, opened 2020-11-02
Remove plugin in favor of ign-plugin

mjcarroll:


ahcorde (changes requested):
``` 
  In file included from /usr/local/include/ignition/common4/ignition/common.hh:23,
                   from /github/workspace/examples/assert_example.cc:1:
  /usr/local/include/ignition/common4/ignition/common/detail/Plugin.hh:24:10: fatal error: ignition/common/Plugin.hh: No such file or directory
   #include "ignition/common/Plugin.hh"
```

mjcarroll:
I actually don't think that @chapulina is going to let us do this, because we should be technically doing tick tock.

chapulina:
> I actually don't think that @chapulina is going to let us do this, because we should be technically doing tick tock.

You shouldn't let yourself do that either :wink:

PR #118, opened 2020-11-02
Added FlagSet utility class

peci1:
This class helps managing a set of flags that are defined as a C++11 enum class.

Adapted from https://github.com/mrts/flag-set-cpp (MIT license, compatible with Apache 2.0).

I have a use for this class in my next PR (not yet published as it needs polishing). But I think it might be generally useful to have a canonical way to work with flag sets.

### DIFF
--- a/include/ignition/common/FlagSet.hh
+++ b/include/ignition/common/FlagSet.hh
@@ -39,16 +39,23 @@ namespace ignition::common
 {
 
 /// \brief Set of flags defined by a C++11 enum class.
-/// \note For FlagSet to work, the enum has to have a value called '_' (just the
-/// underscore) as the very last element.
+/// \note If the enum's last element is '_' (just underscore), the convenience
+/// signature can be used specifying just the enum type (the underscore will be
+/// automatically used as LastElement, and will be excluded from the list of
+/// valid values).
 /// \note FlagSet only works for enums that do not contain negative values.
 /// \note The underlying bitset representing this FlagSet will have as many bits
-/// as is the value of the '_' element. Be aware of that when calling Count(),
-/// Size() etc. Also be aware of that when creating a FlagSet of an enum that
-/// would contain very high numbers in the underlying representation.
+/// as is the size_t value of LastElement (+- 1). Be aware of that when calling
+/// Count(), Size() etc. Also be aware of that when creating a FlagSet of an
+/// enum that would contain very high numbers in the underlying representation.
 /// \tparam T The enum class type. The underlying type of the enum has to be
-/// implicitly convertible to size_t.
-template<typename T>
+/// convertible to size_t.
+/// \tparam LastElement Last element of the enum. If the enum's last element is
+/// '_' (just underscore), it can be deduced automatically.
+/// \tparam ExcludeLast If true, LastElement is considered as an element beyond
+/// all valid elements and will not be treated as a valid value of the FlagSet.
+/// If false, the last element is treated as a valid value.
+template<typename T, T LastElement = T::_, bool ExcludeLast = true>
 class FlagSet
 {
   /// \brief Create an empty FlagSet (no flags set).
@@ -63,6 +70,16 @@ class FlagSet
   public: FlagSet(const T& _val)
   {
     flags.set(static_cast<UnderlyingType>(_val));
+  }
+
+  /// \brief Construct a FlagSet with the given flags set and all other unset.
+  /// \param[in] _list The list of flags to set.
+  public: explicit FlagSet(const std::initializer_list<T>& _list)
+  {
+    for (const auto& val : _list)
+    {
+      flags.set(static_cast<UnderlyingType>(val));
+    }
   }
 
   // Binary operations.
@@ -174,11 +191,17 @@ class FlagSet
 
   // Methods from std::bitset.
 
+  /// \brief Test FlagSet equality.
+  /// \param[in] _o The other FlagSet.
+  /// \return Whether the FlagSets represent the same set of flags.
   public: bool operator==(const FlagSet& _o) const
   {
     return flags == _o.flags;
   }
 
+  /// \brief Test FlagSet inequality.
+  /// \param[in] _o The other FlagSet.
+  /// \return Whether the FlagSets represent different sets of flags.
   public: bool operator!=(const FlagSet& _o) const
   {
     return !(*this == _o);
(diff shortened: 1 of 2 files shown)


mjcarroll:
This is a bit difficult to evaluate without more context, do you think you could open a draft PR showing the intended use?

peci1:
Okay, here's the example usage:

https://github.com/ignitionrobotics/ign-common/blob/2367c1cb1ae78b4b517efbdbdfa85bc6d525c10f/av/include/ignition/common/HWVideo.hh#L23-L38

https://github.com/ignitionrobotics/ign-common/blob/2367c1cb1ae78b4b517efbdbdfa85bc6d525c10f/av/src/VideoEncoder.cc#L165-L192

include/ignition/common/FlagSet.hh
```diff
@@ -0,0 +1,348 @@
...
+
+  /// \brief Test whether no flag is set.
+  /// \return True if at no flag is set to true.
+  public: bool None() const
+  {
+    return flags.none();
+  }
+
+  /// \brief Retrurn a FlagSet with all flags set to true.
+  public: static FlagSet ALL()
```
mjcarroll, line 278: I think also that `All` vs `ALL` may be a bit confusing.

Style Nit: We haven't been using all-caps for static methods (that I'm aware of).

mjcarroll (commented):
One piece of API feedback in comments.

Does it also make sense to implement `std::hash` the same way that `bitset` does so that it is usable in things like `unordered_map` and `unordered_set`?

include/ignition/common/FlagSet.hh
chapulina: No, static methods follow the same `CamelCase` rule as other member functions.

peci1:
Do you have a use-case for indexing by bitsets? I can imagine indexing by enum values, but by whole bitsets?

include/ignition/common/FlagSet.hh
peci1: Okay, what could be a better name if we'd want to reserve `All()` to the corresponding `all()` function of the bitset? Also `NONE` should be renamed. Some suggestions (I don't have a favorite one yet):

1. AllTrue + AllFalse
1. AllTrue + NoneTrue
1. AllEnabled + NoneEnabled
1. AllEnabled + AllDisabled
1. AllSet + NoneSet

but all of these names still don't clearly differ from the names usually used for getters. The idea was to provide something like `ignition::math::Vector3d::One` to avoid construction patterns like `FlagSet<>().Set()`.

The `NONE` helper is not really needed as the default behavior of the constructor is to set no flags, but I prefered parity of the offered functions.

Another option could be to add a `FlagSet<>(bool)` constructor which would take a parameter telling whether all flags should be set or not. But I think it's less readable and verbose.

mjcarroll: I like `All`, `Any`, `None` as our three checks, as it mirrors STL's bitset, so I think we're good there.

I think my preference for the `static` methods would be `AllEnabled`, `AllDisabled`, but I think it would be worth asking some non-native English speakers to see if it makes as much sense to them.

peci1: I am non-native :-D But it's a good idea. Where to ask? I assume few people will now have others around to ask them on a coffee break...

mjcarroll:
> Do you have a use-case for indexing by bitsets? I can imagine indexing by enum values, but by whole bitsets?

I was mainly thinking of parity with the `bitset` API, but off hand, here are two things I could think of:

Dispatching based on a certain configuration of flags, so maybe:

```
std::unorderd_map<FlagSet, std::function<void(void)>> dispatch = {
  {FlagSet::AllEnabled, doFoo},
  {FlagSet::AllDisabled, doBar},
};
```

Or potentially keeping  a set of valid configurations to check against:

```
std::set<FlagSet> valid;
valid.contains(FlagSet::AllEnabled);
```

include/ignition/common/FlagSet.hh
mjcarroll: I'm pinging our company chat to recruit some eyeballs.

chapulina: Keeping `All` / `Any` / `Set` as parallels to the `std::bitset` API also make sense to me.

For the static methods, `AllSet` / `NoneSet` make slightly more sense to me - the class is called `FlagSet` afterall. But it's not a strong preference.

mjcarroll:
@peci1 friendly ping

peci1:
I'll have a look at it over Xmas. This week's my vacation.

chapulina:
@peci1 , do you need this to go into `ign-common3` so it's available to released collections (Citadel and Dome)? If not, we're starting a new package to hold this kind of utility that doesn't bring any dependencies: https://github.com/ignitionrobotics/ign-utils/. That will be released into Edifice, but it's not clear yet what downstream libraries will depend on it.

chapulina:
> do you need this to go into ign-common3

Oh I just saw that you're using it for the video encoder in #125 , so that answers my question.

peci1:
Yes, you're right, I'd like to use this PR in the hardware support for video encoder. But it could be possible to factor this out to a new package starting with Edifice.

include/ignition/common/FlagSet.hh
peci1: Okay, renamed according to Louise.

peci1:
I did what was requested (renamed the static constructors and added support for std::hash).

I got one more idea - now the class requires that the last element of the enum is an underscore. That is because of two things:

1. determine the size of the bitset in `private: std::bitset<static_cast<UnderlyingType>(T::_)> flags;`
2. disable `FlagSet operator|(const T& _lhs, const T& _rhs)` for enums (or other types) not containing the underscore (as a safety measure, otherwise all types would be OReable to a FlagSet)

I think 1. can be overcome by exposing an explicit template argument for the number of elements in the enum. That would allow much broader use of FlagSet even with enums you do not control. You would just have to pass the correct number of elements. I know this can be difficult to maintain, but many projects use something other than underscore, e.g. `END` or other values to signal the last enum element.

It seems that it is not possible to overcome 2., so let's say only underscore-ending enums would get this syntactic sugar (ORing two base values creates the FlagSet containing both of them). But that's just syntactic sugar.

What do you think about this?

mjcarroll:
I don't think I mind having the underscore as the last entry in this case.  That seems easier (maintenance-wise) than having to keep track of the number of elements.  I do see the advantage of exposing the other alternative in the case that it is an enum that we do not maintain/control.

I also agree that disabling the `|` for things without the underscore is reasonable, and that functionality is reserved for enums with the correct sentinel value.

mjcarroll:
Friendly ping @peci1, is there anything else you want to do here?

peci1:
I implemented the option for explicit specification of the last element (and whether to treat it as a valid member of the enum or not). I also added some more convenience methods like initializer-list constructor.

I'm satisfied with this PR now (but feel free to comment on the new changes).

mjcarroll (approved):

PR #120, opened 2020-11-04
Set LC_NUMERIC=C to avoid issues reading files

ahcorde:
Related with this issue https://github.com/ignitionrobotics/ign-rendering/issues/101

### DIFF
--- a/graphics/include/ignition/common/MeshLoader.hh
+++ b/graphics/include/ignition/common/MeshLoader.hh
@@ -17,6 +17,7 @@
 #ifndef IGNITION_COMMON_MESHLOADER_HH_
 #define IGNITION_COMMON_MESHLOADER_HH_
 
+#include <clocale>
 #include <string>
 
 #include <ignition/common/graphics/Export.hh>
@@ -32,7 +33,9 @@ namespace ignition
     class IGNITION_COMMON_GRAPHICS_VISIBLE MeshLoader
     {
       /// \brief Constructor
-      public: MeshLoader() = default;
+      public: MeshLoader() {
+        std::setlocale(LC_NUMERIC, "C");
+      }
 
       /// \brief Destructor
       public: virtual ~MeshLoader() = default;


graphics/include/ignition/common/MeshLoader.hh
```diff
@@ -32,7 +33,9 @@ namespace ignition
     class IGNITION_COMMON_GRAPHICS_VISIBLE MeshLoader
     {
       /// \brief Constructor
-      public: MeshLoader() = default;
+      public: MeshLoader() {
```
chapulina, line 36:
```suggestion
      public: MeshLoader()
      {
```

chapulina (commented):
Could you write a test to make sure we don't break it and that it works with all platforms? There are some examples here:

* https://github.com/osrf/sdformat/blob/master/test/integration/locale_fix_cxx.cc
* https://github.com/osrf/sdformat/blob/master/test/integration/locale_fix.cc

mjcarroll:
@j-rivero to add tests.

graphics/include/ignition/common/MeshLoader.hh
```diff
@@ -32,7 +33,9 @@ namespace ignition
...
+        std::setlocale(LC_NUMERIC, "C");
```
traversaro, line 37: This is influencing the global state of the process in a way that is not documented and may not be ideal for a library. Furthermore, other part of the process can then set again the `LC_NUMERIC` to other values, breaking the functionality of the class **after** it has been created.

j-rivero: I think that this is relevant to what you explained here https://github.com/ignitionrobotics/ign-rendering/issues/136, right? So instead of this PR we should try to identify which part of the code is using locale dependent commands and apply changes there.

traversaro: Exactly. To be perfectly clear, it is not tragic if this is merged, but I still wanted to raise the point that is not ideal to perturb the global state from a library.

mjcarroll: I agree, this is kind of a concerning workaround.  It may not break anything as-is, but it's sort of a bad "code smell" that this is necessary.

chapulina: As discussed in one of the meetings, let's decline this PR and try to address the underlying issue instead.

chapulina:
We should address the underlying issue instead.

PR #123, opened 2020-11-06
Add bazel buildsystem support

mjcarroll:
This introduces necessary infrastructure and changes to make ign-common build with Bazel.

### DIFF
--- a/graphics/src/ColladaExporter.cc
+++ b/graphics/src/ColladaExporter.cc
@@ -25,6 +25,10 @@
 
 #include "tinyxml2.h"
 
+#if TINYXML2_MAJOR_VERSION >= 6
+  #define TINYXML2_MAJOR_VERSION_GE_6
+#endif
+
 #ifdef _WIN32
   #define snprintf _snprintf
 #endif
--- a/graphics/src/ColladaExporter.cc
+++ b/graphics/src/ColladaExporter.cc
@@ -25,8 +25,10 @@
 
 #include "tinyxml2.h"
 
+#ifndef TINYXML2_MAJOR_VERSION_GE_6
 #if TINYXML2_MAJOR_VERSION >= 6
   #define TINYXML2_MAJOR_VERSION_GE_6
+#endif
 #endif
 
 #ifdef _WIN32
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,0 +1,104 @@
+load(
+    "//ign_bazel:build_defs.bzl",
+    "IGNITION_FEATURES",
+    "IGNITION_ROOT",
+    "IGNITION_VISIBILITY",
+    "generate_include_header",
+    "ign_config_header",
+    "ign_export_header",
+)
+
+package(
+    default_visibility = IGNITION_VISIBILITY,
+    features = IGNITION_FEATURES,
+)
+
+licenses(["notice"])
+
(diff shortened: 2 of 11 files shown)


.github/ci-bionic/packages.apt
```diff
@@ -0,0 +1,11 @@
+pkg-config
```
chapulina, line 1: The file is already under `.github/ci`

https://github.com/ignitionrobotics/ign-common/blob/ign-common3/.github/ci/packages.apt

graphics/src/ColladaExporter.cc
```diff
@@ -41,24 +41,10 @@ static void LogTinyXml2DocumentError(
   {
     warning += std::to_string(_doc.ErrorID()) + "): ";
 
-#ifdef TINYXML2_MAJOR_VERSION_GE_6
```
chapulina, line 44: Have you checked the TinyXML2 versions on all our supported platforms?

chapulina (commented):
Same comment here about separating the Bazel addition from the other changes.

JShep1 (commented):
Looks good, but I'm getting a test failure in `PluginLoader` [here](https://github.com/ignitionrobotics/ign-common/blob/ign-common3/src/PluginLoader_TEST.cc#L55-L58).  Are you getting any failures on your end?

mjcarroll:
> Looks good, but I'm getting a test failure in PluginLoader here. Are you getting any failures on your end?

Fixed

graphics/src/ColladaExporter.cc
mjcarroll: I reverted this.

JShep1 (changes requested):
All is working for me locally except for `av`, I'm getting some error messages on Bionic for missing headers in `ffmpeg`: `hdr_dynamic_metadata.h` and `ac3_parser.h`.  I was thinking this could possibly be a versioning thing, I searched `ffmpeg` in my `/usr/include` directory and couldn't find those two files even though I'm fully updated locally.

Additionally, I'm getting a build failure with `colcon`, but changing `test/util.hh` to `test_util.hh` [here](https://github.com/ignitionrobotics/ign-common/blob/bazel-common3/test/integration/encoder_timing.cc#L21) fixes that.

graphics/src/ColladaExporter.cc
```diff
@@ -25,6 +25,11 @@
 
 #include "tinyxml2.h"
 
+#if TINYXML2_MAJOR_VERSION >= 6
```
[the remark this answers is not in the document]
JShep1: I replicated both above issues on my Bionic desktop and both.  Removing:
```
 - libavcodec/ac3_parser.h
 - libavcodec/adts_parser.h
 - libavutil/hdr_dynamic_metadata.h
 - libavutil/encryption_info.h
 - libavutil/hwcontext_mediacodec.h
 - libavutil/tx.h
```
from the file [here](https://github.com/ignitionrobotics/ign-bazel/blob/master/third_party/ffmpeg.BUILD) fixed the `ffmpeg` issue for me.  The second issue was also resolved by me defining that tinyxml2 variable.  All tests passing after I did those steps, again, not sure how we actually want to define that variable without hardcoding the version number.

chapulina (approved):
I only checked the non-Bazel parts and they look reasonable to me :+1:

PR #124, opened 2020-11-10
Disable AV integration test when component not available

mjcarroll:
Disable the encoder integration test, which depends on the `av` component, which is currently not building on Windows.

I originally wanted to check for the `av` target first via:

```
if (NOT TARGET ${PROJECT_LIBRARY_TARGET_NAME}-av)
  list(REMOVE_ITEM tests encoder_timing.cc)
endif()
```

But because integration tests are configured _before_ components, those targets don't exist yet.  I'm not sure if this could be considered a bug in ign-cmake.

### DIFF
--- a/test/integration/CMakeLists.txt
+++ b/test/integration/CMakeLists.txt
@@ -6,7 +6,7 @@ ign_get_sources(tests)
 # FIXME the mesh test does not work
 list(REMOVE_ITEM tests mesh.cc)
 
-if (WIN32)
+if (${SKIP_av})
   list(REMOVE_ITEM tests encoder_timing.cc)
 endif()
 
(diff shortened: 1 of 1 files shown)


traversaro:
Just a random idea: could it make sense to check for the `SKIP_av` variable (see https://github.com/ignitionrobotics/ign-cmake/blob/c8a47188f70b223ec4f26e3387321fca7d2ea57f/cmake/IgnUtils.cmake#L227) for detecting if the av component is configured or not? Because it may be possible that the av component is disabled (either explicitly or because there are missing dependencies) also on Linux or macOS, and ideally it would be nice to be able to compile the test suite also in those cases.

mjcarroll:
> Just a random idea: could it make sense to check for the SKIP_av variable

Ah! That's a great idea, I had forgotten that existed.  It also is evaluated before the tests are added, so the ordering problem will be gone.

jennuine (approved):
LGTM! Thanks!

mjcarroll:
I'm going to go ahead and merge this into the other PR, and then sort out the buildfarm issues over there.

PR #127, opened 2020-11-11
Fix IGN_HOMEDIR on Windows

peci1:
HOMEPATH alone is a pretty wicked concept. For its use to make sense, it has to be joined with HOMEDRIVE, because HOMEPATH is a path relative to HOMEDRIVE. So when you lookup just env("HOMEPATH") and use it as a file path, it gets resolved relative to any random drive letter that currently happens to be in CWD. I guess this could not be the inteded behavior.

USERPROFILE, on the other hand, contains the full absolute path to the user's home directory.

I looked through most libraries in github.com/ignitionrobotics, and none seemed to use IGN_HOMEDIR in a way that would assume the previous (wrong) value on Windows.

I don't consider this change a API or ABI break (the only thing it could break is already broken code :) ). However, as it is a change of a macro definition, it would need a re-release of all downstream libraries. But only on Windows :) And if I'm not mistaken, there are no binary builds for Windows yet, so that shouldn't be a problem (workspaces of people building from source would get correctly rebuilt automatically).

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -30,6 +30,8 @@ release will remove the deprecated code.
 
 1. Requires c++17.
 
+1. (New in 3.8.0) On Windows, the value of C++ macro `IGN_HOMEDIR` changed from `HOMEPATH` to `USERPROFILE`. It is usually used to read the path to the user's home from environment. The old value pointed to a path relative to the (a) current drive letter as reported by `pwd`, not the system drive letter. The new value correctly points to an environment variable that contains the full absolute path to the user's profile. If the code did not use the macro in some unexpected way, the new behavior should work either the same or even better (it would work even when the current directory is on a non-system drive). If the code relied on this value to be relative to the current drive letter, it needs to be changed to use `HOMEPATH` directly.
+
 ## Ignition Common 1.X to 2.X
 
 ### Modifications
--- a/include/ignition/common/Util.hh
+++ b/include/ignition/common/Util.hh
@@ -32,7 +32,7 @@
 // Defines
 
 #ifdef _WIN32
-# define IGN_HOMEDIR "HOMEPATH"
+# define IGN_HOMEDIR "USERPROFILE"
 #else
 # define IGN_HOMEDIR "HOME"
 #endif


traversaro:
> And if I'm not mistaken, there are no binary builds for Windows yet, so that shouldn't be a problem (workspaces of people building from source would get correctly rebuilt automatically).

Even if not maintained directly by the OR's Simulation team, we distribute binaries for ignition-common at conda-forge ( https://github.com/conda-forge/libignition-common-feedstock ) but at the first glance the only other library that uses `IGN_HOMEDIR` is fuel-tools ( https://github.com/conda-forge/libignition-fuel-tools-feedstock ), and in any case the old and the new behavior in normal use cases seems to be quite compatible, so I think  this change make sense. cc @Tobias-Fischer @wolfv

mjcarroll:
It is also used in `sdformat`, which doesn't make use of ign-common, in at least one case: https://github.com/osrf/sdformat/blob/00937472e6c3f70b0efee04341f48b640d882eae/src/Console.cc#L52-L54

chapulina:
It sounds like this is safe, but I think at least a note in the migration guide would be good to have.

peci1:
Okay, I've added a mention in migration.

mjcarroll (approved):

j-rivero:
> Okay, I've added a mention in migration.

Thanks. The explanation is good enough for me, even to ship it inside a minor bump instead of waiting for a new major. If I understand correctly, the change should be transparent for most of the cases and the few corner cases that could appear are probably buggy or can be improved.

j-rivero (approved):

peci1:
Yes, I wouldn't fear releasing this in a minor version. The potential for breaking downstream code is marginal in my opinion. 

Someone would have to intentionally count on the previous behavior to write code that would be broken by this PR. But as nobody raised an issue or question (AFAIK), I guess nobody knew about the behavior.

Not sure whether e.g. @seanyen doesn't have some experience with Ignition on Windows.

PR #128, opened 2020-11-11
Improve Windows support

peci1:
This is a collection of little fixes here and there that result in ability to enable most tests on Windows.

I hope none of these changes affect ABI backwards-compatibility. API-wise, I changed the value of macro `IGN_TMP_DIR` in `test/utils.hh`, but as it is a test-only file, I hope that's okay.

It has two questionable points worth a debate:

1. commit 9d06c52 switches the Console on Windows to a newer API similar to the UNIX character escape sequences. It isn't necessary, but the old API got deprecated by Microsoft and doesn't e.g. work on remote consoles.
1. commit 187a05d switches `joinPaths` on Windows to use a native function (`PathCombineA`). There was a todo about this, so I did it. However, I'm not sure if the result is better :)

One point worth more thinking (outside the scope of this PR): All Windows API functions in ign-common use the old ASCII versions of the API. That means paths and other strings passed between the library and API depend on the current codepage. I'd consider it a good move to migrate them to the wide-string versions which use UTF-16. But that would need some support infrastructure to convert between UTF-8 and UTF-16 strings and could come with some performance costs.


traversaro:
> One point worth more thinking (outside the scope of this PR): All Windows API functions in ign-common use the old ASCII versions of the API. That means paths and other strings passed between the library and API depend on the current codepage. I'd consider it a good move to migrate them to the wide-string versions which use UTF-16. But that would need some support infrastructure to convert between UTF-8 and UTF-16 strings and could come with some performance costs.

Could it make sense that to support wide-string path on Windows ignition projects switch to use the [C++17's std::filesystem API](https://en.cppreference.com/w/cpp/filesystem) instead of the ign-common Filesystem class? The main blocking issue on using that is that Xcode C++ SDK on macOS did not supported it until Xcode 11 with macOS 10.15 so to be able to use it it may pass some time.

peci1:
>  switch to use the [C++17's std::filesystem API](https://en.cppreference.com/w/cpp/filesystem) instead of the ign-common Filesystem class?

That would definitely make sense. I thought ign-common requires C++17, thus it also requires std::filesystem support. So you say that in Xcode 10 you get most of C++17, but not std::filesystem? Isn't there at least std::experimental::filesystem? I've made a CMake/preprocessor shim that loads either std::filesystem or std::experimental filesystem and exposes a macro telling you which of the two is available. Example usage is:

    // CMake:
    find_package(Filesystem REQUIRED COMPONENTS Final Experimental Boost)
    target_link_libraries(${PROJECT_NAME} LINK_PRIVATE std::filesystem)
    // C++
    #include CXX_FILESYSTEM_INCLUDE
    namespace fs = CXX_FILESYSTEM_NAMESPACE;

traversaro:
> So you say that in Xcode 10 you get most of C++17, but not std::filesystem?

For sure there is no std::filesystem in Xcode 10, I am afraid that also std::experimental::filesystem was not there (even if it was in libc++) because was removed as non-ABI stable (see https://stackoverflow.com/questions/42633477/macos-clang-c17-filesystem-header-not-found). In general there are some aspect of C++17, especially regarding the C++ Standard Library, that unfortunately are not really supported in the major C++ Standard Libraries implementation, another example are `to_chars` support for floating point variables.

include/ignition/common/Filesystem.hh
```diff
@@ -154,6 +156,11 @@ namespace ignition
     /// \return Name of the current directory
     std::string IGNITION_COMMON_VISIBLE cwd();
 
+    /// \brief Change current working directory to _dir.
+    /// \param[in] _dir The directory to change to.
+    /// \return Whether the operation succeeded.
+    bool IGNITION_COMMON_VISIBLE chdir(const std::string& _dir);
```
chapulina, line 162: Style nit

```suggestion
    bool IGNITION_COMMON_VISIBLE chdir(const std::string &_dir);
```

src/Console.cc
```diff
@@ -30,12 +30,11 @@ using namespace common;
 FileLogger ignition::common::Console::log("");
 
 #ifdef _WIN32
```
chapulina, line 32: It looks like this `ifdef` isn't needed anymore

src/Console.cc
```diff
@@ -252,6 +266,17 @@ void FileLogger::Init(const std::string &_directory,
   // (*buf->stream) << PROJECT_VERSION_HEADER << std::endl;
 }
 
+void FileLogger::Close()
```
chapulina, line 269: Style nit

```suggestion
/////////////////////////////////////////////////
void FileLogger::Close()
```

src/Console_TEST.cc
```diff
@@ -25,7 +25,8 @@
 
 #include "test/util.hh"
 
-#ifndef _WIN32
+namespace igncmn = ignition::common;
```
chapulina, line 28: These abbreviations can make code harder to find and search-and-replace later on. 

How about `using namespace ignition::common;`?

src/FilesystemBoost.cc
```diff
@@ -416,6 +422,11 @@ namespace ignition
       }
     }
 
+    bool chdir(const std::string& _dir)
```
chapulina, line 425:
```suggestion
    //////////////////////////////////////////////////
    bool chdir(const std::string& _dir)
```

chapulina (commented):
Thanks for the PR! I just took a quick high-level look and it's looking reasonable. CI is happy too :+1:

src/Console.cc
peci1: Removed, but added warning that it is not a rule that the values are the same on win and linux.

src/Console_TEST.cc
peci1: Yeah, good idea. I've just found `igncmn` in more places, so I thought it is kind of standard.

peci1:
Thank you for the nits. I've addressed all of them.

mjcarroll (approved):

mjcarroll:
This is great! Thanks for the contribution.

mjcarroll:
This is going in.  @peci1 if you feel that the UTF-8/UTF-16 issue warrants additional followup, I think it would be appropriate to open an issue so that it doesn't fall through the cracks.

peci1:
Thanks, issue created: #129.

PR #131, opened 2020-12-01
Support multiple texture coordinate sets

iche033:
The SubMesh class have always supported only texture coordinate sets per submesh. This works for most cases but when it comes do adding lightmaps to meshes, they usually occupy a different texture coordinate set, i.e. they are mapped onto the mesh in a different way than the typical albedo and normal maps. This pull request adds new APIs that enable users to add texture coordinates for different sets in the same submesh. I've also updated the ColladaLoader to load multiple texture coordinate sets.

### DIFF
--- a/graphics/include/ignition/common/SubMesh.hh
+++ b/graphics/include/ignition/common/SubMesh.hh
@@ -115,13 +115,21 @@ namespace ignition
       /// \param[in] _z Position along z
       public: void AddNormal(const double _x, const double _y, const double _z);
 
-      /// \brief Add a texture coord to the mesh
+      /// \brief Add a texture coord to the mesh. If multiple texture
+      /// coordinate sets exist, this function adds it to the first texture
+      /// coordinate set in the submesh. If no previous texture coordinates
+      /// exist, it is added to set 0.
       /// \param[in] _u Position along u
       /// \param[in] _v Position along v
+      /// \sa AddTexCoordBySet
       public: void AddTexCoord(const double _u, const double _v);
 
-      /// \brief Add a texture coordinate to the mesh
+      /// \brief Add a texture coordinate to the mesh. If multiple texture
+      /// coordinate sets exist, this function adds it to the first texture
+      /// coordinate set in the submesh. If no previous texture coordinates
+      /// exist, it is added to set 0.
       /// \param[in] _uv The texture coordinate
+      /// \sa AddTexCoordBySet
       public: void AddTexCoord(const ignition::math::Vector2d &_uv);
 
       /// \brief Add a texture coord to a texture coordinate set of the mesh
@@ -190,9 +198,12 @@ namespace ignition
                   unsigned int _index,
                   unsigned int _setIndex) const;
 
-      /// \brief Set a texture coordinate
+      /// \brief Set a texture coordinate. If multiple texture
+      /// coordinate sets exist, this function sets the texture
+      /// coordinate in the first texture coordinate set in the submesh.
       /// \param[in] _index Index of the texture coordinate that will be set.
       /// \param[in] _uv The new texture coordinate
+      /// \sa SetTexCoordBySet
       public: void SetTexCoord(const unsigned int _index,
                                const ignition::math::Vector2d &_uv);
 
@@ -244,8 +255,13 @@ namespace ignition
       /// \return The number of indices.
       public: unsigned int IndexCount() const;
 
-      /// \brief Return the number of texture coordinates
+      /// \brief Return the number of texture coordinates. If multiple
+      /// texture coordinate sets exist, this function checks the first
+      /// texture coordinate set in the submesh, which by default is set 0,
+      /// unless AddTexCoordBySet is called with a different set index number
+      /// the first time a texture coordinate is added.
       /// \return The number of texture coordinates.
+      /// \sa TexCoordCountBySet
       public: unsigned int TexCoordCount() const;
 
       /// \brief Return the number of texture coordinates for a texture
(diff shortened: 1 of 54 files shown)


graphics/include/ignition/common/SubMesh.hh
```diff
@@ -124,6 +124,20 @@ namespace ignition
       /// \param[in] _uv The texture coordinate
       public: void AddTexCoord(const ignition::math::Vector2d &_uv);
```
chapulina, line 125: How about documenting what set the set-less APIs add to? I noticed it defaults to zero, but may be different if another set has started.

iche033: added more documentation in 228eb7c

graphics/src/ColladaLoader.cc
```diff
@@ -2034,8 +2036,14 @@ void ColladaLoaderPrivate::LoadPolylist(tinyxml2::XMLElement *_polylistXml,
     }
     else if (semantic == "TEXCOORD")
     {
-      this->LoadTexCoords(source, texcoords, texDupMap);
-      inputs[TEXCOORD].insert(ignition::math::parseInt(offset));
+      int offsetInt = ignition::math::parseInt(offset);
+      if (inputs[TEXCOORD].find(offsetInt) == inputs[TEXCOORD].end())
+      {
+        int set = ignition::math::parseInt(polylistInputXml->Attribute("set"));
```
chapulina, line 2042: We haven't been checking for the `set` attribute before, is there a chance we break some existing user that doesn't have `set`? Note that `parseInt` returns `NAN_I` if the string is empty.

graphics/src/ColladaLoader.cc
```diff
@@ -1992,15 +1993,16 @@ void ColladaLoaderPrivate::LoadPolylist(tinyxml2::XMLElement *_polylistXml,
...
+  std::map<unsigned int, unsigned int> texcoordsOffsetToSet;
 
   const unsigned int VERTEX = 0;
   const unsigned int NORMAL = 1;
   const unsigned int TEXCOORD = 2;
   unsigned int otherSemantics = TEXCOORD + 1;
 
   // look up table of position/normal/texcoord duplicate indices
-  std::map<unsigned int, unsigned int> texDupMap;
+  std::map<unsigned int, std::map<unsigned int, unsigned int>> texDupMap;
```
chapulina, line 2005: Do the new maps need to be ordered? I see there's a concept of "first" set for backwards compatibility, but maybe that can be worked around?

graphics/src/ColladaLoader.cc
```diff
@@ -2337,9 +2363,15 @@ void ColladaLoaderPrivate::LoadTriangles(tinyxml2::XMLElement *_trianglesXml,
...
     else if (semantic == "TEXCOORD")
     {
-      // we currently only support one set of UVs
-      this->LoadTexCoords(source, texcoords, texDupMap);
-      inputs[TEXCOORD].insert(ignition::math::parseInt(offset));
+      int offsetInt = ignition::math::parseInt(offset);
+      if (inputs[TEXCOORD].find(offsetInt) == inputs[TEXCOORD].end())
+      {
+        unsigned int set = ignition::math::parseInt(
+            trianglesInputXml->Attribute("set"));
```
chapulina, line 2370: Same comment about whether we need to support empty `set` attributes

graphics/src/ColladaLoader.cc
```diff
@@ -2454,13 +2486,36 @@ void ColladaLoaderPrivate::LoadTriangles(tinyxml2::XMLElement *_trianglesXml,
...
-              texEqual = true;
+            // unsigned int offset = *inputs[TEXCOORD].begin();
+            // unsigned int remappedTexcoordIndex =
+            //     values.at(offset);
+            // int set = texcoordsOffsetToSet[offset];
+            // auto &texDupMapSet = texDupMap[set];
+            // auto texDupMapSetIt = texDupMapSet.find(remappedTexcoordIndex);
+            // if (texDupMapSetIt != texDupMapSet.end())
+            //   remappedTexcoordIndex = texDupMapSetIt->second;
+            // texEqual = iv.texcoordIndex[set] == remappedTexcoordIndex;
```
chapulina, line 2497: Can the comments be removed?

graphics/src/SubMesh.cc
```diff
@@ -220,7 +243,17 @@ bool SubMesh::HasNormal(const unsigned int _index) const
...
+    return false;
+  unsigned firstSetIndex = this->dataPtr->texCoords.begin()->first;
+  return this->HasTexCoordBySet(_index, firstSetIndex);
+}
+
+//////////////////////////////////////////////////
+bool SubMesh::HasTexCoordBySet(unsigned int _index,
+    unsigned int _setIndex) const
+{
+  return _index < this->dataPtr->texCoords[_setIndex].size();
```
chapulina, line 256: Should we check if `texCoords` has `setIndex` first?

There are more instances below.

graphics/src/ColladaLoader.cc
iche033: added check in 50c0e2e

graphics/src/ColladaLoader.cc
iche033: added check. 50c0e2e

graphics/src/ColladaLoader.cc
iche033: removed. 50c0e2e

graphics/src/SubMesh.cc
iche033: yep, added checks. 50c0e2e

graphics/src/ColladaLoader.cc
iche033: Nope, I updated the new outer map to be unordered.

about the "first" uv, I was mainly thinking of a use case where some one wrote a mesh loader that can selectively load any uv set in the mesh, which end up with code like:

e.g. 
```
int set = 1
AddTextureCoordinateBySet(u0, v0, set)
AddTextureCoordinateBySet(u1, v1, set)
...
auto count = TextureCoordinateCount()
```
In this case, I just return the number of texture coordinates that are in the only set in the submesh. We could decide that `count` in the above code will just be 0, and we can get rid of `first*`. I'm open two either option.

chapulina: I don't have a strong opinion here... I'd also be ok returning the first element of the unordered map, which may be undefined if there are multiple sets... Maybe we can print a warning if they use this function and there's more than one set? I'm fine either way.

iche033: ok I tried the simple case of just defaulting the first set to index 0u but ran into a test failure due to a very similar scenario described in my previous comment, i.e. there was a mesh with only one tex coord set with index=1. I ran into another test failure when changing to use `unordered_map` in SubMesh as first tex coord set (indicated by `begin()`) changed when a new set is added. In the end I just added warning messages for the case where there are multiple uv sets in the submesh. bcdac31

chapulina (approved):
Nice! Lets get this in!

PR #100, opened 2020-09-28
Add transformation matrix to ColladaExport

gonzodepedro:
Signed-off-by: Gonzalo de Pedro <gonzalo@depedro.com.ar>

This adds transformation matrixes to nodes con collada files when exporting

Connects to ignitionrobotics/ign-gazebo#307

PR #133, opened 2020-12-03
Fix ColladaExport test and add pathJoins

mjcarroll:
Follow up fix to #100


mjcarroll:
Windows test failures are unrelated to this patch.

chapulina (approved):
Nice cleanup! It looks like some Windows failures have also been recently introduced, but I think the fix doesn't need to be bundled here.

PR #134, opened 2020-12-04
Make ignstrtok private to Util.cc

mahiuchun:
Otherwise we might get ODR issue with `ign-msgs/src/Utility.cc`.


chapulina (approved):
Thanks! LGTM if CI comes back happy.

mjcarroll (approved):

mjcarroll:
Windows failures also unrelated here.

PR #135, opened 2020-12-04
Fix Windows AV CI warnings and test failures

mjcarroll:
I believe that these were introduced by changes in the Windows hosts, allowing the `av` component to be built and tested, so not necessarily a regression.


chapulina (approved):
> I believe that these were introduced by changes in the Windows hosts, allowing the av component to be built and tested

Totally! Nice! It looks like we can close https://github.com/ignitionrobotics/ign-common/issues/122 :tada:

scpeters (approved):

PR #136, opened 2020-12-05
Fix memory leaks.

mahiuchun:


ahcorde (approved):

mjcarroll (approved):

mjcarroll:
Window failure is unrelated.

iche033:
Found that the actor world in ign-gazebo is segfaulting after these changes.. The changes here look fine. It could be that we still have pointers pointing to these old data structures after they are being deleted. I think this could happen if we think we are copying these objects but we're actually just copying pointers. We may have to add copy constructors / assignment operators in these classes

cc @adlarkin

PR #138, opened 2020-12-12
Fix a memory leak.

mahiuchun:


graphics/src/GTSMeshUtils_TEST.cc
```diff
@@ -15,6 +15,7 @@
  *
 */
 
+#include <memory>
```
mjcarroll, line 18: Linter is flagging the include order here, care to fix?

mjcarroll (approved):

chapulina (approved):

PR #142, opened 2020-12-22
Added render order to material class

ahcorde:
This PR is part of the effort to avoid z-fighting between coplanar meshes. 

Related with https://github.com/ignitionrobotics/ign-rendering/pull/188

### DIFF
--- a/graphics/include/ignition/common/Material.hh
+++ b/graphics/include/ignition/common/Material.hh
@@ -177,6 +177,15 @@ namespace ignition
       /// \return The enable two sided rendering value
       public: bool TwoSidedEnabled() const;
 
+      /// \brief Set the render order. The higher value will be rendered on top
+      /// of the other coplanar polygons.
+      /// \param[in] _renderOrder The render order value
+      public: void SetRenderOrder(float _renderOrder);
+
+      /// \brief Get the render order
+      /// \return The render order value
+      public: float RenderOrder() const;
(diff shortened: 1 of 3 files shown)


graphics/src/Material.cc
```diff
@@ -119,6 +123,7 @@ Material::Material(const math::Color &_clr)
     std::to_string(this->dataPtr->counter++);
   this->dataPtr->blendMode = REPLACE;
   this->dataPtr->shadeMode = GOURAUD;
+  this->dataPtr->renderOrder = 0;
```
chapulina, line 126: This is already set during the declaration, do you need to set it again in the constructors?

graphics/include/ignition/common/Material.hh
```diff
@@ -177,6 +177,14 @@ namespace ignition
       /// \return The enable two sided rendering value
       public: bool TwoSidedEnabled() const;
 
+      /// \brief Set the render order
+      /// \param[in] _t The render order value
+      public: void SetRenderOrder(int _renderOrder);
+
+      /// \brief Get the render order
```
chapulina, line 184: Could you document the default value? Also, I'm wondering what a negative value means, is that allowed?

chapulina (commented):
Looks good to me, I just have some questions.

chapulina (approved):

PR #145, opened 2020-12-23
3 ➡️ 4

mjcarroll:
Forward port at version 3.7.0.  This is to make sure that all of the commits from the `ignition-common3` branch are visible in the history of `main`, or else the diffs get a bit awkward.


.github/CODEOWNERS
```diff
@@ -2,3 +2,4 @@
 # https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
 
 *                     @mjcarroll
+tutorials/*           @maryaB-osr
```
chapulina, line 5: I think this is a bad merge, the line should be removed as of https://github.com/ignitionrobotics/ign-common/pull/88/

chapulina (approved):
Besides the codeowner line, LGTM.

I'll just cross-reference the rebased PR here: https://github.com/ignitionrobotics/ign-common/pull/104

.github/CODEOWNERS
mjcarroll: Fixed in the merge commit and force-pushed.

chapulina (approved):

PR #146, opened 2020-12-24
Add Windows installation; move installation in README to tutorial

mabelzhang:
Partially addresses https://github.com/ignitionrobotics/docs/issues/117 and https://github.com/ignitionrobotics/docs/issues/14#issuecomment-750431388 .

- Adds Windows binary and from source install instructions for https://github.com/ignitionrobotics/docs/issues/117 .
   For some reason, I didn't encounter the compilation error @Jshep1 mentioned here https://github.com/ignitionrobotics/docs/issues/96#issuecomment-742096017 .
   My steps are slightly different, as necessitated by these per-package installation tutorials.
   John built everything from source using `colcon`.
   I did `conda-forge` binary install for `ign-cmake` and `ign-math`, and then pure `cmake`.
   Not sure if that makes the difference or I'm missing something.

- Moves installation in README to a new installation tutorial page for https://github.com/ignitionrobotics/docs/issues/14#issuecomment-750431388 .
   TODO: The URL in README currently links to `ign-common` tutorials main page, since the installation tutorial webpage doesn't exist yet. Somebody should change the link after this PR is deployed. Trivial fix of the URL in README line 63.

### DIFF
--- a/tutorials/install.md
+++ b/tutorials/install.md
@@ -5,11 +5,11 @@ Next Tutorial: \ref profiler
 These instructions are for installing only Ignition Common.
 If you're interested in using all the Ignition libraries, check out this [Ignition installation](https://ignitionrobotics.org/docs/latest/install).
 
-We recommend following the Binary Install instructions to get up and running as quickly and painlessly as possible.
+We recommend following the Binary Installation instructions to get up and running as quickly and painlessly as possible.
 
-The Source Install instructions should be used if you need the very latest software improvements, you need to modify the code, or you plan to make a contribution.
+The Source Installation instructions should be used if you need the very latest software improvements, you need to modify the code, or you plan to make a contribution.
 
-# Binary Install
+# Binary Installation
 
 ## Ubuntu
 
@@ -40,7 +40,7 @@ conda install libignition-common<#> --channel conda-forge
 Be sure to replace `<#>` with a number value, such as 2 or 3, depending on
 which version you need.
 
-# Source Install
+# Source Installation
 
 Source installation can be performed by first installing the necessary
 prerequisites followed by building from source.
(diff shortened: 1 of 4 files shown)


traversaro:
>  Not sure if that makes the difference or I'm missing something.

The difference is that the ignition-cmake available in conda-forge contains a patch to fix the GTS-problem reported by @JShep1 : 
* https://github.com/conda-forge/libignition-cmake0-feedstock/blob/master/recipe/FindGTS.cmake.patch
* https://github.com/conda-forge/libignition-cmake0-feedstock/blob/master/recipe/meta.yaml#L15

So the problem is still present if ignition-cmake is compiled from source when building all the ignition projects together with colcon, but not present if ignition-cmake is installed via conda-forge.

tutorials/install.md
```diff
@@ -0,0 +1,150 @@
...
+1. Clone the repository
+
+    ```
+    git clone https://github.com/ignitionrobotics/ign-common
+    ```
+
+1. Configure and build
+
+    ```
+    cd ign-common; mkdir build;cd build; cmake ..;  make
```
JShep1, line 118:
```suggestion
    cd ign-common
    mkdir build
    cd build
    cmake ..
    make
```

JShep1 (commented):
Small consistency nit but LGMT otherwise

mabelzhang:
Same comment about formatting as https://github.com/ignitionrobotics/ign-math/pull/185#issuecomment-756463732
fb6323a tries to fix it.

I'm not sure if that fixes it though. Looking at the profiler tutorial, Troubleshoot section, bullets 2 and 4 have code blocks, one with a preceding empty line and one without, both rendered incorrectly. So the empty line isn't the answer. Do you think changing from 4 spaces to 2 spaces helps??? Notably, the `{.sh}` doesn't help either, because the ign-physics tutorial that renders correctly doesn't have that, these incorrect ones do.
https://ignitionrobotics.org/api/common/3.9/profiler.html

mabelzhang:
Formatting has been fixed, given the tip in the other PR.
So yeah, 2 spaces made a difference!

JShep1 (approved):

chapulina (approved):

PR #147, opened 2020-12-28
Fix memory leak in Pbr.cc

w-michal:
Caught with Address Sanitizer. Acording to ASAN, Pbr.MoveCopy test
would allocate PbrPrivate on Pbr_TEST.cc:177 and leak on
Pbr_TEST.cc:178.


chapulina (approved):
Thanks for catching and for the fix! Mind signing all commits?

PR #149, opened 2020-12-29
Disable failing VideoEncoder and Audioecoder tests on Windows

chapulina:
An attempt at fixing https://github.com/ignitionrobotics/ign-common/issues/148.

The test changes are minimal, I recommend the view ignoring whitespace changes:

* Make test verbose
* Create `video` within a scope, so we can check that it deletes all temp files on destruction
* Output file paths in case the tests fail

The change that I hope will fix it is the use of `common::removeFile` instead of `std::remove`.

### DIFF
--- a/av/src/AudioDecoder_TEST.cc
+++ b/av/src/AudioDecoder_TEST.cc
@@ -17,6 +17,7 @@
 #include <gtest/gtest.h>
 
 #include <ignition/common/AudioDecoder.hh>
+#include <ignition/utilities/ExtraTestMacros.hh>
 
 #include "test/util.hh"
 #include "test_config.h"
@@ -88,7 +89,7 @@ TEST(AudioDecoder, NoCodec)
 }
 
 /////////////////////////////////////////////////
-TEST(AudioDecoder, CheerFile)
+TEST(AudioDecoder, IGN_UTILS_TEST_DISABLED_ON_WIN32(CheerFile))
 {
   common::AudioDecoder audio;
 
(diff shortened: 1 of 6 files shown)


mjcarroll:

JShep1:
@osrf-jenkins run tests please

JShep1:
Seems like the test is still failing on Windows, re-running tests for now, and then will look more into it after

chapulina:
Yeah this doesn't fix the issue... For some reason the temp file can't be deleted on Windows... If no one has a good idea for a fix, I suggest we comment out the test on Windows for now

mjcarroll:
> If no one has a good idea for a fix, I suggest we comment out the test on Windows for now

No suggestion on fixing, but I would prefer to use the macro: https://github.com/ignitionrobotics/ign-cmake/blob/ign-cmake2/include/ignition/utilities/ExtraTestMacros.hh#L28

chapulina:
> I would prefer to use the macro

In general I agree, but in this case, the only offending expectations were at the end of the test, so I suggest we just skip those and still run the rest of the test on Windows. See 32421c9

chapulina:
Also disabled #156

caguero (commented):
There's a problem on Windows:

```
C:\Jenkins\workspace\ignition_common-ci-pr_any-windows7-amd64\ws\ign-common\av\src\AudioDecoder_TEST.cc(20): fatal error C1083: Cannot open include file: 'ignition/utilities/ExtraTestMacros.hh': No such file or directory
```

chapulina:
> There's a problem on Windows:

Ahhh not anymore!

caguero (approved):

Issue #91, opened 2020-09-01
COLLADA Cube Mesh not Colored


jaelrod:
Loading a COLLADA mesh containing a cube with colored surface results in a white cube instead.
[mesh.dae.txt](https://github.com/ignitionrobotics/ign-common/files/5157951/mesh.dae.txt)

COLLADA loaded in Blender:
![image](https://user-images.githubusercontent.com/64549955/91887701-f6724480-ec3f-11ea-89a6-5da4832c0241.png)

Same COLLADA loaded in Ignition Gazebo v3.2.0:
![image](https://user-images.githubusercontent.com/64549955/91887735-04c06080-ec40-11ea-813c-93cd104a3b97.png)

ahcorde:
Hi @jaelrod,

I have loaded your mesh into Blender and imported again to Collada and I'm able to see the color:

![Selección_088](https://user-images.githubusercontent.com/1933907/104439676-aa41f400-5591-11eb-9db1-81e4dd341232.png)

A quick look to your dae file looks like your triangles have no material.  You include the material in the triangles:
Replace this line:

```xml
        <triangles count="12">
```
with:

```xml
        <triangles material="material-cube" count="12">
```

Please close the issue if the problem is resolved.

jaelrod:
Why would <triangles/> be the place to define the material. If I have a geometries library, I would fundamentally prefer that to be decoupled from the material applied on top of the geometry, which I believe is the inspiration behind the COLLADA specification itself separating the definition of geometry vs. materials. If I want to have a blue cube like this and a cube that is similarly red in hue, I should be able to accomplish that with simply applying a different, red material from library_materials onto the same geometry from library_geometries.

In light of that, I think my application of the material makes more sense where it is in the .dae.txt that I uploaded, in the instantiation of the mesh under the visual_scene, rather than to the triangles themselves:
```
  <library_visual_scenes>
    <visual_scene id="Scene" name="Scene">
      <node>
        <matrix sid="transform">0.35 0 0 0 0 0.4 0 0 0 0 1 0 0 0 0 1</matrix>
        <instance_geometry url="#mesh-cube">
          <bind_material>
            <technique_common>
              <instance_material symbol="material" target="#material-cube"/>
            </technique_common>
          </bind_material>
        </instance_geometry>
      </node>
    </visual_scene>
  </library_visual_scenes>
```

[...]

PR #151, opened 2021-01-15
Associate library materials effect with meshes

ahcorde:
This PR is related with this issue https://github.com/ignitionrobotics/ign-common/issues/91#issuecomment-759632701

The material could be defined in Collada like this:

```xml
  <library_materials>
    <material id="material-cube">
      <instance_effect url="#effect-cube"/>
    </material>
  </library_materials>
```

This kind of material definition were never associated with the mesh. This PR try to fix this issue.


mjcarroll (approved):

chapulina:
Would it be possible to add a test? Maybe using the mesh provided in #91 ?

ahcorde:
+1 for the test. Added here https://github.com/ignitionrobotics/ign-common/pull/151/commits/188d06d0ee5d1b9ed3d3afcd2576e268606e8cec

graphics/src/ColladaLoader_TEST.cc
```diff
@@ -51,6 +51,34 @@ TEST_F(ColladaLoader, LoadBox)
   EXPECT_STREQ("Cube", mesh->SubMeshByIndex(0).lock()->Name().c_str());
 }
 
+/////////////////////////////////////////////////
+TEST_F(ColladaLoader, LoadRectangle)
+{
+  common::ColladaLoader loader;
+  common::Mesh *mesh = loader.Load(
+      std::string(PROJECT_SOURCE_PATH) + "/test/data/rectangle.dae");
```
chapulina, line 59: Let's use `common::joinPaths` for cross-platform support

graphics/src/ColladaLoader_TEST.cc
```diff
@@ -51,6 +51,34 @@ TEST_F(ColladaLoader, LoadBox)
...
+  EXPECT_EQ(35u, mesh->VertexCount());
+  EXPECT_EQ(35u, mesh->NormalCount());
+  EXPECT_EQ(36u, mesh->IndexCount());
+  EXPECT_EQ(0u, mesh->TexCoordCount());
+  EXPECT_EQ(1u, mesh->SubMeshCount());
+  EXPECT_EQ(1u, mesh->MaterialCount());
+
+  common::MaterialPtr material = mesh->MaterialByIndex(0);
+  EXPECT_NE(nullptr, material);
+  EXPECT_EQ(math::Color(0.008957128, 0.00619848, 0.64, 1),
```
chapulina, line 74: To fix Windows warning

```suggestion
  EXPECT_EQ(math::Color(0.008957128f, 0.00619848f, 0.64f, 1.0f),
```

The same for all colors below.

graphics/src/ColladaLoader_TEST.cc
```diff
@@ -51,6 +51,34 @@ TEST_F(ColladaLoader, LoadBox)
...
+TEST_F(ColladaLoader, LoadRectangle)
+{
+  common::ColladaLoader loader;
+  common::Mesh *mesh = loader.Load(
+      std::string(PROJECT_SOURCE_PATH) + "/test/data/rectangle.dae");
+
+  EXPECT_STREQ("unknown", mesh->Name().c_str());
+  EXPECT_EQ(ignition::math::Vector3d(0.35, 0.4, 1), mesh->Max());
+  EXPECT_EQ(ignition::math::Vector3d(-0.35, -0.4, -1), mesh->Min());
+  // 36 vertices, 24 unique, 12 shared.
```
chapulina, line 64: This comment doesn't seem right

graphics/src/ColladaLoader.cc
```diff
@@ -549,6 +554,41 @@ void ColladaLoaderPrivate::LoadNode(tinyxml2::XMLElement *_elem, Mesh *_mesh,
...
+      iter = this->materialEffectMap.find(geomURL);
+      if (iter != this->materialEffectMap.end())
+        matStr = iter->second;
+      MaterialPtr mat = this->LoadMaterial(matStr);
+      matIndex = _mesh->IndexOfMaterial(mat.get());
+      if (matIndex < 0)
+      {
+        matIndex = _mesh->AddMaterial(mat);
+      }
+      if (matIndex < 0)
```
chapulina, line 574: Shouldn't this block be combined with the one above?

ahcorde: Not really, the first condional override the variable

https://github.com/ignitionrobotics/ign-common/blob/188d06d0ee5d1b9ed3d3afcd2576e268606e8cec/graphics/src/ColladaLoader.cc#L570-L577

chapulina: ah yes!

peci1:
I'm sorry, @ahcorde , but this PR seems to break more things than fix. X1 SubT robots are no longer blue, EXPLORER_DS1 has a weird red-ish color instead of carbon black...

image

Check the behavior of ColladaLoader with `~/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models/x1 config 1/18/meshes/chassis.dae`.

This file has one `<vertices>` and two `<triangles>`. Each triangle list defines a different symbol for material. However, after loading the mesh with ColladaLoader, both submeshes have material index 1 assigned. Without this PR, submesh 0 correctly has material index 0. Here's a simple (wrongly) failing test for me:

```C++
TEST_F(ColladaLoader, LoadX1)
{
  common::ColladaLoader loader;
  common::Mesh *mesh = loader.Load("/home/peci1/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models/x1 config 1/18/meshes/chassis.dae");

  ASSERT_EQ(2u, mesh->SubMeshCount());
  ASSERT_EQ(2u, mesh->MaterialCount());

  EXPECT_EQ(0u, mesh->SubMeshByIndex(0).lock()->MaterialIndex());
  EXPECT_EQ(1u, mesh->SubMeshByIndex(1).lock()->MaterialIndex());
}
```

Reading through #91 and Collada spec, I think the mesh submitted by @jaelrod is not what he expected it to be. The "triangles" section (p. 188) says:

> material   xs:NCName   Declares a symbol for a material. This symbol is bound to a material at the time of instantiation; see <instance_geometry> and <bind_material>. Optional. If not specified then the lighting and shading results are application defined. 

@jaelrod was concerned about reusability of the library geometries. But I think there was a misunderstanding on what does the `<triangles material="">` attribute actually mean. It does NOT specify the material to use. It specifies the symbol to which a material has to be bound in the `<instance_geometry>`. So it's not a "value", but a "variable name". I think there's nothing against naming the "variable" (material), as long as you can bind different materials to it (referencing the name in the `<instance_material symbol="">` attribute and specifying the selected material in `<instance_material target="">` attribute).

The mesh from #91 could work if the default value for the `<triangles material="">` were `"material"`. But the spec says that if this attribute is missing, shading is application-dependent. This is probably where application behavior starts to differ (and that's correct, as the spec says). To be compatible with Blender (which is generally a nice thing, I would say), this PR should be reverted and the change should be somewhere near

https://github.com/ignitionrobotics/ign-common/blob/abfecc8a262b32e3df7495f11b04fd38f5594f68/graphics/src/ColladaLoader.cc#L2009-L2029

I think this `if` could assume that there is a default value for the `material` attribute - `"material"`. It should just silence the warnings in case it tried to find a material and did not succeed.

PR #154, opened 2021-01-21
Added ellipsoid mesh

ahcorde:
Related with this issue https://github.com/ignitionrobotics/ign-common/issues/140

### DIFF
--- a/graphics/include/ignition/common/MeshManager.hh
+++ b/graphics/include/ignition/common/MeshManager.hh
@@ -154,6 +154,16 @@ namespace ignition
                                   const int _rings,
                                   const int _segments);
 
+      /// \brief Create a ellipsoid mesh
+      /// \param[in] _name the name of the new mesh
+      /// \param[in] _radii the three radius that define a ellipsoid
+      /// \param[in] _rings the number of circles along the height
+      /// \param[in] _segments the number of segment per circle
+      public: void CreateEllipsoid(const std::string &_name,
+                                   const ignition::math::Vector3d _radii,
+                                   const unsigned int _rings,
+                                   const unsigned int _segments);
+
       /// \brief Create a cone mesh
       /// \param[in] _name the name of the new mesh
       /// \param[in] _radius the radius of the cylinder in the x y plane
--- a/graphics/include/ignition/common/MeshManager.hh
+++ b/graphics/include/ignition/common/MeshManager.hh
@@ -160,7 +160,7 @@ namespace ignition
       /// \param[in] _rings the number of circles along the height
       /// \param[in] _segments the number of segment per circle
       public: void CreateEllipsoid(const std::string &_name,
-                                   const ignition::math::Vector3d _radii,
+                                   const ignition::math::Vector3d &_radii,
                                    const unsigned int _rings,
                                    const unsigned int _segments);
 
--- a/graphics/src/MeshManager.cc
+++ b/graphics/src/MeshManager.cc
@@ -836,30 +836,28 @@ void MeshManager::CreateEllipsoid(const std::string &_name,
   mesh->SetName(_name);
   this->dataPtr->meshes.insert(std::make_pair(_name, mesh));
 
-  ignition::math::Vector3d vert, norm;
+  ignition::math::Vector3d vert;
   SubMesh subMesh;
 
-  double umin = -M_PI / 2.0;
-  double umax = M_PI / 2.0;
-  double vmin = 0.0;
-  double vmax = 2.0 * M_PI;
+  const double umin = -M_PI / 2.0;
+  const double umax = M_PI / 2.0;
+  const double vmin = 0.0;
+  const double vmax = 2.0 * M_PI;
 
   unsigned int i, j;
   float theta, phi;
   float d_phi = (umax - umin) / (_rings - 1.0);
   float d_theta = (vmax - vmin) / (_segments - 1.0);
 
-  float c_theta, s_theta, c_phi, s_phi;
-
   for (i = 0, theta = vmin; i < _segments; ++i, theta += d_theta)
   {
-    c_theta = cos(theta);
-    s_theta = sin(theta);
+    const auto c_theta = cos(theta);
+    const auto s_theta = sin(theta);
 
     for (j = 0, phi = umin; j < _rings; ++j, phi += d_phi)
     {
-      c_phi = cos(phi);
-      s_phi = sin(phi);
+      const auto c_phi = cos(phi);
+      const auto s_phi = sin(phi);
 
       // Compute vertex
       subMesh.AddVertex(ignition::math::Vector3d(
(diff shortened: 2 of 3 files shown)


graphics/src/MeshManager.cc
```diff
@@ -822,6 +822,87 @@ void MeshManager::CreateCamera(const std::string &_name, float _scale)
...
+  mesh->SetName(_name);
+  this->dataPtr->meshes.insert(std::make_pair(_name, mesh));
+
+  ignition::math::Vector3d vert, norm;
+  SubMesh subMesh;
+
+  double umin = -M_PI / 2.0;
+  double umax = M_PI / 2.0;
+  double vmin = 0.0;
+  double vmax = 2.0 * M_PI;
```
mjcarroll, line 845:
```suggestion
  const double umin = -M_PI / 2.0;
  const double umax = M_PI / 2.0;
  const double vmin = 0.0;
  const double vmax = 2.0 * M_PI;
```

graphics/src/MeshManager.cc
```diff
@@ -822,6 +822,87 @@ void MeshManager::CreateCamera(const std::string &_name, float _scale)
...
+
+  unsigned int i, j;
+  float theta, phi;
+  float d_phi = (umax - umin) / (_rings - 1.0);
+  float d_theta = (vmax - vmin) / (_segments - 1.0);
```
mjcarroll, line 850: Should these be `double` since everything else is?

```suggestion
  double theta, phi;
  const double d_phi = (umax - umin) / (_rings - 1.0);
  const double d_theta = (vmax - vmin) / (_segments - 1.0);
```

graphics/src/MeshManager.cc
```diff
@@ -822,6 +822,87 @@ void MeshManager::CreateCamera(const std::string &_name, float _scale)
...
+
+  float c_theta, s_theta, c_phi, s_phi;
```
mjcarroll, line 852: I don't think these have to be declared outside the loop since they aren't used outside the loop

graphics/src/MeshManager.cc
```diff
@@ -822,6 +822,87 @@ void MeshManager::CreateCamera(const std::string &_name, float _scale)
...
+
+  for (i = 0, theta = vmin; i < _segments; ++i, theta += d_theta)
+  {
+    c_theta = cos(theta);
+    s_theta = sin(theta);
```
mjcarroll, line 857:
```suggestion
    const auto c_theta = cos(theta);
    const auto s_theta = sin(theta);
```

graphics/src/MeshManager.cc
```diff
@@ -822,6 +822,87 @@ void MeshManager::CreateCamera(const std::string &_name, float _scale)
...
+
+    for (j = 0, phi = umin; j < _rings; ++j, phi += d_phi)
+    {
+      c_phi = cos(phi);
+      s_phi = sin(phi);
```
mjcarroll, line 862:
```suggestion
      const auto c_phi = cos(phi);
      const auto s_phi = sin(phi);
```

graphics/src/MeshManager.cc
```diff
@@ -822,6 +822,87 @@ void MeshManager::CreateCamera(const std::string &_name, float _scale)
...
+        +(_radii.Y() * c_phi) * c_theta,
+        0.0);
+      ignition::math::Vector3d dv(
+        -_radii.X() * s_phi * c_theta,
+        -_radii.Y() * s_phi * s_theta,
+        _radii.Z() * c_phi);
+
+      ignition::math::Vector3d normal = du.Cross(dv);
+
+      subMesh.AddNormal(norm);
```
mjcarroll, line 882: `norm` is never set

mjcarroll (changes requested):
Some comments, I think we could also use some tests here.

graphics/include/ignition/common/MeshManager.hh
```diff
@@ -154,6 +154,16 @@ namespace ignition
                                   const int _rings,
                                   const int _segments);
 
+      /// \brief Create a ellipsoid mesh
+      /// \param[in] _name the name of the new mesh
+      /// \param[in] _radii the three radius that define a ellipsoid
+      /// \param[in] _rings the number of circles along the height
+      /// \param[in] _segments the number of segment per circle
+      public: void CreateEllipsoid(const std::string &_name,
+                                   const ignition::math::Vector3d _radii,
```
mjcarroll, line 163: I think `const &` is better here

```suggestion
                                   const ignition::math::Vector3d &_radii,
```

graphics/src/MeshManager.cc
```diff
@@ -822,6 +822,85 @@ void MeshManager::CreateCamera(const std::string &_name, float _scale)
   mesh->RecalculateNormals();
 }
 
+void MeshManager::CreateEllipsoid(const std::string &_name,
+                                  const ignition::math::Vector3d _radii,
```
mjcarroll, line 826: As above

```suggestion
                                  const ignition::math::Vector3d &_radii,
```

graphics/src/MeshManager.cc
mjcarroll: Still seeing a mismatch here. I don't think the precision should matter too much, but it is better to be consistent. 

If we are using `float`, then `cos` and `sin` need to be changed to `cosf` and `sinf`, and `ignition::math::Vector3d` should be `Vector3f` I believe.

If we are using double, these should be marked as such.

graphics/src/MeshManager.cc
```diff
@@ -822,6 +822,85 @@ void MeshManager::CreateCamera(const std::string &_name, float _scale)
...
+  }
+
+  Mesh *mesh = new Mesh();
+  mesh->SetName(_name);
+  this->dataPtr->meshes.insert(std::make_pair(_name, mesh));
+
+  ignition::math::Vector3d vert;
+  SubMesh subMesh;
+
+  const double umin = -M_PI / 2.0;
```
mjcarroll, line 842: Windows is dumb about the M_PI constant.

An easy way to get it (without changing defines and causing issues):

`const double pi = std::acos(-1);`

ahcorde: Looking at the rest of the methods `IGN_PI` is the solution.

graphics/src/MeshManager.cc
```diff
@@ -822,6 +822,85 @@ void MeshManager::CreateCamera(const std::string &_name, float _scale)
   mesh->RecalculateNormals();
 }
 
+void MeshManager::CreateEllipsoid(const std::string &_name,
+                                  const ignition::math::Vector3d &_radii,
+                                  const unsigned int _rings,
+                                  const unsigned int _segments)
+{
+  if (this->HasMesh(_name))
```
mjcarroll, line 830: Can we add a test that tries to add the same mesh name, just for completeness :)

mjcarroll (approved):
One small coverage nit, then LGTM.  Windows CI is unrelated to this change.

ahcorde:
VideoEncoder Test is failing on Windows but it's not related with this PR. Merging!

PR #160, opened 2021-01-29
Cleanup and add Implementation pointer to Animation classes

luca-della-vedova:
**Note! This PR introduces a dependency on ign-utils1 for the implementation pointer.**

There was a static map hack added in ign-common3 to add a member variable to the `Animation` class that didn't have an implementation pointer.
This PR removes the hack, as well as adding an implementation pointer to all the classes in the file and doing some scattered cleanups mainly:

- [Templating the CreateKeyFrame function](https://github.com/ignitionrobotics/ign-common/blob/luca/refactor_animation/graphics/src/Animation.cc#L214-L226). Since both `PoseAnimation` and `NumericAnimation` had the same code with the only difference being the type I thought it might be good to remove the duplicated code and create an `Animation` templated function.
- [Change from raw to shared pointers for splines](https://github.com/ignitionrobotics/ign-common/commit/ed4b953da5d68232843c354c6f6b78800fb121b4), just to be able to remove the custom destructor and make sure there is no issues in case of objects being copied and deleted. Tthere was no copy operator in either `PoseAnimation` or `Animation` and they all had raw pointer member variables that were deleted on destruction. If there happened to be a copy created and destructed, the memory would be freed and the original wouldn't be able to access it anymore.

Also I removed a few `const` from functions that were changing the state like `BuildInterpolationSplines()`. They were able to "get away with it" since there is a lot of mutable member variables but I can revert that if it's not a great idea.

Finally, I'm tempted to remove all the [unit tests](https://github.com/ignitionrobotics/ign-common/blob/luca/refactor_animation/graphics/src/Animation_TEST.cc#L192-L248) for copy / copy assignment / move operators for `TrajectoryInfo` since they are now just the default ones and they should be tested by the ImplPtr tests.


chapulina:
>This PR introduces a dependency on ign-utils1

I ticketed https://github.com/ignitionrobotics/ign-common/issues/161 to keep track of that. 

---

CI is all green right now because it's skipping the `graphics` component:

```
   -- 	Cannot build component [graphics] - Missing: ignition-utils1
```

chapulina:
@osrf-jenkins run tests again, thanks!

mjcarroll:
The Focal action is a performance test, so I think we're okay there.

The Windows failures are AV related, which I also believe are expected

The MacOS failure is a bit suspicious, `reinterpet_pointer_cast` is added in 17, which should be available on MacOS.  @scpeters do you have any insight here?

mjcarroll:

mjcarroll (approved):

PR #162, opened 2021-02-02
Fix saving image to PNG (swap red and blue)

chapulina:
We noticed while working on https://github.com/ignitionrobotics/ign-gui/pull/170 that the red and blue colors are swapped when saving an image. The `ign-sensors` [save_image example](https://github.com/ignitionrobotics/ign-sensors/blob/ign-sensors4/examples/save_image/main.cc) also saves the colors swapped, the sphere is supposed to be red and the box blue:

image

I did some debugging and noticed that the colors returned with `Data` and `Pixel` seem correct, but the saved image still looked swapped.

It looks like FreeImage uses BGR internally, and the `FreeImage_Save` function doesn't offer a way to choose what format to save in. So the approach I took was to swap red and blue using code I borrowed from [here](https://github.com/imazen/freeimage/blob/171744f4bc384f8872fd9437b23bbb5d54563253/Source/FreeImage/Conversion.cpp#L56-L78).

Keeping in draft until I can write some tests.


jennuine (approved):
LGTM, thanks for the help @chapulina!

PR #163, opened 2021-02-03
Add implementation pointer to ign-common classes

luca-della-vedova:
To be merged after #160, for now only classes in `graphics` have been refactored, rest of the library will follow.


mjcarroll:
Thanks @luca-della-vedova for getting this cleaned up!

graphics/src/STLLoader.cc
```diff
@@ -28,9 +28,16 @@
 using namespace ignition;
 using namespace common;
 
+
+//////////////////////////////////////////////////
+class ignition::common::STLLoader::Implementation
+{
+};
+
```
mjcarroll, line 36: Since this one is empty was this some sort of scripting/automation thing?

luca-della-vedova: Not really I just went and added implementation pointers to all the classes, the main idea was that if in the future we had to add any data member it wouldn't cause API breakage but I'm not sure how likely the case is for the `STLLoader` class

mjcarroll: Ah this makes more sense, good call. May not be universally useful, but does potentially save some headaches later.

mjcarroll (approved):

PR #165, opened 2021-02-04
Added unit capsule and ellipsoid

ahcorde:
Created unit ellipsoid and capsule

Related with:

 - https://github.com/ignitionrobotics/ign-rendering/pull/203
 - https://github.com/ignitionrobotics/ign-rendering/pull/200


mjcarroll (approved):

graphics/src/MeshManager.cc
```diff
@@ -91,6 +91,10 @@ MeshManager::MeshManager()
   this->CreateCylinder("unit_cylinder", 0.5, 1.0, 1, 32);
   this->CreateCone("unit_cone", 0.5, 1.0, 5, 32);
   this->CreateCamera("unit_camera", 0.5);
+  this->CreateCapsule("unit_capsule", 0.5f, 0.3f, 32, 32);
+  this->CreateEllipsoid("unit_ellipsoid",
+    ignition::math::Vector3d(0.3f, 0.5f, 1.0f),
+    32, 32);
```
chapulina, line 97: All our current "unit" shapes have a `(1, 1, 1)` bounding box. That wouldn't be the case for the new unit shapes introduced here. I understand that for a capsule and an ellipsoid to have these dimensions, they'd be spheres, so that may be confusing. On the other hand, a user scaling these shapes may be confused to see that scaling "unit" shapes results in unexpected sizes.

My vote would be to give them a unit bounding box, even if it means they look like spheres. But I'd like to hear what @scpeters thinks.

mjcarroll: I think if that's the case, we should also add a test to make sure that our primitives fit in a bounding box. 

It makes sense to do so, I just wasn't aware of it :)

chapulina: This came to mind because @scpeters talked about it during one of the meetings. I think that when we insert them from the GUI, we don't need to insert the unit ones. But the unit shapes will be used to create scaled versions of any other size, so I think it makes sense they have a clear unit value.

I just realized though that the capsule doesn't work like the other unit shapes. It's possible to adjust a unit cylinder to any given radius and length by scaling its 3 dimensions, but that can't be done with a capsule. So I think we may need to treat the capsule differently :confused:

ahcorde: what about be more flexible in the definition? such as: something inside the `(1, 1, 1)` bounding box.

chapulina: The problem is how to use these unit shapes. For example, if a user wants to create an ellipse with radii `(1, 2, 3)`, they will need to scale the unit mesh by `(1/0.3, 2/0.5, 3/1.0)`, right? And for the capsule, it won't be possible to create a capsule with radius 2.0 and length 0.5 from the unit capsule.

scpeters: can't an ellipsoid just be created by rescaling a unit sphere?

aside from the size of the bounding box, capsules don't scale the same way as these other shapes. if you want to stretch the length of the cylinder, that should not affect the shape of the hemispheres at all, it should just push them farther apart. but if you want to change the radius, you have to change two dimensions of the cylinder and all three dimensions of the hemispheres. for the capsule, would it be possible to create unit hemisphere with bounding box size [1, 1, 0.5] and construct the mesh by concatenating hemispheres on each end of a cylinder?

chapulina: > can't an ellipsoid just be created by rescaling a unit sphere?

+1

> construct the mesh by concatenating hemispheres on each end of a cylinder?

I think the `CreateCapsule` function above could be used to achieve the same thing in a more efficient way. The downside is that it departs from the approach used by all other shapes (i.e. scale a unit mesh). But I don't think there's a simpler way to go about this...

ahcorde: The scale capability is unabled in [ignition gazebo](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/gui/plugins/transform_control/TransformControl.qml#L31-L35). There is no chance to scale the ellipsoid to give it a different shape than a sphere (at least using the transform control).

having said that is there any other thoughts about the unit bounding box ? or how to procced here ?

chapulina: > There is no chance to scale the ellipsoid to give it a different shape than a sphere (at least using the transform control).

Oh sorry, I wasn't referring to the transform control tool. I meant a developer writing code that uses the `unit_ellipsoid` that is being introduced here, like it's done in https://github.com/ignitionrobotics/ign-rendering/pull/203.

For example, with this `ign-rendering` code, I'd expect to create an ellipsoid with radii `(1, 2, 3)`:

```
  VisualPtr ellipsoid = _scene->CreateVisual();
  ellipsoid->AddGeometry(_scene->CreateEllipsoid());
  ellipsoid->SetLocalScale(1, 2, 3);
```

But if the `CreateEllipsoid` function creates a "unit ellipsoid" with size `(0.3, 0.5, 1.0)`, then the code above will generate an ellipsoid with radii `(0.1, 1.0, 3.0)` instead.

> how to procced here ?

My suggestion is to decline this PR, because:

* `unit_sphere` can be scaled to create ellipsoids
* `unit_capsule` can't be scaled to create capsules with different radius / length

ahcorde:
Declined PR

PR #168, opened 2021-02-06
Added C language bindings for SystemPaths::FindFile and Console::SetVerbosity

peci1:
# New feature

Helps closing https://github.com/ignitionrobotics/ign-launch/issues/92 .

## Summary
The functionality in `SystemPaths` is already quite mature, so I think it makes sense to expose it to other languages via C bindings. This will help code re-use in other parts of Ignition libraries, e.g. in Ruby scripts.

## Test it
Unit tests were added. Otherwise, you can test it e.g. from Ruby by importing the common library and calling some of the added functions.

## Checklist
- [x] Added tests
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+archived%3Afalse+) to support the maintainers

### DIFF
--- a/src/ign.cc
+++ b/src/ign.cc
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2021 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+#include <cstring>
+
+#include <ignition/common/Console.hh>
+#include <ignition/common/SystemPaths.hh>
+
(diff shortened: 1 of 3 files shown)


chapulina:
We have been treating these C functions as internal to each library, and we often break their APIs. It hasn't been our intention so far to use them across libraries. I can see how this is useful, but I think we need to be careful before we start using them this way.

I don't have a concrete alternative proposal at the moment. I think that once all libraries have transitioned to loading binaries instead of libraries (https://github.com/ignitionrobotics/ign-tools/issues/7), we should be able to use the C++ API for this kind of thing.

peci1:
Ahh, that's a pity. So I'm inclined to close this PR and re-implement what SystemPaths do in ign-launch in Ruby (probably not so sophisticated). Because I want the ign-launch PR to appear in Dome, and I assume the ign binaries will start being used in Edifice. 

Do you have any idea how easy or difficult would it be to actually use the C++ API via Ruby FFI? Wouldn't it be too cumbersome? C API is usually much easier to use via FFI.

If you ever plan to support also other language bindings like Python, it might actually be beneficial to provide a stable C API. That way, contributors to the library would only need to make sure their change either does not need changing the C API (which would mean the bindings need no change at all), or that it needs a change in the C API, and that would mean all bindings also need to be updated. But I think the first case would be much more common, making contributing to the libraries much easier. I think there's not that many developers that are fluent in C++, C, and e.g. Ruby, Python and Java (which are some possible candidates for language bindings).

PR #188, opened 2021-03-25
Fix crash when using BVH animations

chapulina:
# 🦟 Bug fix

## Summary

Actors using BVH animations have been crashing `ign-gazebo`, probably since #136, see https://github.com/ignitionrobotics/ign-common/pull/136#issuecomment-805449865.

The problem is that we were creating a temporary `Skeleton`, getting a pointer to its animation, and then deleting the original skeleton. Now that the skeleton is properly destroyed, we also destroy that pointer. The solution is to make a copy so we can keep ownership.

This needs to be backported to `ign-common3`, I'm targeting `main` so we can get it into Edifice's first release. I believe this is a critical bug fix.

@adlarkin , I believe you've been looking into this. Maybe you can see if this works?

## Checklist
- [ ] Added tests - I suggest we just manually test this right now, and after Edifice's release we come back to add proper tests.

### DIFF
--- a/graphics/src/Skeleton.cc
+++ b/graphics/src/Skeleton.cc
@@ -30,7 +30,7 @@ class ignition::common::Skeleton::Implementation
     RawNodeWeights;
 
   /// \brief the root node
-  public: SkeletonNode *root;
+  public: SkeletonNode *root{nullptr};
 
   /// \brief The dictionary of nodes, indexed by name
   public: SkeletonNodeMap nodes;
@@ -68,7 +68,6 @@ class ignition::common::Skeleton::Implementation
 Skeleton::Skeleton()
 : dataPtr(ignition::utils::MakeUniqueImpl<Implementation>())
 {
-  this->dataPtr->root = nullptr;
 }
 
 //////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


adlarkin (commented):
This seems to fix the actor crashing issue. However, I still see this error that was mentioned in  https://github.com/ignitionrobotics/ign-gazebo/issues/647 when I run one of the example actor worlds (for example, `actor.sdf` or `actor_crowd.sdf`):
```
[GUI] [Err] [RenderUtil.cc:919] invalid animation update data
```

It looks like this is coming from the `RenderUtil` class in `ign-gazebo`, so I'm not sure if this issue can be fixed in this PR or not. Any thoughts? Should we just merge this first to fix the critical bug, and then try to resolve this error message later?

Also, this PR may be able to close https://github.com/ignitionrobotics/ign-gazebo/issues/647, unless we want to leave that issue open while we resolve the `RenderUtil` error message.

chapulina:
> this PR may be able to close ignitionrobotics/ign-gazebo#647

Oh thanks! I totally forgot there was an issue for this!

> It looks like this is coming from the RenderUtil class in ign-gazebo

I also see that. Yeah I suspect that's something to be fixed on the `ign-gazebo` side. The animation seems to work for me even when the error is printed, so I think that's not critical to fix.

> leave that issue open while we resolve the RenderUtil error message.

That works for me. I think we also shouldn't close it until this is backported and we add some tests.

PR #197, opened 2021-04-09
Improved Windows support

ahcorde:

# :factory_worker:  Improvement

## Summary

Improved Windows support. This PR is required here https://github.com/ignitionrobotics/ign-fuel-tools/pull/178

## Checklist
- [x] Added tests


ahcorde:
There is an issue removing a TMP_*.mp4 file in one of tests. This is not making the test to fail but I think we can open an issue. Probably this is because the file is not closed.

include/ignition/common/Util.hh
```diff
@@ -186,6 +186,8 @@ namespace ignition
         void const *_buffer, std::size_t _byteCount);
 
     #ifdef _MSC_VER
+      // Disable warning C4307
```
mjcarroll, line 189: Suggestion in https://github.com/ignitionrobotics/ign-common/pull/198

mjcarroll (changes requested):
Can we add a test for the new functionality?

include/ignition/common/Util.hh
mjcarroll: #198 came back clean, so I merged it here.

ahcorde:
> Can we add a test for the new functionality?

@mjcarroll, should we make this function a method (public or private)?

mjcarroll (approved):

mjcarroll:
> @mjcarroll, should we make this function a method (public or private)?

We talked offline, this doesn't need to be a public method as it is just fixing tests.  Codecov drop is because of Linux vs Windows execution paths.

PR #199, opened 2021-04-12
Backport #188: Fix crash when using BVH animations

adlarkin:
Signed-off-by: Ashton Larkin <ashton@openrobotics.org>

# 🦟 Bug fix

Related to https://github.com/ignitionrobotics/ign-gazebo/issues/647

## Summary
This is a backport of #188. Once this is merged, we should no longer see actors crash at launch (the `ign-common3` branch applies to both Citadel and Dome).

To test this, run `ign gazebo -r actor.sdf` and make sure no crash occurs.

## Checklist
- [ ] Added tests

### DIFF
--- a/graphics/include/ignition/common/SkeletonAnimation.hh
+++ b/graphics/include/ignition/common/SkeletonAnimation.hh
@@ -47,6 +47,15 @@ namespace ignition
       /// the animations
       public: ~SkeletonAnimation();
 
+      // NOTE: this is not needed starting in ign-common4 since ign-common4 uses
+      // the IGN_UTILS_IMPL_PTR instead of a raw impl pointer. So, this
+      // method should not be included in forward ports from ign-common3 to
+      // ign-common4
+      /// \brief Assignment operator
+      /// \param[in] _other The new SkeletonAnimation
+      /// \return A reference to this instance
+      public: SkeletonAnimation &operator=(const SkeletonAnimation &_other);
+
       /// \brief Changes the name
       /// \param[in] _name the new name
       public: void SetName(const std::string& _name);
--- a/graphics/src/Skeleton.cc
+++ b/graphics/src/Skeleton.cc
@@ -30,7 +30,7 @@ class ignition::common::SkeletonPrivate
     RawNodeWeights;
 
   /// \brief the root node
-  public: SkeletonNode *root;
+  public: SkeletonNode *root{nullptr};
 
   /// \brief The dictionary of nodes, indexed by name
   public: SkeletonNodeMap nodes;
@@ -68,7 +68,6 @@ class ignition::common::SkeletonPrivate
 Skeleton::Skeleton()
   : data(new SkeletonPrivate)
 {
-  this->data->root = nullptr;
 }
 
 //////////////////////////////////////////////////
@@ -82,10 +81,14 @@ Skeleton::Skeleton(SkeletonNode *_root)
 //////////////////////////////////////////////////
 Skeleton::~Skeleton()
 {
-  for (auto& kv : this->data->nodes)
+  for (auto &kv : this->data->nodes)
     delete kv.second;
-  for (auto& a : this->data->anims)
+  this->data->nodes.clear();
+
+  for (auto &a : this->data->anims)
     delete a;
+  this->data->anims.clear();
+
   delete this->data;
   this->data = NULL;
 }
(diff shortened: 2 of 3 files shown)


graphics/src/Skeleton.cc
```diff
@@ -82,12 +81,13 @@ Skeleton::Skeleton(SkeletonNode *_root)
...
-  for (auto& kv : this->data->nodes)
+  for (auto &kv : this->data->nodes)
     delete kv.second;
-  for (auto& a : this->data->anims)
+  this->data->nodes.clear();
+
+  for (auto &a : this->data->anims)
     delete a;
-  delete this->data;
-  this->data = NULL;
```
iche033, line 90: keep the last two lines for deleting `this->data` and setting to `nullptr`? It's still a raw ptr in common3

adlarkin: Ah, yeah, sorry about that - fixed in https://github.com/ignitionrobotics/ign-common/pull/199/commits/c9a3d7221731a7c96b91f8fa161e536ae44c292e

adlarkin:
I tried testing this by running `ign gazebo -r actor.sdf`. I'm not sure why, but this does not seem to fix the actor crashing on citadel or dome. For citadel, the simulation starts, but then crashes after a few seconds. For dome, it crashes on launch with this error:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
```

I'm not sure why this is happening. I'll need to look into it some more, unless someone else knows what's going on here?

mjcarroll:

chapulina:
I spent some time debugging the crashes on Citadel but the more I dig, the uglier it gets. I have added some safety checks on the following 2 branches:

* https://github.com/ignitionrobotics/ign-common/commit/5186393ea1af28ab5122557b564ac7dfb808d980
* https://github.com/ignitionrobotics/ign-gazebo/commit/30c10c087f54d495187c6ab764ca5f48369b58a7

My favorite one is the `std::map` that ends up with a size larger than its `max_size` :woman_shrugging: 

---

I suspect that the best course of action right now should be to revert https://github.com/ignitionrobotics/ign-common/pull/136/, because as the documentation says, the `Skeleton` doesn't own its `Animation`s, so it shouldn't be deleting them. I don't know what is supposed to be deleting them though, so I'm not sure what's the correct way to avoid a memory leak. Maybe there was never a proper way to delete it.

What's best, a memory leak, or a segfault? :balance_scale: 

https://github.com/ignitionrobotics/ign-common/blob/b7ac4ed8329d50aee7d8c3e92cdf9b307aef7c0b/graphics/include/ignition/common/Skeleton.hh#L135-L138

adlarkin:
@chapulina I believe I may have fixed the issue :crossed_fingers: since `ign-common3` still uses a [raw impl pointer](https://github.com/ignitionrobotics/ign-common/blob/ignition-common3_3.12.0/graphics/include/ignition/common/SkeletonAnimation.hh#L129-L130), we need to do a deep copy here, or else we are just copying the pointer, which eventually goes out of scope. [`ign-common4` uses the new utils impl ptr](https://github.com/ignitionrobotics/ign-common/blob/ignition-common4_4.0.0/graphics/include/ignition/common/SkeletonAnimation.hh#L128-L129), so that's why this wasn't needed in #188. I have added an assignment operator in https://github.com/ignitionrobotics/ign-common/pull/199/commits/a3983b6cb7059244118febb2b3b2f4497f4302a7, and that seems to fix the crash for me.

More information about this can be found in https://github.com/ignitionrobotics/ign-common/pull/136#issuecomment-805449865.

graphics/include/ignition/common/SkeletonAnimation.hh
```diff
@@ -47,6 +47,11 @@ namespace ignition
       /// the animations
       public: ~SkeletonAnimation();
 
+      /// \brief Assignment operator
+      /// \param[in] _other The new SkeletonAnimation
+      /// \return A reference to this instance
+      public: SkeletonAnimation &operator=(const SkeletonAnimation &_other);
```
iche033, line 53: this function should not be merged forward to ign-common4 right? Can you add a note as a reminder?

adlarkin: Added: https://github.com/ignitionrobotics/ign-common/pull/199/commits/af8b84ad9c5be203be79fc655db07346b964213f

Let me know if there's a better way/format of adding a note like this. I don't think it should be a `TODO` since nothing really needs to be "done", and I also assume it shouldn't be a doxygen comment since this isn't relevant for user documentation (it's more of just an internal developer note).

chapulina (approved):
Niiice! It works for me with the assignment operator :smiley:

graphics/include/ignition/common/SkeletonAnimation.hh
iche033: looks good to me.

PR #204, opened 2021-04-20
Remove ColladaExporter path constraint

### LINKED ISSUE #202: Collada Exporter not taking into account all the material/effects
## Description
* Expected behavior: 

When using the Collada Exporter from ign-gazebo, all textures files should be exported into the final result.

* Actual behavior: 
They only get exported if their full path name includes "meshes/". 

## Steps to reproduce
In an environment that includes an Ignition installation:

1. Add the ColladaExporter plugin to the `tunnel.sdf` ign-gazebo example file.
2. Run the Collada Exporter on that file.
ign gazebo -v 3 -s -r --iterations 1 tunnel.sdf
3. Open the .dae exported to the path ./tunnel/meshes/tunnel.sdf

However, this can't be reproduced at the moment because it depends on fixing https://github.com/ignitionrobotics/ign-gazebo/issues/761.

I was able to track this problem to these conditional include statements like this one:
https://github.com/ignitionrobotics/ign-common/blob/d496d79d9d90b9474e7ccd298210115084e472ad/graphics/src/ColladaExporter.cc#L511

The collada exporter is only taking into account effects/textures if the path included "meshes/' in its name. However, `ign-gazebo` is able to open all files in its environment path variable and fuel models are usually loaded from `~/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models/`, in neither of those cases the file includes the `meshes`.

I imagine there was an old context for this check statement, I didn't find any in the history of this file though. If there are no problems I can open a PR to remove those checks or add to the check at least paths including "materials" as well. Let me know your thoughts.

Blast545:
Signed-off-by: Jorge Perez <jjperez@ekumenlabs.com>

# 🦟 Bug fix

Fixes #202 

## Summary
This PR removes the ColladaExporter constraint on the path to include "meshes" in their names to be exported. As per discussion in #202 this check might be related to code from Ignition Gazebo and may not be needed now.

## Checklist
- [x] Added tests

### DIFF
--- a/graphics/src/ColladaExporter.cc
+++ b/graphics/src/ColladaExporter.cc
@@ -508,7 +508,7 @@ int ColladaExporter::Implementation::ExportImages(
       this->mesh->MaterialByIndex(i);
     std::string imageString = material->TextureImage();
 
-    if (imageString.find("meshes/") != std::string::npos)
+    if (imageString.find("/") != std::string::npos)
     {
       char id[100];
       snprintf(id, sizeof(id), "image_%u", i);
@@ -589,7 +589,7 @@ void ColladaExporter::Implementation::ExportEffects(
         this->mesh->MaterialByIndex(i);
     std::string imageString = material->TextureImage();
 
-    if (imageString.find("meshes/") != std::string::npos)
+    if (imageString.find("/") != std::string::npos)
     {
       tinyxml2::XMLElement *newParamXml =
         _libraryEffectsXml->GetDocument()->NewElement("newparam");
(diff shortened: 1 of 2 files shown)


Blast545:
Changed the check to check for `/` as the functions can receive empty paths, and those should be taken into account. 

Also the test is addressing properly the fix included here, this is ready for review.

mjcarroll:
Homebrew warnings are known because of `libav` versions.

mjcarroll (approved):

PR #205, opened 2021-04-23
Add function to convert single channel image data to RGB image

iche033:
# 🎉 New feature

## Summary

Used by https://github.com/ignitionrobotics/ign-gui/pull/212 and https://github.com/ignitionrobotics/ign-launch/pull/112

Added a generic helper function for converting single channel image data, e.g. thermal and depth images, to an RGB image. This is intended to be used for visualization. Optional arguments can be specified to tweak the visualization. 

## Test it

The corresponding PRs in ign gui updates the Image Display to use this new function: https://github.com/ignitionrobotics/ign-gui/pull/212

You can test this together with that PR by running the `sensors_demo.sdf` world to see the visualization in Image Display

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [ ] Added example and/or tutorial
- [x] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #206, opened 2021-04-26
Remove raw new/delete calls from #205

mjcarroll:
I think that we should generally avoid raw `new` and `delete` calls in code going forward.  In this case, the test was also allocating memory without deallocating.

I replaced heap allocations with `std::vector<T>` where it made sense.


iche033 (approved):
looks good to me, thanks!

ahcorde (approved):

PR #208, opened 2021-04-29
 Fix loading collada files with multiple texcoord sets using the same offset

iche033:
# 🦟 Bug fix

## Summary

When loading texture coordinates, there was an assumption in the ColladaLoader that the `offset` values (which specify the index in which the uv coordinates should be read from) need to be unique. This is found that be incorrect assumption after looking at Collada 1.5 spec ([link](https://www.khronos.org/files/collada_spec_1_5.pdf) to pdf. There is an example on page 89). 

This PR updates the loader to support loading duplicate `offset` values.

One way to test this is to use Edifice. You'll need to use this ign-common4 branch: `collada_texcoord_offset_4`, and launch a world in ign-gazebo with this [model](https://github.com/ignitionrobotics/ign-common/files/6401985/WheelChairUnfolded.zip) (zip file). Without this change, the light map is rendered incorrectly on the model and it'll appear all white. With this fix, the model should have right textures but with some subtle difference in shading.

## Checklist
- [x] Added tests

### DIFF
--- a/graphics/src/ColladaLoader.cc
+++ b/graphics/src/ColladaLoader.cc
@@ -2167,7 +2167,7 @@ void ColladaLoaderPrivate::LoadPolylist(tinyxml2::XMLElement *_polylistXml,
               if (!inputs[TEXCOORD].empty())
               {
                 texEqual = true;
-                for (auto pair: texcoordsOffsetToSet)
+                for (auto &pair : texcoordsOffsetToSet)
                 {
                   unsigned int offset = pair.first;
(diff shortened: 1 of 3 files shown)


luca-della-vedova (approved):
Didn't look too much at the implementation since I'm not very familiar with the `ColladaLoader` but it fixed the issue for me! (when cherry-picked to `common4` for edifice)

mjcarroll:
Warnings on homebrew are expected, I think this is good to go.

PR #209, opened 2021-04-30
Fix join paths to remove duplicate separators

nkoenig:
# 🦟 Bug fix

## Summary

The `joinPaths` function would remove trailing slashes from the function arguments. This can cause downstream applications, like fuel tools using libzip, to treat a directory as a file.

## Checklist
- [X] Signed all commits for DCO
- [X] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [X] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [X] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

PR #217, opened 2021-05-10
Revert "Fix join paths to remove duplicate separators"

mjcarroll:
Rather than landing https://github.com/ignitionrobotics/ign-common/pull/215, I believe the correct thing is to revert #209, perform a release, and then get https://github.com/ignitionrobotics/ign-fuel-tools/pull/178 landed, which should resolve the issues that we seeing (https://github.com/ignitionrobotics/ign-fuel-tools/issues/183)

Reverts ignitionrobotics/ign-common#209


ahcorde (changes requested):
@mjcarroll  DCO

PR #228, opened 2021-07-12
Export lights to dae

ddengster:
# 🎉 New feature

Closes [#901](https://github.com/ignitionrobotics/ign-gazebo/issues/901)

[Downstream PR](https://github.com/ignitionrobotics/ign-gazebo/pull/912)

## Summary
- Exports lights for the collada world exporter. Based off the [collada 1.4 spec](https://www.khronos.org/files/collada_spec_1_4.pdf)
- Test added for collada exporter exporting lights

## Test it
run `./build/ignition-common4/bin/UNIT_ColladaExporter_TEST`

## Checklist
- [X] Added tests

### DIFF
--- a/graphics/include/ignition/common/ColladaExporter.hh
+++ b/graphics/include/ignition/common/ColladaExporter.hh
@@ -38,7 +38,7 @@ namespace ignition
     struct ColladaLight
     {
       std::string name;
-      std::string type; //either "point", "directional" or "spot"
+      std::string type;  // either "point", "directional" or "spot"
       math::Vector3d direction;
       math::Vector3d position;
       math::Color diffuse;
@@ -69,7 +69,8 @@ namespace ignition
       public: void Export(const Mesh *_mesh,
           const std::string &_filename, bool _exportTextures,
           const std::vector<math::Matrix4d> &_submeshToMatrix,
-          const std::vector<ColladaLight> &_lights = std::vector<ColladaLight>());
+          const std::vector<ColladaLight> &_lights =
+            std::vector<ColladaLight>());
 
       /// \brief Pointer to private data.
       IGN_UTILS_IMPL_PTR(dataPtr)
--- a/graphics/include/ignition/common/ColladaExporter.hh
+++ b/graphics/include/ignition/common/ColladaExporter.hh
@@ -37,16 +37,35 @@ namespace ignition
     /// Defaults set based on collada 1.4 specifications
     struct ColladaLight
     {
-      std::string name;  // name of the light
-      std::string type;  // either "point", "directional" or "spot"
-      math::Vector3d direction;  // direction (directional/spot lights only)
-      math::Vector3d position;  // light position (non directional lights only)
-      math::Color diffuse;  // light diffuse color
-      double constant_attenuation = 1.0;  // Constant attentuation
-      double linear_attenuation = 0.0;  // Linear attentuation
-      double quadratic_attenuation = 0.0;  // Quadratic attentuation
-      double falloff_angle_deg = 180.0;  // Falloff angle in degrees
-      double falloff_exponent = 0.0;  // Fallof exponent
+      /// \brief Name of the light
+      std::string name;
+
+      /// \brief Type of the light. Either "point", "directional" or "spot"
+      std::string type;
+
+      /// \brief Light direction (directional/spot lights only)
+      math::Vector3d direction;
+
+      /// \brief Light position (non directional lights only)
+      math::Vector3d position;
+
+      /// \brief Light diffuse color
+      math::Color diffuse;
+
+      /// \brief Constant attentuation
+      double constantAttenuation = 1.0;
+
+      /// \brief Linear attentuation
+      double linearAttenuation = 0.0;
+
+      /// \brief Quadratic attentuation
+      double quadraticAttenuation = 0.0;
+
+      /// \brief Falloff angle in degrees
+      double falloffAngleDeg = 180.0;
+
+      /// \brief Fallof exponent
+      double falloffExponent = 0.0;
     };
 
     /// \brief Class used to export Collada mesh files
@@ -66,10 +85,23 @@ namespace ignition
       public: virtual void Export(const Mesh *_mesh,
           const std::string &_filename, bool _exportTextures = false);
 
+      /// \brief Export a mesh to a file
+      /// \param[in] _mesh Pointer to the mesh to be exported
+      /// \param[in] _filename Exported file's path and name
+      /// \param[in] _exportTextures True to export texture images to
+      /// '../materials/textures' folder
+      /// \param[in] _submeshToMatrix Matrices of submeshes
       public: void Export(const Mesh *_mesh,
           const std::string &_filename, bool _exportTextures,
           const std::vector<math::Matrix4d> &_submeshToMatrix);
 
+      /// \brief Export a mesh to a file
+      /// \param[in] _mesh Pointer to the mesh to be exported
+      /// \param[in] _filename Exported file's path and name
+      /// \param[in] _exportTextures True to export texture images to
+      /// '../materials/textures' folder
+      /// \param[in] _submeshToMatrix Matrices of submeshes
+      /// \param[in] _lights List of lights to export
       public: void Export(const Mesh *_mesh,
           const std::string &_filename, bool _exportTextures,
           const std::vector<math::Matrix4d> &_submeshToMatrix,
--- a/graphics/include/ignition/common/ColladaExporter.hh
+++ b/graphics/include/ignition/common/ColladaExporter.hh
@@ -24,8 +24,8 @@
 #include <ignition/common/MeshExporter.hh>
 #include <ignition/common/graphics/Export.hh>
 
-#include <ignition/math/Matrix4.hh>
 #include <ignition/math/Color.hh>
+#include <ignition/math/Matrix4.hh>
 
 #include <ignition/utils/ImplPtr.hh>
 
@@ -33,20 +33,20 @@ namespace ignition
 {
   namespace common
   {
-    // Light data structure specifically for collada export
-    // Defaults set based on collada 1.4 specifications
+    /// \brief This struct contains light data specifically for collada export
+    /// Defaults set based on collada 1.4 specifications
     struct ColladaLight
     {
-      std::string name;
+      std::string name;  // name of the light
       std::string type;  // either "point", "directional" or "spot"
-      math::Vector3d direction;
-      math::Vector3d position;
-      math::Color diffuse;
-      double constant_attenuation = 1.0;
-      double linear_attenuation   = 0.0;
-      double quadratic_attenuation = 0.0;
-      double falloff_angle_deg = 180.0;
-      double falloff_exponent = 0.0;
+      math::Vector3d direction;  // direction (directional/spot lights only)
+      math::Vector3d position;  // light position (non directional lights only)
+      math::Color diffuse;  // light diffuse color
+      double constant_attenuation = 1.0;  // Constant attentuation
+      double linear_attenuation = 0.0;  // Linear attentuation
+      double quadratic_attenuation = 0.0;  // Quadratic attentuation
+      double falloff_angle_deg = 180.0;  // Falloff angle in degrees
+      double falloff_exponent = 0.0;  // Fallof exponent
     };
 
     /// \brief Class used to export Collada mesh files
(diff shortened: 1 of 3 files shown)


graphics/src/ColladaExporter.cc
```diff
@@ -239,10 +242,75 @@ void ColladaExporter::Export(const Mesh *_mesh, const std::string &_filename,
...
+    tinyxml2::XMLElement *lightTypeXml =
+      xmlDoc.NewElement(light.type.c_str());
+    techniqueCommonXml->LinkEndChild(lightTypeXml);
+
+    // color
+    tinyxml2::XMLElement *colorXml =
+      xmlDoc.NewElement("color");
+    char color_str[100] = { 0 };
+    snprintf(color_str, sizeof(color_str), "%g %g %g",
+      light.diffuse.R(), light.diffuse.G(), light.diffuse.B());
```
ahcorde, line 268: can we use `std::string` instead of `char*`. Same below

graphics/src/ColladaExporter_TEST.cc
```diff
@@ -315,6 +315,151 @@ TEST_F(ColladaExporter, ExportMeshWithSubmeshes)
...
+  int node_with_light_count = 0;
+  auto node_ele = scene->FirstChildElement("node");
+  for (; node_ele != nullptr; node_ele = node_ele->NextSiblingElement())
+  {
+    const char* node_name_cstr = node_ele->Attribute("name");
+    ASSERT_TRUE(node_name_cstr);
+    std::string node_name = node_name_cstr;
+    if (node_name == "sun" || node_name == "lamp" || node_name == "torch")
+    {
+      auto inst_light = node_ele->FirstChildElement("instance_light");
```
ahcorde, line 450: do you need these variables ?

graphics/src/ColladaExporter_TEST.cc
```diff
@@ -315,6 +315,151 @@ TEST_F(ColladaExporter, ExportMeshWithSubmeshes)
...
+      auto qatt = point->FirstChildElement("quadratic_attenuation");
+      EXPECT_TRUE(qatt);
+    }
+    else if (light_name == "torch")
+    {
+      auto technique = light_ele->FirstChildElement("technique_common");
+      EXPECT_TRUE(technique);
+      auto spot = technique->FirstChildElement("spot");
+      EXPECT_TRUE(spot);
+      auto color = spot->FirstChildElement("color");
```
ahcorde, line 415: Do you need these variables?

graphics/src/ColladaExporter_TEST.cc
```diff
@@ -315,6 +315,151 @@ TEST_F(ColladaExporter, ExportMeshWithSubmeshes)
...
+      auto color = directional->FirstChildElement("color");
+      EXPECT_TRUE(color);
+    }
+    else if (light_name == "lamp")
+    {
+      auto technique = light_ele->FirstChildElement("technique_common");
+      EXPECT_TRUE(technique);
+      auto point = technique->FirstChildElement("point");
+      EXPECT_TRUE(point);
+      auto color = point->FirstChildElement("color");
```
ahcorde, line 400: Do you need these variables?

ahcorde (changes requested):

graphics/src/ColladaExporter_TEST.cc
ddengster: it's good to just check for the tags no?

graphics/include/ignition/common/ColladaExporter.hh
```diff
@@ -25,13 +25,30 @@
...
   namespace common
   {
+    // Light data structure specifically for collada export
+    // Defaults set based on collada 1.4 specifications
+    struct ColladaLight
+    {
+      std::string name;
+      std::string type;  // either "point", "directional" or "spot"
+      math::Vector3d direction;
+      math::Vector3d position;
```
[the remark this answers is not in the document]
ddengster: defaults are based off the collada 1.4 spec. They don't specify the default values so I leave them alone.

graphics/src/ColladaExporter.cc
ddengster: there's a couple of reasons why i use non-c++ methods here:
1) The syntax is a lot more readable; the c++ style is a lot more cumbersome
2) There are other areas in this file where we use snprintf; I'd rather we stick to the snprintf styles to maintain consistency and get the benefits of (1)

graphics/include/ignition/common/ColladaExporter.hh
```diff
@@ -25,13 +25,30 @@
...
+      math::Color diffuse;
+      double constant_attenuation = 1.0;
+      double linear_attenuation   = 0.0;
```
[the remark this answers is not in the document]
ddengster: 0350c58

graphics/include/ignition/common/ColladaExporter.hh
```diff
@@ -25,13 +25,30 @@
...
 
 #include <ignition/utils/ImplPtr.hh>
 
 namespace ignition
 {
   namespace common
   {
+    // Light data structure specifically for collada export
+    // Defaults set based on collada 1.4 specifications
+    struct ColladaLight
```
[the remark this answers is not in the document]
ddengster: documented 0350c58

graphics/include/ignition/common/ColladaExporter.hh
```diff
@@ -25,13 +25,30 @@
 #include <ignition/common/graphics/Export.hh>
 
 #include <ignition/math/Matrix4.hh>
+#include <ignition/math/Color.hh>
```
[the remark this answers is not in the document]
ddengster: alphabetize 0350c58

graphics/src/ColladaExporter_TEST.cc
ahcorde: I'm fine with the checks, I was just suggesting something like:

```
      EXPECT_TRUE(node_ele->FirstChildElement("translate"));
      EXPECT_TRUE(node_ele->FirstChildElement("rotate"));
```

ahcorde (approved):

graphics/include/ignition/common/ColladaExporter.hh
```diff
@@ -51,7 +68,9 @@ namespace ignition
 
       public: void Export(const Mesh *_mesh,
           const std::string &_filename, bool _exportTextures,
-          const std::vector<math::Matrix4d> &_submeshToMatrix);
+          const std::vector<math::Matrix4d> &_submeshToMatrix,
```
mjcarroll, line 71: Adding a default-valued parameter to an existing function in a release will break ABI, asyou can see in the checker. 

https://build.osrfoundation.org/job/ignition_common-abichecker-any_to_any-ubuntu_auto-amd64/659/API_5fABI_20report/


In order to work around this and land it in ign-common4, I suggest introducing a second `Export` function with the new number of args and then calling that one from the old function.

```
void Export(const Mesh *_mesh,
          const std::string &_filename, bool _exportTextures,
          const std::vector<math::Matrix4d> &_submeshToMatrix,
          const std::vector<ColladaLight> &_lights)
{
    // Current impl goes here
}

public: void Export(const Mesh *_mesh,
          const std::string &_filename, bool _exportTextures,
          const std::vector<math::Matrix4d> &_submeshToMatrix)
{
    Export(_mesh, _filename, _exportTextures, _submeshToMatrix, std::vector<ColladaLight>());
}
```

ddengster: abi compatibility f6e2f92

graphics/include/ignition/common/ColladaExporter.hh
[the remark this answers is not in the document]
chapulina: Ignition Math will initialize these to zero anyway though

graphics/include/ignition/common/ColladaExporter.hh
```diff
@@ -32,6 +33,22 @@ namespace ignition
 {
   namespace common
   {
+    /// \brief This struct contains light data specifically for collada export
+    /// Defaults set based on collada 1.4 specifications
+    struct ColladaLight
+    {
+      std::string name;  // name of the light
```
chapulina, line 40: Please document with `\brief` above the members, inline comments are discouraged

graphics/include/ignition/common/ColladaExporter.hh
```diff
@@ -53,6 +70,11 @@ namespace ignition
           const std::string &_filename, bool _exportTextures,
           const std::vector<math::Matrix4d> &_submeshToMatrix);
 
+      public: void Export(const Mesh *_mesh,
```
chapulina, line 73: It looks like the function above is also missing docs; it would be nice to document both.

graphics/include/ignition/common/ColladaExporter.hh
```diff
@@ -32,6 +33,22 @@ namespace ignition
...
+      std::string name;  // name of the light
+      std::string type;  // either "point", "directional" or "spot"
+      math::Vector3d direction;  // direction (directional/spot lights only)
+      math::Vector3d position;  // light position (non directional lights only)
+      math::Color diffuse;  // light diffuse color
+      double constant_attenuation = 1.0;  // Constant attentuation
+      double linear_attenuation = 0.0;  // Linear attentuation
+      double quadratic_attenuation = 0.0;  // Quadratic attentuation
+      double falloff_angle_deg = 180.0;  // Falloff angle in degrees
+      double falloff_exponent = 0.0;  // Fallof exponent
```
chapulina, line 49: Style nit: use `camelCase`

ddengster: camelCased 1b7602b

graphics/include/ignition/common/ColladaExporter.hh
ddengster: documented both functions 1b7602b

graphics/include/ignition/common/ColladaExporter.hh
ddengster: documented ColladaLight members 1b7602b

graphics/src/ColladaExporter_TEST.cc
ddengster: removed 1b7602b

graphics/src/ColladaExporter_TEST.cc
ddengster: removed 1b7602b

graphics/src/ColladaExporter_TEST.cc
ddengster: remove 1b7602b

mjcarroll (approved):

graphics/include/ignition/common/ColladaExporter.hh
```diff
@@ -32,6 +33,41 @@ namespace ignition
...
+      std::string name;
+
+      /// \brief Type of the light. Either "point", "directional" or "spot"
+      std::string type;
+
+      /// \brief Light direction (directional/spot lights only)
+      math::Vector3d direction;
+
+      /// \brief Light position (non directional lights only)
+      math::Vector3d position;
```
chapulina, line 50: As I commented on https://github.com/ignitionrobotics/ign-gazebo/pull/912#discussion_r681903981, we should use the light's full `math::Pose3d` and apply a direction on top of that.

Now that this has been released, we can't extend this struct though. That's why we prefer full classes with getters and setters using the PIMPL pattern; they involve a lot more boilerplate, but at least they can be extended without breaking ABI.

I think the solution will have to be calculating the `direction` field based on the light's pose orientation together with the light's direction.

> /// \brief Light position (non directional lights only)

For directional lights, the position doesn't affect how they light up the scene, but it's often useful to set so we choose where the light's visual representation shows up on the GUI.

ddengster: Sorry for the delayed reply, was pushing some deadlines.

> Now that this has been released, we can't extend this struct though.

If I may ask, why so? Is there some technical reason?

This struct is specifically used for collada exporting - meaning if anyone decides to use it for something else they are taking on certain risks. I'm of the opinion that we should be able iterate quickly to what we want at an early stage when few are using it.

chapulina: > If I may ask, why so? Is there some technical reason?

ABI compatibility :confused: 

---

I think you may have solved the issue in https://github.com/ignitionrobotics/ign-gazebo/pull/912#discussion_r690976656

PR #231, opened 2021-07-21
Fix a typo in VideoEncoder_TEST.

mahiuchun:
The same typo is also in `ign-common4` and `main` branches.

### DIFF
--- a/av/src/VideoEncoder_TEST.cc
+++ b/av/src/VideoEncoder_TEST.cc
@@ -32,9 +32,33 @@ class VideoEncoderTest : public common::testing::AutoLogFixture
 };
 
 /////////////////////////////////////////////////
-TEST_F(VideoEncoderTest, StartStop)
+TEST_F(VideoEncoderTest, StartStopDefault)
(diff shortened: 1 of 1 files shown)


mahiuchun:
Looks like `mpg` rather than `mp4` passes in the CI, is this the right behavior though?

scpeters:
> Looks like `mpg` rather than `mp4` passes in the CI, is this the right behavior though?

I see why you made this change, but it causes the test to fail. I don't understand the code well enough to say why

mjcarroll:
> I see why you made this change, but it causes the test to fail. I don't understand the code well enough to say why

I agree that the change is correct, but the failure is very strange.

mjcarroll:
The issue here was that the test was calling `Start` twice without a `Stop` or `Reset` in between the two `Start` calls.  I have added some more comprehensive tests to test default, `mpg`, and `mp4` encoding, as well as testing the multiple `Start` case.

chapulina (approved):

PR #237, opened 2021-08-06
Suggestion to #227: More data race fixes

azeey:
# 🦟 Bug fix

## Summary
I ran the `PERFORMANCE_logging` test under TSAN and found more data races. One deals with `std::localtime`, which apparently is not thread-safe as it returns a pointer to an internal static object. The others deal with the fact that the `Buffer` class has a `std::string` member inherited from `std::stringbuf` and it (the string) also needs synchronization.

I put the mutex as a member in `Buffer` since it's needed in two functions. This means that each console stream (i.e. ignmsg, ignwarn, etc) gets its own mutex. It turns out we don't really need to synchronize on the `stdout` or `stderr` 
According to https://timsong-cpp.github.io/cppwp/n3337/iostream.objects#overview-4, accessing `stdout` or `stderr` from multiple threads should not cause a data race, so we don't need a global mutex for them. It does say we should synchronize if we want to avoid interleaved messages, but since we write the `prefixString` and the main message separately, we wouldn't be able to avoid interleaved messages even with global synchronization.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/common/Console.hh
+++ b/include/ignition/common/Console.hh
@@ -206,7 +206,7 @@ namespace ignition
                    /// \brief Level of verbosity
                    public: int verbosity;
 
-                   /// \brief Mutex to syncronize writes to the string buffer
+                   /// \brief Mutex to synchronize writes to the string buffer
                    /// and the output stream.
                    public: std::mutex syncMutex;
                  };
--- a/include/ignition/common/Console.hh
+++ b/include/ignition/common/Console.hh
@@ -183,6 +183,13 @@ namespace ignition
                    /// \brief Destructor.
                    public: virtual ~Buffer();
 
+                   /// \brief Writes _count characters to the string buffer
+                   /// \param[in] _char Input rharacter array.
+                   /// \param[in] _count Number of characters in array.
+                   /// \return The number of characters successfully written.
+                   public: virtual std::streamsize xsputn(
+                        const char *_char, std::streamsize _count) override;
+
                    /// \brief Sync the stream (output the string buffer
                    /// contents).
                    /// \return Return 0 on success.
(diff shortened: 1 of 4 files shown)


include/ignition/common/Console.hh
```diff
@@ -267,6 +278,7 @@ namespace ignition
       /// \brief A custom prefix. See SetPrefix().
       private: static std::string customPrefix;
       IGN_COMMON_WARN_RESUME__DLL_INTERFACE_MISSING
+
```
ahcorde, line 281:
```suggestion
```

ahcorde (changes requested):

include/ignition/common/Console.hh
```diff
@@ -183,6 +183,13 @@ namespace ignition
                    /// \brief Destructor.
                    public: virtual ~Buffer();
 
+                   /// \brief Writes _count characters to the string buffer
+                   /// \param[in] _char Input rharacter array.
+                   /// \param[in] _count Number of characters in array.
+                   /// \return The number of characters successfully written.
+                   public: virtual std::streamsize xsputn(
```
mjcarroll, line 190: I believe that `sync` below also needs to be marked override to make macos happy.

include/ignition/common/Console.hh
```diff
@@ -198,6 +205,10 @@ namespace ignition
 
                    /// \brief Level of verbosity
                    public: int verbosity;
+
+                   /// \brief Mutex to syncronize writes to the string buffer
```
ahcorde, line 209:
```suggestion
                   /// \brief Mutex to synchronize writes to the string buffer
```

ahcorde (changes requested):

mjcarroll (approved):

ahcorde (approved):

Issue #27, opened 2017-12-04
Race condition when running tests in parallel


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-common/issues/27)) by Steve Peters (Bitbucket: [Steven Peters](https://bitbucket.org/%7B2ccfed09-18b8-4921-8d58-15ef01092802%7D/), GitHub: [scpeters](https://github.com/scpeters)).**

----------------------------------------
The debbuilder job runs the tests in parallel after building and often fails during the `UNIT_Console_TEST`:

* https://build.osrfoundation.org/view/debbuild/job/ign-common-debbuilder/53/
* https://build.osrfoundation.org/view/debbuild/job/ign-common-debbuilder/52/
* https://build.osrfoundation.org/view/debbuild/job/ign-common-debbuilder/55/

There is probably a race condition somewhere in our testing code. I think a short-term fix is to disable parallel testing in our jenkins metadata.

cc: @\_jrivero_

azeey:
I think this has been fixed by #378. It would be interesting to run tests in parallel again.

Issue #26, opened 2017-11-28
Tinyxml2 asserts disabled in internal copy of XMLDocument::Clear


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-common/issues/26)) by Steve Peters (Bitbucket: [Steven Peters](https://bitbucket.org/%7B2ccfed09-18b8-4921-8d58-15ef01092802%7D/), GitHub: [scpeters](https://github.com/scpeters)).**

----------------------------------------
In [pull request #69](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-common/pull-requests/69), some asserts were disabled in our internal copy of tinyxml2 since they were failing on windows. From bee642890803046b9f75ee1c2686222fc1dca72a:

~~~
      // Dev notes by MXG:
      // These assertions fail when the DEBUG macro is defined. They fail invisibly on 
      // non-Windows systems because the DEBUG macro is not checking for NDEBUG, which 
      // is the standard non-Windows preprocessor token for specifying (not) debug mode.
      // Testing against tinyxml2 version 4 also triggers these asserts.
      // Testing against version 5 procuces a stack overflow, which is a much worse situation.
      // It is not clear to me whether this is an inherent issue with tinyxml2 or whether 
      // we are simply misusing it somewhere.
~~~

azeey:
We no longer have a copy of tinyxml, so I'll go ahead and close this.

Issue #25, opened 2017-10-17
Cmake configuration error on homebrew


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-common/issues/25)) by Steve Peters (Bitbucket: [Steven Peters](https://bitbucket.org/%7B2ccfed09-18b8-4921-8d58-15ef01092802%7D/), GitHub: [scpeters](https://github.com/scpeters)).**

----------------------------------------
Starting with [build 74](https://build.osrfoundation.org/job/ignition_common-ci-default-homebrew-amd64/74), the [ignition_common-ci-default-homebrew-amd64](https://build.osrfoundation.org/job/ignition_common-ci-default-homebrew-amd64) build has been failing. It fails during `make install` due to a permissions issue:

~~~
Install the project...
-- Install configuration: "RelWithDebInfo"
Scanning dependencies of target man
Built target man
CMake Error at cmake_install.cmake:43 (file):
  file cannot create directory: /ignition/common0/ignition/common.  Maybe
  need administrative privileges.


make: *** [install] Error 1
~~~

but it is using the wrong install prefix for one of the header files: `/ignition/common0/ignition/common`. From the generated `cmake_install.cmake` in the build folder:

~~~
if("${CMAKE_INSTALL_COMPONENT}" STREQUAL "headers" OR NOT CMAKE_INSTALL_COMPONENT)
  list(APPEND CMAKE_ABSOLUTE_DESTINATION_FILES
   "/ignition/common0/ignition/common/config.hh")
  if(CMAKE_WARN_ON_ABSOLUTE_INSTALL_DESTINATION)
    message(WARNING "ABSOLUTE path INSTALL DESTINATION : ${CMAKE_ABSOLUTE_DESTINATION_FILES}")
  endif()
  if(CMAKE_ERROR_ON_ABSOLUTE_INSTALL_DESTINATION)
    message(FATAL_ERROR "ABSOLUTE path INSTALL DESTINATION forbidden (by caller): ${CMAKE_ABSOLUTE_DESTINATION_FILES}")
  endif()
file(INSTALL DESTINATION "/ignition/common0/ignition/common" TYPE FILE FILES "/Users/jenkins/workspace/ignition_common-ci-default-homebrew-amd64/build/include/ignition/common/config.hh")
endif()
~~~

[...]

Issue #24, opened 2017-10-11
VideoEncoder test fails on osx due to lack of ogv encoder


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-common/issues/24)) by Steve Peters (Bitbucket: [Steven Peters](https://bitbucket.org/%7B2ccfed09-18b8-4921-8d58-15ef01092802%7D/), GitHub: [scpeters](https://github.com/scpeters)).**

----------------------------------------
The VideoEncoder test has been failing on homebrew for a loooooong time:

* https://build.osrfoundation.org/job/ignition_common-ci-default-homebrew-amd64/72/testReport/(root)/VideoEncoderTest/StartStop/

There are more clues in the console output than in the failed expectations:

~~~
65: [Err] [VideoEncoder.cc:276] Codec for[vp8] not found. Video encoding is not started.
...
65: /Users/jenkins/workspace/ignition_common-ci-default-homebrew-amd64/ign-common/src/VideoEncoder_TEST.cc:47: Failure
65: Value of: "ogv"
65: Expected: video.Format().c_str()
65: Which is: "mp4"
~~~

The test [explicitly asks to encode a video with type ogv](https://github.com/ignitionrobotics/ign-common/blob/bf92cff5f2917b0dc02f0febd7433a42665e4b8b/src/VideoEncoder_TEST.cc#L45), but I can confirm that this codec isn't supported by the default installation of ffmpeg on homebrew:

~~~
$ ffmpeg -i video.mp4 video.ogv
...
Automatic encoder selection failed for output stream #0:0.
Default encoder for format ogg (codec theora) is probably disabled.
Please choose an encoder manually.
Error selecting an encoder for stream 0:0
~~~

PR #238, opened 2021-08-10
Update to use std::filesystem rather than Boost

### LINKED ISSUE #232: An API for getting temp path in Filesystem.hh
## Desired behavior

`Filesystem.hh` should provide a function for getting a temp path suitable for the current operating system.

## Alternatives considered

N/A

## Implementation suggestion

Use the behavior as described in here:
https://www.boost.org/doc/libs/1_75_0/libs/filesystem/doc/reference.html#temp_directory_path
To cover both Linux/macOS and Windows.

## Additional context

`common::cwd()` is often used to "get a path to save a file" currently.
The problem with this approach is that CWD is not necessarily writable.
`av/src/VideoEncoder.cc` contains an example of this pattern.

mjcarroll:
# 🎉 Remove FilesystemBoost.cc in favor of std::filesystem

## Summary

Remove the vendored `FilesystemBoost.cc` file in favor of using `std::filesystem`.

## Test it

Functionality should remain the same across the stack, and all local unit/integration tests should pass on all platforms.

## Checklist
- [x] Added tests


include/ignition/common/TempDirectory.hh
```diff
@@ -0,0 +1,76 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#ifndef IGNITION_COMMON_TEMPDIRECTORY_HH_
+#define IGNITION_COMMON_TEMPDIRECTORY_HH_
+
+#include <string>
```
ahcorde, line 21: alphabetize

include/ignition/common/TempDirectory.hh
```diff
@@ -0,0 +1,76 @@
...
+    std::string IGNITION_COMMON_VISIBLE createTempDirectory(
+        const std::string &_baseName,
+        const std::string &_parentPath,
+        const FilesystemWarningOp _warningOp = FSWO_LOG_WARNINGS);
+
+    /// \class TempDirectory TempDirectory.hh ignitin/common/TempDirectory.hh
+    /// \brief Create a temporary directory in the OS temp location.
+    class TempDirectory
+    {
+      public: TempDirectory(const std::string &_prefix = "temp_dir",
```
ahcorde, line 60: document methods

src/DirIter.cc
```diff
@@ -0,0 +1,80 @@
...
+
+DirIter::DirIter(const std::string &_in):
+  DirIter()
+{
+  try {
+    this->dataPtr->it = fs::directory_iterator(_in);
+  }
+  catch (const fs::filesystem_error &ex)
+  {
+  }
```
ahcorde, line 48: Error msgs ?

ahcorde (changes requested):
I made some minor comments and I noticed that it's still a draft. 

Can you also test `ign-fuel-tools` ? just to check that we are not breaking things on Windows

src/DirIter.cc
mjcarroll: This doesn't really need an error message.  If the path is invalid, then this will be set to the `::end()` iterator, so any loops will end.

mjcarroll:
`ignition_common-ci-pr_any-homebrew-amd64` is expected to fail because `std::filesystem` isn't supported on 10.14

ahcorde (changes requested):
Conflicts

ahcorde (changes requested):
Mac and Windows are failing

mjcarroll:
> Mac and Windows are failing

Windows is failing because the `ign-utils` dependency introduced in #244.

It is expected that the Jenkins macOS build will fail, as it is using 10.14.  The new GitHub actions-based macOS build is the one we should be referring to from this PR forward.

mjcarroll:

chapulina:
Homebrew is not happy yet:

```
/Users/jenkins/workspace/ignition_common-ci-pr_any-homebrew-amd64/ign-common/src/DirIter.cc:31:15: error: 'directory_iterator' is unavailable: introduced in macOS 10.15
  public: fs::directory_iterator it;
```

mjcarroll:
> Homebrew is not happy yet:

Homebrew is expected to fail starting with this PR.  `std::filesystem` is not supported on macOS 10.14.  This is why I added the second actions-based CI against macOS 10.15

chapulina:
> Homebrew is expected to fail starting with this PR.

mjcarroll:
> So should we disable that job, or update it to use 10.15?

Probably a question best answered by @scpeters and @j-rivero .

My understanding is that 10.14 will go EOL sometime around the end of October, sometime after Monterey comes out.

At the very least, we should probably consider moving to 10.15 for all CI workers before 10.14 goes EOL.

chapulina:
I made homebrew checks not required for stable branches, ticketed https://github.com/ignitionrobotics/ign-common/issues/249 and updated https://github.com/ignition-tooling/release-tools/issues/398

chapulina:
Ignition officially supports gcc >= 8, but we have a project that's been using gcc 7 and skipping things like tests and some dependencies. I believe that merging this would break that use case.

We don't have an obligation to support that use case, but considering that this PR is not addressing a bug, how about retargeting to `main` so we don't break it?

mjcarroll:
This has now been targeted to `main`/`ign-common5` and could use another review.

nkoenig:
Should the DirIter class be marked as deprecated? There are some no-op functions, and it looks like we should directly use `std::filesystem`.

scpeters:
I'll start a windows CI / colcon build using this branch to see how it affects downstream packages

scpeters:
> I'll start a windows CI / colcon build using this branch to see how it affects downstream packages

but nothing depends on ignition-common5 yet, so never mind

mjcarroll:
> I'll start a windows CI / colcon build using this branch to see how it affects downstream packages

I did run a full build on my system when it was still targeting ign-common4 with no issues. Since this is largely exactly replicating existing functionality, I don't anticipate any problems.

mjcarroll:
> Should the DirIter class be marked as deprecated? There are some no-op functions, and it looks like we should directly use std::filesystem.

The same could be said with a few of the other functions in here.  This was originally intended to be a minimally-invasive drop in replacement for our current implementation.  

Since this is targeting `main`, we may consider doing this in two passes, the first to update everything to not use `FilesystemBoost`, and then another round for deprecations?

mjcarroll:
This is now ready for another review.  For ease of review, it is now split into two commits:

* f10377d - This commit is purely changed to the source code, with no corresponding test changes. This should demonstrate that the changes are compatible.
* 3a3f1cb - Changes to the tests to remove functionality that is replaced by std::filesystem  (This is in pr #282 )

src/DirIter.cc
```diff
@@ -0,0 +1,84 @@
...
+}
+
+//////////////////////////////////////////////////
+void DirIter::Next() {}
+
+//////////////////////////////////////////////////
+void DirIter::SetInternalEmpty() {}
+
+//////////////////////////////////////////////////
+void DirIter::CloseHandle() {}
```
chapulina, line 80: Do we need these empty functions? They're private, we could just remove them without breaking API

src/DirIter.cc
```diff
@@ -0,0 +1,84 @@
...
+namespace fs = std::filesystem;
+
+namespace ignition
+{
+namespace common
+{
+
+class DirIter::Implementation
+{
+  public: fs::directory_iterator it;
```
chapulina, line 31:
```suggestion
  /// \brief Filesystem iterator that this class is wrapping
  public: fs::directory_iterator it;
```

src/Filesystem.cc
```diff
@@ -36,466 +30,293 @@
...
 #endif
 
-namespace igncmn = ignition::common;
-using namespace ignition;
-using namespace igncmn;
-
-/////////////////////////////////////////////////
-bool ignition::common::isFile(const std::string &_path)
+// Return true if success, false if error
+inline bool fs_warn(const std::string &_fcn,
```
chapulina, line 44:
```suggestion
inline bool fsWarn(const std::string &_fcn,
```

src/Filesystem.cc
```diff
@@ -36,466 +30,293 @@
...
-# define IGN_PATH_MAX PATH_MAX
-#elif defined(_XOPEN_PATH_MAX)
-# define IGN_PATH_MAX _XOPEN_PATH_MAX
+namespace fs = std::filesystem;
+
+#ifndef _WIN32
+static const char preferred_separator = fs::path::preferred_separator;
 #else
-# define IGN_PATH_MAX _POSIX_PATH_MAX
+static const char preferred_separator = '\\';
```
chapulina, line 40: Do we need these static separators? What's the difference between this and the `::separator` function?

Also

```suggestion
static const char preferredSeparator = '\\';
```

chapulina (commented):
Great to see all that custom code going away! I just have minor comments below

src/DirIter.cc
mjcarroll: 50455b6

src/Filesystem.cc
mjcarroll: 7b0ddfc

src/Filesystem.cc
mjcarroll: Not sure, I think it was something windows related, I'll try to drop it and see what happens.

mjcarroll:

mjcarroll:
@chapulina down to three really suspicious failures here:

```
UNIT_KeyEvent_TEST.test_ran
UNIT_Pbr_TEST.test_ran
UNIT_Profiler_Remotery_TEST.test_ran
```

KeyEvent and `Profiler_Remotery` both segault and neither make use of filesystem, as far as I can tell.

chapulina:
> down to three really suspicious failures here:

Suspicious indeed. They all seem to segfault even before the test starts, so maybe it's something within GTest? Could the `CXX_FILESYSTEM_LIBRARIES` have affected it? I can't tell what these tests have in common that's specific to them. They all use `TEST` and don't have a `main`, but other tests do too and they're passing.

scpeters:
> > down to three really suspicious failures here:
> 
> Suspicious indeed. They all seem to segfault even before the test starts, so maybe it's something within GTest? Could the `CXX_FILESYSTEM_LIBRARIES` have affected it? I can't tell what these tests have in common that's specific to them. They all use `TEST` and don't have a `main`, but other tests do too and they're passing.

weird that the tests only fail in jenkins, not in Github actions

scpeters:
> > > down to three really suspicious failures here:
> > 
> > 
> > Suspicious indeed. They all seem to segfault even before the test starts, so maybe it's something within GTest? Could the `CXX_FILESYSTEM_LIBRARIES` have affected it? I can't tell what these tests have in common that's specific to them. They all use `TEST` and don't have a `main`, but other tests do too and they're passing.
> 
> weird that the tests only fail in jenkins, not in Github actions

those failures occurred on the r2d2 machine, which I have taken offline for testing. I see the following when I try to load one of the failing tests in gdb on r2d2:

~~~
Reading symbols from bin/UNIT_Pbr_TEST...
(gdb) r
Starting program: /var/lib/jenkins/workspace/ignition_common-ci-pr_any-ubuntu_auto-amd64/build/bin/UNIT_Pbr_TEST 
warning: Error disabling address space randomization: Operation not permitted
warning: Could not trace the inferior process.
warning: ptrace: Operation not permitted
During startup program exited with code 127.
~~~

scpeters:
I downloaded the `Dockerfile` and `build.sh` script from the jenkins job and with a few modifications was able to get it running on my local machine. I collected the following backtrace from `UNIT_Pbr_TEST`:

~~~
Reading symbols from bin/UNIT_Pbr_TEST...
(gdb) r
Starting program: /home/jenkins/workspace/ignition_common-ci-pr_any-ubuntu_auto-amd64/build/bin/UNIT_Pbr_TEST 
warning: Error disabling address space randomization: Operation not permitted
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
std::filesystem::__cxx11::path::~path (this=0x0, __in_chrg=<optimized out>)
   at /usr/include/c++/8/bits/fs_path.h:209
209      ~path() = default;
(gdb) thread apply all bt

Thread 1 (Thread 0x7fb0423e1c40 (LWP 4110)):
#0 std::filesystem::__cxx11::path::~path (this=0x0, __in_chrg=<optimized out>) at /usr/include/c++/8/bits/fs_path.h:209
#1 0x00007fb0438d6434 in std::filesystem::__cxx11::path::_Cmpt::~_Cmpt (this=0x0, __in_chrg=<optimized out>) at /usr/include/c++/8/bits/fs_path.h:644
#2 std::_Destroy<std::filesystem::__cxx11::path::_Cmpt> (__pointer=0x0) at /usr/include/c++/8/bits/stl_construct.h:98
#3 std::_Destroy_aux<false>::__destroy<std::filesystem::__cxx11::path::_Cmpt*> (__last=<optimized out>, __first=0x0) at /usr/include/c++/8/bits/stl_construct.h:108
#4 std::_Destroy<std::filesystem::__cxx11::path::_Cmpt*> (__last=<optimized out>, __first=<optimized out>) at /usr/include/c++/8/bits/stl_construct.h:137
#5 std::_Destroy<std::filesystem::__cxx11::path::_Cmpt*, std::filesystem::__cxx11::path::_Cmpt> (__last=0x2, __first=<optimized out>) at /usr/include/c++/8/bits/stl_construct.h:206
#6 std::vector<std::filesystem::__cxx11::path::_Cmpt, std::allocator<std::filesystem::__cxx11::path::_Cmpt> >::~vector (this=0x55e9e44c34b0, __in_chrg=<optimized out>) at /usr/include/c++/8/bits/stl_vector.h:567
#7 std::filesystem::__cxx11::path::~path (this=0x55e9e44c3490, __in_chrg=<optimized out>) at /usr/include/c++/8/bits/fs_path.h:209
#8 std::filesystem::__cxx11::path::_Cmpt::~_Cmpt (this=0x55e9e44c3490, __in_chrg=<optimized out>) at /usr/include/c++/8/bits/fs_path.h:644
#9 std::_Destroy<std::filesystem::__cxx11::path::_Cmpt> (__pointer=0x55e9e44c3490) at /usr/include/c++/8/bits/stl_construct.h:98
#10 std::_Destroy_aux<false>::__destroy<std::filesystem::__cxx11::path::_Cmpt*> (__last=<optimized out>, __first=0x55e9e44c3490) at /usr/include/c++/8/bits/stl_construct.h:108
#11 std::_Destroy<std::filesystem::__cxx11::path::_Cmpt*> (__last=<optimized out>, __first=<optimized out>) at /usr/include/c++/8/bits/stl_construct.h:137
#12 std::_Destroy<std::filesystem::__cxx11::path::_Cmpt*, std::filesystem::__cxx11::path::_Cmpt> (__last=0x0, __first=0x55e9e44c3490) at /usr/include/c++/8/bits/stl_construct.h:206
--Type <RET> for more, q to quit, c to continue without paging--c
#13 std::vector<std::filesystem::__cxx11::path::_Cmpt, std::allocator<std::filesystem::__cxx11::path::_Cmpt> >::~vector (this=<synthetic pointer>, __in_chrg=<optimized out>) at /usr/include/c++/8/bits/stl_vector.h:567
#14 std::vector<std::filesystem::__cxx11::path::_Cmpt, std::allocator<std::filesystem::__cxx11::path::_Cmpt> >::_M_move_assign (__x=..., this=0x7ffe85f32920) at /usr/include/c++/8/bits/stl_vector.h:1679
#15 std::vector<std::filesystem::__cxx11::path::_Cmpt, std::allocator<std::filesystem::__cxx11::path::_Cmpt> >::operator= (__x=..., this=0x7ffe85f32920) at /usr/include/c++/8/bits/stl_vector.h:601
#16 std::filesystem::__cxx11::path::operator= (__p=..., this=0x7ffe85f32900) at /usr/include/c++/8/bits/fs_path.h:797
#17 std::filesystem::__cxx11::path::operator= (__p=..., this=0x7ffe85f32900) at /usr/include/c++/8/bits/fs_path.h:791
#18 ignition::common::joinPaths (_path1="/home/jenkins", _path2=".ignition") at /home/jenkins/workspace/ignition_common-ci-pr_any-ubuntu_auto-amd64/ign-common/src/Filesystem.cc:143
#19 0x00007fb0438e3cf9 in ignition::common::SystemPaths::SystemPaths (this=0x55e9e44c3460) at /usr/include/c++/8/bits/basic_string.h:252
#20 0x00007fb0438cccc5 in __static_initialization_and_destruction_0 (__initialize_p=1, __priority=65535) at /home/jenkins/workspace/ignition_common-ci-pr_any-ubuntu_auto-amd64/ign-common/src/Util.cc:564
#21 _GLOBAL__sub_I_Util.cc(void) () at /home/jenkins/workspace/ignition_common-ci-pr_any-ubuntu_auto-amd64/ign-common/src/Util.cc:564
#22 0x00007fb043e2eb8a in ?? () from /lib64/ld-linux-x86-64.so.2
#23 0x00007fb043e2ec91 in ?? () from /lib64/ld-linux-x86-64.so.2
#24 0x00007fb043e1e13a in ?? () from /lib64/ld-linux-x86-64.so.2
#25 0x0000000000000001 in ?? ()
#26 0x00007ffe85f34d24 in ?? ()
#27 0x0000000000000000 in ?? ()
(gdb)
~~~

src/Util.cc
```diff
@@ -55,8 +57,19 @@
...
+///
+/// This uses the NeverDestroyed pattern to prevent static initialization and
+/// destruction order fiasco issues.
+ignition::common::SystemPaths* GetSystemPaths()
+{
+  static const
+    ignition::utils::NeverDestroyed<ignition::common::SystemPaths*> paths{
+      new ignition::common::SystemPaths()
+    };
+  return paths.Access();
```
azeey, line 71: I was under the impression you can just do 
```c++
  static const ignition::utils::NeverDestroyed<ignition::common::SystemPaths> paths;
  return &paths.Access();
```

without the `new`.

mjcarroll: Ah, good point, I think this may have been a workaround for something `const`.  Will fix.

mjcarroll:
Waiting on https://github.com/osrf/homebrew-simulation/pull/1796

scpeters:
> Waiting on [osrf/homebrew-simulation#1796](https://github.com/osrf/homebrew-simulation/pull/1796)

I just merged it and restarted the homebrew build

src/SystemPaths.cc
```diff
@@ -403,6 +390,8 @@ std::string SystemPaths::FindFile(const std::string &_filename,
   // Try appending to local paths
   else
   {
+    ignerr << "Here: " << cwd() << " " << filename << std::endl;
```
scpeters, line 393: remove this debugging message

src/SystemPaths.cc
```diff
@@ -465,9 +454,12 @@ std::string SystemPaths::FindFile(const std::string &_filename,
 std::string SystemPaths::LocateLocalFile(const std::string &_filename,
     const std::vector<std::string> &_paths)
 {
+  std::cerr << "LocateLocalFile: " << _filename << "\n";
```
scpeters, line 457: remove this debugging message

src/SystemPaths.cc
```diff
@@ -465,9 +454,12 @@ std::string SystemPaths::FindFile(const std::string &_filename,
...
+
   std::string foundPath = "";
   for (auto const &path : _paths)
   {
+    std::cout << "\tPath: " << path << "\n";
```
scpeters, line 462: remove this debugging message

src/SystemPaths.cc
```diff
@@ -531,6 +523,8 @@ void SystemPaths::AddFindFileURICallback(
 /////////////////////////////////////////////////
 std::list<std::string> SystemPaths::PathsFromEnv(const std::string &_env)
 {
+  std::cerr << "PathsFromEnv" << std::endl;
```
scpeters, line 526: remove this debugging message

scpeters (commented):
a few debug messages can be removed

test/performance/logging.cc
```diff
@@ -196,7 +196,7 @@ TEST_P(LoggingTest, RunThreads)
 }
 
 INSTANTIATE_TEST_SUITE_P(LoggingTest, LoggingTest,
-                         ::testing::Values(1, 2, 4, 8, 16, 32));
+                         ::testing::Values(1, 2, 4, 8, 16));
```
scpeters, line 199: these changes will conflict with #302

mjcarroll: Let's get the forward port done first, and then I will sort it out.

src/SystemPaths.cc
scpeters: this one too?

test/performance/logging.cc
scpeters: this is close; I'll just resolve the conflicts when merging forward

src/SystemPaths.cc
scpeters: removed in https://github.com/ignitionrobotics/ign-common/pull/238/commits/6ab5b607bd7d36acea1fec722965a13bb1d507dc

src/Filesystem.cc
```diff
@@ -36,466 +30,288 @@
...
+  else
+    is_url = true;
 
-  // This is a Windows path, convert all '/' into backslashes
-  std::string result = std::regex_replace(_path, std::regex(R"(/)"), "\\");
-  std::string drive_letters;
+  if (_path2.find("://") == std::string::npos)
+    p2 = p2.lexically_normal();
+  else
+    is_url = true;
```
scpeters, line 150: shouldn't it be an error if `_path2` is a URL? I guess unless `_path1` is empty. Maybe our existing tests don't address that, but it seems like a problem

scpeters (commented):
two comments about Filesystem.cc

test/performance/logging.cc
scpeters: forward port is done, feel free to resolve conflicts

src/Filesystem.cc
mjcarroll: You are correct that it's strange.  It _likely_ should happen, but there is nothing stopping users from doing it.

I think I'd like to address this in a follow-up PR as it's out of the scope of our current test suite.

src/Filesystem.cc
```diff
@@ -36,466 +30,288 @@
...
+  return fsWarn("createDirectory", ec);
+}
 
-  return ignition::common::removeDirectoryOrFile(_path, _warningOp);
+/////////////////////////////////////////////////
+bool ignition::common::createDirectories(const std::string &_path)
+{
+  std::error_code ec;
+  // Disregard return here because it may return false if the
+  // directory is not actually created (already exists)
```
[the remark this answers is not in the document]
mjcarroll: I wasn't actually doing anything with the `bool` return.  I think I have made it more clear in [66f716c](https://github.com/ignitionrobotics/ign-common/pull/238/commits/66f716ca291a170b261a49cdb9cad4d90cba2859)

src/Filesystem.cc
scpeters: can you add a todo comment before merging this?

mjcarroll: Done.

src/Filesystem.cc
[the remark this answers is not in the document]
mjcarroll: Reworded to indicate which return value I'm talking about.

scpeters (approved):

PR #267, opened 2021-11-19
Geospatial component for heightmap & DEMs

jennuine:
# 🎉 New feature

* Part of https://github.com/ignitionrobotics/ign-gazebo/issues/235

## Summary

- Moved heightmap and DEM classes to a new geospatial component, which depends on graphics
- GDAL is a required dependency for the geospatial component
- Ported over DEM updates & tests from `gazebo11` except for [DemTest.NegDem](https://github.com/osrf/gazebo/blob/135a8c1fd9f28e18c0daa171bc82ad9afdb9883d/gazebo/common/Dem_TEST.cc#L171) since https://github.com/osrf/gazebo/issues/2881 is still an issue and will hopefully be resolved in a future PR

## Test it

```bash
/path/to/build/bin/UNIT_Dem_TEST
```

## Checklist
- [x] Added tests


geospatial/src/Dem.cc
```diff
@@ -103,46 +97,55 @@ int Dem::Load(const std::string &_filename)
...
-
-  // Set the world width and height
-  this->dataPtr->worldWidth =
-     math::SphericalCoordinates::Distance(upLeftLat, upLeftLong,
-                                            upRightLat, upRightLong);
-  this->dataPtr->worldHeight =
-     math::SphericalCoordinates::Distance(upLeftLat, upLeftLong,
-                                            lowLeftLat, lowLeftLong);
+  try
+  {
```
chapulina, line 122: I know this was ported straight from Gazebo classic, but can this try block be narrower, including just the lines that throw an exception?

geospatial/src/Dem.cc
```diff
@@ -175,9 +201,11 @@ double Dem::Elevation(double _x, double _y)
 {
   if (_x >= this->Width() || _y >= this->Height())
   {
-    gzthrow("Illegal coordinates. You are asking for the elevation in (" <<
-          _x << "," << _y << ") but the terrain is [" << this->Width() <<
-           " x " << this->Height() << "]\n");
+    throw std::invalid_argument(
```
chapulina, line 204: We should consider getting rid of these exceptions and printing errors instead. We don't throw exceptions often.

chapulina (commented):
Great work! It would be good to test compiling a downstream library that uses heightmap against this branch to make sure all the includes are found and everything links correctly.

There are some CI failures, probably due to a different GDAL version on Bionic?

jennuine:
Thanks for the review!

> It would be good to test compiling a downstream library that uses heightmap against this branch to make sure all the includes are found and everything links correctly

I was able to run `ign gazebo heightmap.sdf` in a Garden workspace without any issues. Does this suffice?

> There are some CI failures, probably due to a different GDAL version on Bionic?

Yes, this is the likely cause but are we dropping Bionic support for Garden?

chapulina:
> I was able to run ign gazebo heightmap.sdf in a Garden workspace without any issues. Does this suffice?

Cool! So you updated all locally libraries to use `ign-common5`?

> Yes, this is the likely cause but are we dropping Bionic support for Garden?

Ahh good call. I didn't realize that would bear fruit so early. We need to update CI

jennuine:
> Cool! So you updated all locally libraries to use ign-common5?

Yup, specifically `ign-fuel-tools`, `ign-gazebo`, `ign-gui`, `ign-launch`, `ign-physics`, `ign-rendering`, and `ign-sensors`.

> Ahh good call. I didn't realize that would bear fruit so early. We need to update CI

Should I create an issue somewhere?

chapulina:
> Should I create an issue somewhere?

I created some issues:

* https://github.com/ignition-tooling/release-tools/pull/565
* https://github.com/ignition-tooling/release-tools/issues/566

Also see https://github.com/ignitionrobotics/ign-common/pull/270

geospatial/src/Dem.cc
jennuine: I've updated it to an error instead and am returning `inf` if reached. What do you think? 6f6769aedb2ca1689d3a08f295036edec461c099

geospatial/src/Dem.cc
jennuine: I've gone ahead and removed the `try/catch` block and replaced it with an error instead. Also, instead of continuing through the code, it returns when the error is reached 6f6769aedb2ca1689d3a08f295036edec461c099

scpeters:
I think it definitely makes sense to move `Dem.hh` to the `geospatial` component, but I'm not sure about moving `HeightmapData.hh` and `ImageHeightmap.hh` out of `graphics`. `gdal` isn't used by `ImageHeightmap`, but moving this code to `geospatial` would add that dependency to this code.

jennuine:
> I think it definitely makes sense to move `Dem.hh` to the `geospatial` component, but I'm not sure about moving `HeightmapData.hh` and `ImageHeightmap.hh` out of `graphics`. `gdal` isn't used by `ImageHeightmap`, but moving this code to `geospatial` would add that dependency to this code.

That makes sense. @chapulina do you have any thoughts before I make the changes?

chapulina:
I agree that requiring GDAL to use image heightmaps may be undesired, but when we have a `geospatial` component, I think it will be confusing for users that not all heightmap classes are in it, and some are actually in `graphics`.

The reason to add the `geospatial` component was to avoid adding that dependency to all `graphics` users, but we chose that name because we wanted the component to be semantically meaningful, and not just referent to a dependency. We need to draw the line at some point and inevitably some users will always get a bit more than they actually need. That's true for other `ign-common` components and other Ignition libraries too. Also note that all our "internal" users (that is, all Ignition libraries that will use this: rendering and physics) will support DEMs anyway.

I'm leaning towards being more semantically precise instead of trying to reduce the dependency surface for hypothetical users.

jennuine:
> I'm leaning towards being more semantically precise instead of trying to reduce the dependency surface for hypothetical users.

I agree, I think this will avoid confusion for users.

scpeters:
> I agree that requiring GDAL to use image heightmaps may be undesired, but when we have a `geospatial` component, I think it will be confusing for users that not all heightmap classes are in it, and some are actually in `graphics`.
> 
> The reason to add the `geospatial` component was to avoid adding that dependency to all `graphics` users, but we chose that name because we wanted the component to be semantically meaningful, and not just referent to a dependency. We need to draw the line at some point and inevitably some users will always get a bit more than they actually need. That's true for other `ign-common` components and other Ignition libraries too. Also note that all our "internal" users (that is, all Ignition libraries that will use this: rendering and physics) will support DEMs anyway.
> 
> I'm leaning towards being more semantically precise instead of trying to reduce the dependency surface for hypothetical users.

if users depend on the `geospatial` component, then they will get `ImageHeightmap` either way. I just wanted to point out that removing the currently functional heightmap classes from `graphics` will require migration action for downstream packages. If we proceed in this manner, then we need to update the migration guide to reflect this

chapulina:
> I just wanted to point out that removing the currently functional heightmap classes from graphics will require migration action for downstream packages.

Ok, that's a good point, users currently depending on `graphics` would need to update to depend on `geospatial`.

At least this PR is targeting `main`. Ideally we'd tick-tock the change, but in this case I think it could get complicated.

scpeters:
> > I just wanted to point out that removing the currently functional heightmap classes from graphics will require migration action for downstream packages.
> 
> Ok, that's a good point, users currently depending on `graphics` would need to update to depend on `geospatial`.
> 
> At least this PR is targeting `main`. Ideally we'd tick-tock the change, but in this case I think it could get complicated.

so we should update the migration guide?

chapulina:
> so we should update the migration guide?

That seems the most natural thing to me. When users update their `find_package` calls from `comon4` to `common5`, they also need to add the `geospatial` component. I know it's not ideal, but some things are tough to tick-tock.

jennuine:
I have updated `Migration.md` 90e55a19f2616c8a19d880eb2193d5aee9b21625

geospatial/src/Dem.cc
scpeters: yeah we had to throw an exception in gazebo11 since the `GeoReference` method returned `void` and its API was already frozen. I notice that you changed that API to return `bool` in this PR, which is much better. 👍

geospatial/src/Dem.cc
```diff
@@ -161,11 +167,34 @@ int Dem::Load(const std::string &_filename)
...
 
-  // Set the min/max heights
-  this->dataPtr->minElevation = *std::min_element(&this->dataPtr->demData[0],
-      &this->dataPtr->demData[0] + this->dataPtr->side * this->dataPtr->side);
-  this->dataPtr->maxElevation = *std::max_element(&this->dataPtr->demData[0],
-      &this->dataPtr->demData[0] + this->dataPtr->side * this->dataPtr->side);
+  // Check for nodata value in dem data. This is used when computing the
+  // min elevation. If nodata value is not defined, we assume it will be one
+  // of the commonly used values such as -9999, -32768, etc.
+  // For simplicity, we will treat values <= -9999 as nodata values and
```
scpeters, line 173: is there documentation we can reference about these values?

geospatial/src/Dem.cc
```diff
@@ -161,11 +167,34 @@ int Dem::Load(const std::string &_filename)
...
+    noDataValue = defaultNoDataValue;
+
+  double min = ignition::math::MAX_D;
+  double max = -ignition::math::MAX_D;
+  for (auto d : this->dataPtr->demData)
+  {
+    if (d < min && d > noDataValue)
+      min = d;
+    if (d > max && d > noDataValue)
+      max = d;
```
scpeters, line 188: an alternative:

~~~
if (d > noDataValue)
{
  if (d < min)
    min = d;
  else if (d > max)
    max = d
}
~~~

geospatial/src/Dem.cc
jennuine: I haven't been able to find documentation on this other than this blog post https://rashms.com/gis/raster-file-resolution-extent-no-data-value-in-qgis/

geospatial/src/Dem.cc
jennuine: 41efc61bbd6eb6a9470ad5bafdff021cda203c8c

geospatial/src/Dem.cc
scpeters: I found a reference to -9999 in the ArcGIS documentation: https://desktop.arcgis.com/en/arcmap/10.8/manage-data/raster-and-images/nodata-in-raster-datasets.htm

how do you feel about adding that link?

jennuine: Works for me, thanks for finding 5eadb52caf977cf1b891a8d00c00a1c48ea368b7

scpeters (approved):

mjcarroll (commented):
One comment about error messages

geospatial/src/Dem.cc
```diff
@@ -89,11 +87,13 @@ int Dem::Load(const std::string &_filename)
...
   if (!exists(findFilePath(fullName)))
-    fullName = common::find_file(_filename);
+    fullName = common::findFile(_filename);
+
+  this->dataPtr->filename = fullName;
 
   if (!exists(findFilePath(fullName)))
   {
-    gzerr << "Unable to open DEM file[" << _filename
+    ignerr << "Unable to open DEM file[" << _filename
```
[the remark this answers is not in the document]
jennuine: Thanks for catching. I tried setting `IGN_FILE_PATH` but it did not help locating the file. I went ahead and removed the statement from the error message 05744fc0a05c3c1f11f27b5bdd261616bd936b79

scpeters:
I think this will require an update to https://github.com/ignition-release/ign-common5-release when it is merged

mjcarroll (approved):

jennuine:
Please wait to merge, I may add a small update.

PR #272, opened 2021-11-26
Use libexec to install binary remotery_vis

### LINKED ISSUE #271: Remotery visualizator is not being packaged in ign-common4 packages
Seems like during the transition to ign-common4 the ign profiler tool remotery visualizator was not packaged. Seems like we would have to enable the option `IGN_PROFILER_REMOTERY` in our packaging. Was this a bug or it was intentional?

j-rivero:
# 🎉 New feature

Closes #271

## Summary
We are installing the profiler visualizer binary under the standard library path which is something that does not fit well according to the [FHR](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrlibexec). libexec seems a better place for them.

> 4.7.1. Purpose
> /usr/libexec includes internal binaries that are not intended to be executed directly by users or shell scripts. Applications may use a single subdirectory under /usr/libexec.

### DIFF
--- a/profiler/src/CMakeLists.txt
+++ b/profiler/src/CMakeLists.txt
@@ -88,7 +88,7 @@ if(TARGET UNIT_Profiler_Remotery_TEST)
 endif()
 
 if(IGN_PROFILER_REMOTERY)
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

mjcarroll (approved):

osrf-triage:

PR #273, opened 2021-11-30
Remove all deprecated functionality from main

mjcarroll:
Remove functionality previously marked as `deprecated` from `main`/`ign-common5`


chapulina (approved):

mjcarroll:
These tests weren't failing locally, going to take another look.

mjcarroll:
Reproduced locally, this ended up requiring removing some tests that (I believe) are no longer relevant, since we dropped some of the callbacks.

I would appreciate a second set of eyes on those modifications, though.

chapulina (approved):
The test changes look good to me too :+1:

PR #274, opened 2021-11-30
Use ImplPtr where relevant

mjcarroll:
Since we can add a dependency on `ign-utils1`, update to use the `ImplPtr` pattern instead of `std::unique_ptr` where relevant.

### DIFF
--- a/include/ignition/common/Battery.hh
+++ b/include/ignition/common/Battery.hh
@@ -22,15 +22,13 @@
 #include <functional>
 #include <memory>
 #include <ignition/common/Export.hh>
-#include <ignition/common/SuppressWarning.hh>
+
+#include <ignition/utils/ImplPtr.hh>
 
 namespace ignition
 {
   namespace common
   {
-    // Forward declare private data class.
-    class BatteryPrivate;
-
     /// \brief A battery abstraction
     ///
     /// The default battery model is ideal: It just takes the initial voltage
@@ -59,13 +57,22 @@ namespace ignition
       /// \param[in] _battery Battery to copy.
       public: Battery(const Battery &_battery);
 
-      /// \brief Assignment operator
+      /// \brief Copy assignment operator
       /// \param[in] _battery The new battery
       /// \return a reference to this instance
       public: Battery &operator=(const Battery &_battery);
 
+      /// \brief Move constructor
+      /// \param[in] _battery Battery to move.
+      public: Battery(Battery &&_battery);
+
+      /// \brief Move assignment operator
+      /// \param[in] _battery The new battery
+      /// \return a reference to this instance
+      public: Battery &operator=(Battery &&_battery);
+
       /// \brief Destructor.
-      public: virtual ~Battery();
+      public: virtual ~Battery() = default;
 
       /// \brief Equal to operator
       /// \param[in] _battery the battery to compare to
@@ -163,11 +170,8 @@ namespace ignition
       /// \return New battery voltage.
       private: double UpdateDefault(Battery *_battery);
 
-      IGN_COMMON_WARN_IGNORE__DLL_INTERFACE_MISSING
-      /// \internal
-      /// \brief Private data pointer.
-      private: std::unique_ptr<BatteryPrivate> dataPtr;
-      IGN_COMMON_WARN_RESUME__DLL_INTERFACE_MISSING
+      /// \brief Pointer to private data.
+      IGN_UTILS_UNIQUE_IMPL_PTR(dataPtr)
     };
 
     /// \def BatteryPtr
--- a/include/ignition/common/Battery.hh
+++ b/include/ignition/common/Battery.hh
@@ -17,10 +17,11 @@
 #ifndef IGNITION_COMMON_BATTERY_HH_
 #define IGNITION_COMMON_BATTERY_HH_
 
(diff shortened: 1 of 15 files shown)


include/ignition/common/Battery.hh
```diff
@@ -59,13 +57,22 @@ namespace ignition
...
       public: Battery &operator=(const Battery &_battery);
 
+      /// \brief Move constructor
+      /// \param[in] _battery Battery to move.
+      public: Battery(Battery &&_battery);
+
+      /// \brief Move assignment operator
+      /// \param[in] _battery The new battery
+      /// \return a reference to this instance
+      public: Battery &operator=(Battery &&_battery);
```
chapulina, line 72: Shouldn't the `ImplPtr` let us use rule-of-zero?

include/ignition/common/Time.hh
```diff
@@ -1,324 +0,0 @@
-/*
```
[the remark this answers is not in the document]
mjcarroll: This includes the commit from #273, so once that lands this diff will shrink.

This is a better diff: https://github.com/ignitionrobotics/ign-common/compare/remove_main_deprecations...main_implptr

include/ignition/common/Battery.hh
mjcarroll: In general, it will.  Unfortunately, `BatteryPrivate` has a `std::mutex` in it, which has no copy or move constructors, so we have to implement the full rule of 5.

In retrospect, I think that the locking should have happened outside of `Battery` by the downstream user, but I think that's a more substantial change that should come in a different PR.

ahcorde (dismissed):
CI is failing, it not able to find this header:

```
     26 | #include <ignition/utils/ImplPtr.hh>
```

mjcarroll:
> CI is failing, it not able to find this header:

I think we have to make `ign-utils` a core dependency.  @j-rivero what are the packaging implications?

chapulina:
> I think we have to make ign-utils a core dependency

I think it's already there since `ign-common4`

* https://github.com/ignition-release/ign-common4-release/pull/4
* https://github.com/osrf/homebrew-simulation/pull/1313

include/ignition/common/Filesystem.hh
```diff
@@ -22,7 +22,8 @@
 #include <string>
```
chapulina, line 22: Can remove `memory`?

include/ignition/common/SystemPaths.hh
```diff
@@ -34,9 +34,10 @@
 #include <vector>
```
chapulina, line 34: Here too remove `memory`? The same goes for various other files.

include/ignition/common/TempDirectory.hh
```diff
@@ -21,15 +21,14 @@
 #include <memory>
 #include <string>
 
-#include <ignition/common/Export.hh>
```
chapulina, line 24: `Export` is needed for `IGNITION_COMMON_VISIBLE`

src/URI_TEST.cc
```diff
@@ -890,8 +890,8 @@ TEST(URITEST, HasAuthority)
     // Modifyng path updates string
     uri.Path() = URIPath("new_authority.com/another/path");
 
-    EXPECT_EQ("new_authority.com/another/path/", uri.Path().Str());
-    EXPECT_EQ("https://new_authority.com/another/path/", uri.Str());
+    EXPECT_EQ("new_authority.com/another/path", uri.Path().Str());
+    EXPECT_EQ("https://new_authority.com/another/path", uri.Str());
```
chapulina, line 894: I don't understand why these expectations would change :thinking:

mjcarroll: Ah yes, I forgot that I did this.  I don't understand, either...

mjcarroll: I found out why.  By using the automatically generated functions, we get one that copies all of the members.

The issue is that the `URIPath::Implementation` struct has a `trailingSlash` member: https://github.com/ignitionrobotics/ign-common/blob/ign-common4/src/URI.cc#L54-L73

The originally written copy-assignment operator doesn't copy this member:
https://github.com/ignitionrobotics/ign-common/blob/ign-common4/src/URI.cc#L586-L591

And the test exercises this copy-assignment: 
https://github.com/ignitionrobotics/ign-common/blob/ign-common4/src/URI_TEST.cc#L890-L894

In this case, `ImplPtr` has shown that the previous behavior is actually a bug, and we should fix it in previous versions.

mjcarroll: Oh, all these refs were to ign-common4, but the same is true on main

include/ignition/common/TempDirectory.hh
mjcarroll: 4edd805

include/ignition/common/SystemPaths.hh
mjcarroll: 4edd805

src/URI_TEST.cc
chapulina: Ohhhh so this PR includes more than QOL, it's also a bugfix! Nice detective work :+1:

chapulina (approved):

PR #275, opened 2021-12-03
Fix bug in URIPath assignment operator

mjcarroll:
# 🦟 Bug fix

Discovered in #274.

## Summary

The issue is that the `URIPath::Implementation` struct has a `trailingSlash` member: https://github.com/ignitionrobotics/ign-common/blob/ign-common4/src/URI.cc#L54-L73

The originally written copy-assignment operator doesn't copy this member:
https://github.com/ignitionrobotics/ign-common/blob/ign-common4/src/URI.cc#L586-L591

And the test exercises this copy-assignment: 
https://github.com/ignitionrobotics/ign-common/blob/ign-common4/src/URI_TEST.cc#L890-L894

In this case, `ImplPtr` has shown that the previous behavior is actually a bug, and we should fix it in previous versions.

### DIFF
--- a/src/URI.cc
+++ b/src/URI.cc
@@ -587,6 +587,7 @@ URIPath &URIPath::operator=(const URIPath &_path)
 {
   this->dataPtr->path = _path.dataPtr->path;
   this->dataPtr->isAbsolute = _path.dataPtr->isAbsolute;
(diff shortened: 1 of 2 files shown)


ahcorde (approved):

osrf-triage:

PR #282, opened 2021-12-21
Update tests to use more STL functionality

mjcarroll:
Test changes for #238, broken out for easier review.

### DIFF
--- a/src/Filesystem_TEST.cc
+++ b/src/Filesystem_TEST.cc
@@ -17,204 +17,96 @@
 
 #include <gtest/gtest.h>
 
-#ifndef _WIN32
-#include <fcntl.h>
-#include <limits.h>
-#include <stdlib.h>
-#include <sys/stat.h>
-#include <sys/types.h>
-#include <unistd.h>
+#include <ignition/utilities/ExtraTestMacros.hh>
 
-#include "ignition/common/Util.hh"
+#include <ignition/common/Console.hh>
+#include <ignition/common/Filesystem.hh>
+#include <ignition/common/TempDirectory.hh>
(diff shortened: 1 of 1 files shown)


src/Filesystem_TEST.cc
```diff
@@ -293,27 +200,29 @@ TEST_F(FilesystemTest, fileOps)
...
 TEST_F(FilesystemTest, exists)
 {
-  std::string new_temp_dir;
-  ASSERT_TRUE(create_and_switch_to_temp_dir(new_temp_dir));
   ASSERT_TRUE(create_new_empty_file("newfile"));
   ASSERT_TRUE(createDirectory("fstestexists"));
 
   const auto &relativeSubdir = joinPaths("fstestexists", "newfile2");
-  const auto &absoluteSubdir = joinPaths(new_temp_dir, relativeSubdir);
+  const auto &absoluteSubdir = joinPaths(cwd(), relativeSubdir);
```
varunagrawal, line 225: How would this play when the user has a non-standard build path? I believe it should be fine, but just something to think about.

varunagrawal (approved):
This PR LGTM sans one minor comment.

src/Filesystem_TEST.cc
mjcarroll: The test is still executing in a temp directory, but that functionality has been moved to the fixture: https://github.com/ignitionrobotics/ign-common/pull/282/files#diff-cbe626f392b0654d4e901fc284e80468c4194ab344d0187fb391203e21d65a58R43-R44

The behavior shouldn't change at all.

PR #283, opened 2021-12-22
ign-common5: update to use ign-math7

mjcarroll:


src/Filesystem.cc
```diff
@@ -161,8 +161,14 @@ std::string ignition::common::joinPaths(
 
   auto ret = (p1 / p2);
 
-  if (!is_url)
+  if (is_url)
+  {
+    ret = copyToUnixPath(ret);
```
chapulina, line 166: This looks like a change from a different branch that got mixed here

chapulina:
I cherry-picked edad65faf5d674643b8ff88c93a881aee2614f41 into #285

PR #285, opened 2021-12-27
Bumps in garden: use ignition-math7

chapulina:
See https://github.com/ignition-tooling/release-tools/issues/574

### DIFF
--- a/src/Filesystem.cc
+++ b/src/Filesystem.cc
@@ -242,8 +242,16 @@ std::string checkWindowsPath(const std::string _path)
   std::string result = std::regex_replace(_path, std::regex(R"(/)"), "\\");
   std::string drive_letters;
 
+<<<<<<< HEAD
   // only Windows contains absolute paths starting with drive letters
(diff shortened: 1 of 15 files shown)


src/Filesystem.cc
```diff
@@ -242,8 +242,16 @@ std::string checkWindowsPath(const std::string _path)
...
+<<<<<<< HEAD
   // only Windows contains absolute paths starting with drive letters
   if (result.length() > 3 && 0 == result.compare(1, 2, ":\\"))
+=======
+  if (is_url)
+  {
+    ret = copyToUnixPath(ret);
+  }
+  else
+>>>>>>> edad65f... Fix deprecations when using ign-math7
```
scpeters, line 254: looks like a merge conflict?

chapulina: argh I thought I had fixed that

chapulina: I just hadn't pushed :grimacing:

scpeters (approved):

PR #293, opened 2022-01-21
Reduce verbosity of logging performance test

scpeters:
# 🦟 Bug fix

Improve readability of CI console log files

## Summary

The `PERFORMANCE_logging` test currently dominates our CI output logs. As an example, scroll through the following log and note how much of the log is prefixed with [47]

* https://build.osrfoundation.org/job/ignition_common-ci-ign-common4-focal-amd64/2/consoleText

The eliminates a `std::cout` call and replaces an `ignmsg` with `ignlog`. Please confirm if the test is still valid with these changes.

## Checklist
- [X] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [X] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [X] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #294, opened 2022-01-24
Move performance test to examples

mjcarroll:
This performance test flooded the console during CI without adding much value.

This was initially incorporated to prove out worst-case latency when adding locking mechanisms to the console log.  Since it doesn't actually make any assertions about behavior or performance, I believe that it makes more sense to live as an example.

The example will generate a CSV file that can be visualized as a histogram to show console logging latency with a variable number of threads.

Alternative to #293


chapulina (approved):
Makes sense :+1:

scpeters:
> Failing test of focal action is another somewhat-flaky performance test

restarted

PR #296, opened 2022-01-25
Utilize ImplPtr in AV component

mjcarroll:
Use ImplPtr where relevant in the AV component in common.

### DIFF
--- a/av/include/ignition/common/AudioDecoder.hh
+++ b/av/include/ignition/common/AudioDecoder.hh
@@ -22,24 +22,18 @@
 #include <memory>
 
 #include <ignition/common/av/Export.hh>
-#include <ignition/utils/SuppressWarning.hh>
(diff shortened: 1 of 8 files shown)


av/include/ignition/common/AudioDecoder.hh
```diff
@@ -22,25 +22,19 @@
 #include <memory>
```
iche033, line 22: I think this header can be removed now?

iche033: same for the other `<memory>` includes in this PR

iche033 (approved):
just one very minor comment otherwise looks good.

av/include/ignition/common/AudioDecoder.hh
mjcarroll: Ah, good catch, thanks.

PR #297, opened 2022-01-25
Miscellaneous cleanup of Events API

mjcarroll:
There were two issues with the Events API:

1. There were several places where we were using `const primitive` arguments to functions. https://clang.llvm.org/extra/clang-tidy/checks/readability-avoid-const-params-in-decls.html
2. There were several places where we marked a `SetXXX` function as `const` where we mutated internal data, breaking const-correctness.

These are split in two commits for easier review.

### DIFF
--- a/events/include/ignition/common/Event.hh
+++ b/events/include/ignition/common/Event.hh
@@ -53,7 +53,7 @@ namespace ignition
 
       /// \brief Set whether this event has been signaled.
       /// \param[in] _sig True if the event has been signaled.
-      public: void SetSignaled(const bool _sig);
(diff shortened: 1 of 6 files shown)


mjcarroll:
Test failures are flakes:

- macOS action: `PERFORMANCE_plugin_specialization`
- `codecov` dropped by 0.01%

iche033:
I can't reproduce the `PERFORMANCE_plugin_specialization` test failure and I think it is unrelated to the changes in this PR

iche033 (approved):

PR #299, opened 2022-01-25
Utilize ImplPtr for Timer

mjcarroll:
Updates the Timer class to use the ImplPtr pattern.

I also noticed that there was no coverage, so I added a few unittests.


ahcorde (approved):

src/Timer_TEST.cc
```diff
@@ -0,0 +1,83 @@
...
+/////////////////////////////////////////////////
+TEST(Timer_TEST, Elapsed)
+{
+  ignition::common::Timer t;
+  t.Start();
+  std::this_thread::sleep_for(std::chrono::milliseconds(50));
+  t.Stop();
+
+  // Loose margin just to make sure it is close without being flaky
+  EXPECT_NEAR(0.05, t.ElapsedTime().count(), 0.01);
```
ahcorde, line 54: https://github.com/ignitionrobotics/ign-common/runs/4942752187?check_suite_focus=true

ahcorde (changes requested):

src/Timer_TEST.cc
mjcarroll: Well, I thought that would be wide enough for any flakiness.  Maybe need to eliminate any time-based if the macOS actions are that bad.

mjcarroll: Changed to just make sure elapsed time is more than 0.

PR #300, opened 2022-01-25
Add SKIP_DEPRECATED functionality to ign-common

mjcarroll:
Implementation of ignitionrobotics/ign-cmake#204


CMakeLists.txt
```diff
@@ -40,6 +40,15 @@ elseif(WIN32)
   option(USE_EXTERNAL_TINYXML2 "Use a system-installed version of tinyxml2" OFF)
 endif()
 
+#--------------------------------------
+# Option: Should we skip tests of deprecated functionality 
+option(SKIP_DEPRECATED_TESTS "Skip testing deprecated funcationality" OFF)
```
j-rivero, line 45: Reading the description, I would have a question: is this skipping the build of the deprecated tests or is this skipping the run of the deprecated tests?

j-rivero:
Looks good to me the feature. Do you prefer going on here or move it to `ign-cmake` and close this one?

CMakeLists.txt
mjcarroll: > Reading the description, I would have a question: is this skipping the build of the deprecated tests or is this skipping the run of the deprecated tests?

The idea is to skip both.  This would primarily be for local development as building/running those tests isn't particularly valuable when you are iterating quickly.

mjcarroll:
>  Do you prefer going on here or move it to ign-cmake and close this one?

I'm going to make an ign-cmake version, I think it makes more sense there.

PR #308, opened 2022-02-09
Fix filesystem::separator behavior

mjcarroll:
As it turns out, separator's behavior was not matching the `ign-common5` behavior.  Passing an empty string should return the preferred separator for the operating system.  We were not currently testing for this.

### DIFF
--- a/src/Filesystem.cc
+++ b/src/Filesystem.cc
@@ -99,8 +99,7 @@ bool ignition::common::createDirectories(const std::string &_path)
 /////////////////////////////////////////////////
 std::string const ignition::common::separator(std::string const &_s)
 {
-  fs::path path(_s);
-  return (_s / fs::path("")).string();
+  return _s + std::string{fs::path::preferred_separator};
 }
 
 /////////////////////////////////////////////////
@@ -143,6 +142,11 @@ std::string ignition::common::joinPaths(
 {
   fs::path p1{_path1};
   fs::path p2{_path2};
+
+  if (p1.empty())
+  {
+    p1 = ignition::common::separator("");
+  }
 
   bool is_url = false;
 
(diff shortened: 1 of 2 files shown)


src/Filesystem.cc
```diff
@@ -195,11 +199,12 @@ bool ignition::common::chdir(const std::string &_dir)
 /////////////////////////////////////////////////
 std::string ignition::common::basename(const std::string &_path)
 {
-  fs::path p(_path);
-  // Maintain compatibility with ign-common
-  if (*_path.rbegin() == fs::path::preferred_separator)
-    p = fs::path(_path.substr(0, _path.size()-1));
-  return p.filename().string();
+  fs::path p = fs::path(_path);
```
iche033, line 202: how about just `fs::path p(_path);`?

src/Filesystem.cc
```diff
@@ -195,11 +199,12 @@ bool ignition::common::chdir(const std::string &_dir)
...
+  p /= "FOO.TXT";
+  p = p.lexically_normal();
+  p = p.parent_path();
+  return p.filename().string().empty() ? 
+    std::string{fs::path::preferred_separator} : p.filename().string();
```
iche033, line 207: I see that the `parentPath` function below is using the same logic that's being replaced here. Maybe that needs to be updated too?

mjcarroll: Well, as it turns out, I wasn't matching behavior correctly.  I opened a PR against `ign-common4` with some more tests to make sure that basename/parentPath were matching between the two versions: #309  and then cherry-picked it here.

chapulina (approved):
I addressed review feedback in d453ab664033441b004c2a0ac66523c2e52cdda8. This looks good to me with happy CI. I'd like to get it in because it's currently not possible to download Fuel models that have directories without this PR.

PR #309, opened 2022-02-10
Add some tests for parentPath/basename

mjcarroll:
Verify the behavior of parentPath and basename in ign-common4 to assert that #308 is correct.

### DIFF
--- a/src/Filesystem_TEST.cc
+++ b/src/Filesystem_TEST.cc
@@ -513,22 +513,38 @@ TEST_F(FilesystemTest, decomposition)
 
   std::string nobase = joinPaths("baz", "");
   EXPECT_EQ(basename(nobase), "baz");
+#ifndef _WIN32
   EXPECT_EQ(parentPath(nobase), "baz/");
+#else
+  EXPECT_EQ(parentPath(nobase), "baz\\");
+#endif
 
   std::string multiple_slash = separator("baz") + separator("") + separator("")
     + separator("");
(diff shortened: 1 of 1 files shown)


src/Filesystem_TEST.cc
```diff
@@ -493,38 +493,47 @@ TEST_F(FilesystemTest, cwd_error)
 }
 
 /////////////////////////////////////////////////
-TEST_F(FilesystemTest, basename)
+TEST_F(FilesystemTest, decomposition)
 {
   std::string absolute = joinPaths("", "home", "bob", "foo");
   EXPECT_EQ(basename(absolute), "foo");
+  EXPECT_EQ(parentPath(absolute), "/home/bob");
```
iche033, line 500: test failed here on windows:
https://build.osrfoundation.org/job/ign_common-pr-win/197/testReport/junit/(root)/FilesystemTest/decomposition/

chapulina (approved):
LGTM with happy CI.

Note that the absolute `parentPath` test is having different results for Windows on this branch and on `main` with #308 

* Here: `EXPECT_EQ(parentPath(absolute), "home\\bob")`
* On `main`: `EXPECT_EQ(parentPath(absolute), "\\home\\bob")`

If we consider the behaviour on this branch to be correct (since it's a stable version), then a fix should go into `main` later. Otherwise, if we think this branch is wrong, we could change the behaviour here and document it in the migration guide (but that may affect users).

PR #314, opened 2022-02-23
Add common::testing module

mjcarroll:
# 🎉 New feature

## Summary

This introduces `ignition::common::testing`, a component to hold all of our testing support functionality. 

The main motivation is to do away with `test_config.h.in` files across our components, and provide a more reliable way of getting source files, test support files, and temporary directories from the file system.

This should end up with any uses of `PROJECT_SOURCE_PATH` or `CMAKE_SOURCE_PATH` removed from our codebase in a centralized module.

I also envision this holding other common functionality, such as forking executables as part of a test.

## Test it

See #315 for how this is used

Requires ignitionrobotics/ign-cmake/pull/206


scpeters:
should this go in ign-utils so it can be used by ign-math and ign-plugin?

mjcarroll:
> should this go in ign-utils so it can be used by ign-math and ign-plugin?

I think ign-utils is pretty safe because it doesn't have a lot of need for it.  

In theory we could make the dependency from ign-plugin to ign-common _purely_ for the tests, but this really only works if you are building an entire source workspace with colcon or bazel rather than depending on packaged binaries.

mjcarroll:
ign-plugin also circumvents this a bit by directly injecting paths to libraries in the tests via preprocessor definitions.  It works fine for a few examples and is probably okay because plugin assumes that most of the path munging will have already been handed by common.

include/ignition/common/TempDirectory.hh
```diff
@@ -74,6 +74,22 @@ namespace ignition
...
+      /// up when the object goes out of scope.
+      ///
+      /// The TempDirectory will have the form _root/_subdir/_prefixXXXXX/
+      ///
+      /// \param[in[ _root Explicitly set the root directory
+      /// \param[in] _prefix String to be expanded for the template
+      /// \param[in] _subDir Subdirectory in OS _root, if desired
+      /// \param[in] _cleanup True to indicate that the filesystem should
+      ///   be cleaned as part of the destructor
+      public: explicit TempDirectory(const std::string &_root,
```
chapulina, line 88: Can we add a unit test for this constructor?

testing/include/ignition/common/testing/AutoLogFixture.hh
```diff
@@ -0,0 +1,109 @@
...
+namespace ignition::common::testing
+{
+/// \brief A utility class that stores test logs in ~/.ignition/test_logs.
+/// This functionality is needed to keep all the log information reported
+/// by ignition during continuous integration. Without this, debugging
+/// failing tests is significantly more difficult.
+class AutoLogFixture : public ::testing::Test
+{
+  /// \brief Setup the test fixture. This gets called by gtest.
+  protected: virtual void SetUp()
```
chapulina, line 39:
```suggestion
  protected: virtual void SetUp() override
```

testing/include/ignition/common/testing/AutoLogFixture.hh
```diff
@@ -0,0 +1,109 @@
...
+
+  /// \brief Default destructor.
+  public: virtual ~AutoLogFixture()
+  {
+    ignLogClose();
+    EXPECT_TRUE(ignition::common::unsetenv(IGN_HOMEDIR));
+  }
+
+  /// \brief String with the full path of the logfile
+  private: std::string logFilename;
```
chapulina, line 96: How about PIMPLing this class?

testing/include/ignition/common/testing/BazelTestPaths.hh
```diff
@@ -0,0 +1,37 @@
...
+#define IGNITION_COMMON_TESTING_BAZELTESTPATHS_HH_
+
+#include <string>
+
+#include "ignition/common/testing/TestPaths.hh"
+
+namespace ignition::common::testing
+{
+
+class BazelTestPaths: public TestPaths
```
chapulina, line 27: Missing doxygen

testing/include/ignition/common/testing/CMakeTestPaths.hh
```diff
@@ -0,0 +1,37 @@
...
+#define IGNITION_COMMON_TESTING_CMAKETESTPATHS_HH_
+
+#include <string>
+
+#include "ignition/common/testing/TestPaths.hh"
+
+namespace ignition::common::testing
+{
+
+class CMakeTestPaths: public TestPaths
```
chapulina, line 27: Doxygen

testing/include/ignition/common/testing/TestPaths.hh
```diff
@@ -0,0 +1,118 @@
...
+//////////////////////////////////////////////////
+enum class BuildType
+{
+  kUnknown,
+  kCMake,
+  kBazel
+};
+
+//////////////////////////////////////////////////
+class TestPaths
```
chapulina, line 46: doxygen

testing/include/ignition/common/testing/TestPaths.hh
```diff
@@ -0,0 +1,118 @@
...
+  std::string dataDir;
+  testPaths->ProjectSourcePath(dataDir);
+  return common::joinPaths(dataDir, args...);
+}
+
+//////////////////////////////////////////////////
+template <typename... Args>
+std::string TestFile(Args const &... args)
+{
+  return SourceFile("test", args...);
```
chapulina, line 103:
```suggestion
  return this->SourceFile("test", args...);
```

testing/include/ignition/common/testing/Utils.hh
```diff
@@ -0,0 +1,59 @@
...
+  std::mt19937 randGenerator(seed);
+
+  // Create a random number based on an integer converted to string.
+  std::uniform_int_distribution<int32_t> d(_min, _max);
+
+  return std::to_string(d(randGenerator));
+}
+
+/////////////////////////////////////////////////
+bool create_new_empty_file(const std::string &_filename)
```
chapulina, line 44:
```suggestion
bool createNewEmptyFile(const std::string &_filename)
```

Also missing docs

testing/src/BazelTestPaths_TEST.cc
```diff
@@ -0,0 +1,90 @@
...
+#include "ignition/common/Filesystem.hh"
+#include "ignition/common/testing/TestPaths.hh"
+#include "ignition/common/testing/BazelTestPaths.hh"
+
+using namespace ignition::common;
+
+char **kEnv;
+
+/////////////////////////////////////////////////
+TEST(TestPaths, DISABLED_DumpEnv)
```
chapulina, line 28: Not sure if disabling is a good idea here, this gets printed out when you run tests and looks like something to be fixed. I think if it were enabled it would just pass, right?

testing/src/CMakeTestPaths.cc
```diff
@@ -0,0 +1,44 @@
...
+#include "ignition/common/testing/CMakeTestPaths.hh"
+
+namespace ignition::common::testing
+{
+
+//////////////////////////////////////////////////
+CMakeTestPaths::~CMakeTestPaths() = default;
+
+//////////////////////////////////////////////////
+bool CMakeTestPaths::ProjectSourcePath(std::string &_sourceDir) {
```
chapulina, line 27:
```suggestion
bool CMakeTestPaths::ProjectSourcePath(std::string &_sourceDir)
{
```

testing/src/CMakeTestPaths.cc
```diff
@@ -0,0 +1,44 @@
...
+
+  if (!projectSourcePath.empty())
```
chapulina, line 29:
```suggestion
  if (!this->projectSourcePath.empty())
```

chapulina (commented):
Did a 1st pass

testing/src/BazelTestPaths_TEST.cc
mjcarroll: Removed

testing/include/ignition/common/testing/TestPaths.hh
mjcarroll: This one isn't a member function, it is just calling another template.

testing/include/ignition/common/testing/AutoLogFixture.hh
mjcarroll: I PIMPL'ed it, but to keep from linking a shared library with `gtest`, I put the implementation in the `detail` directory.  This keeps the AutoLogFixture header-only.

mjcarroll:

testing/include/ignition/common/testing/Utils.hh
```diff
@@ -0,0 +1,41 @@
...
+#include <cstdint>
+#include <string>
+
+namespace ignition::common::testing
+{
+
+/////////////////////////////////////////////////
+/// \brief Get a random number based on an integer converted to string.
+/// \return A random integer converted to string.
+std::string getRandomNumber(int32_t _min = 0, int32_t _max = INT_MAX);
```
nkoenig, line 30: `ign-math` has helper functions to return random numbers. I looks like the `getRandomNumber` is only used in the tests.

You could replace those instances with:
```
auto val1 = std::to_string(math::Rand::IntUniform(min,max));
```

mjcarroll: Good call, this got lifted straight out of a `test_config.h` somewhere, so probably not needed.

testing/src/TestPaths.cc
```diff
@@ -0,0 +1,96 @@
...
+TestPaths::~TestPaths() = default;
+
+//////////////////////////////////////////////////
+BuildType TestBuildType(const std::string &_projectSourcePath)
+{
+  std::string ign_bazel;
+  bool ign_bazel_set = common::env("IGN_BAZEL", ign_bazel);
+  bool ign_cmake_set = !_projectSourcePath.empty();
+
+  if (ign_bazel_set)
```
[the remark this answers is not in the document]
mjcarroll: I generally assumed that IGN_BAZEL should only be set within the context of the bazel build/test system.  It isn't an environment variable that anyone should be setting.

I don't think it should be an issue, but I will invert the logic just to be safe.

testing/include/ignition/common/testing/detail/AutoLogFixture.hh
```diff
@@ -0,0 +1,114 @@
...
+
+  this->dataPtr->temp = std::make_unique<TempDirectory>(
+      "test", "ign_common", false);
+  ASSERT_TRUE(this->dataPtr->temp->Valid());
+  common::setenv(IGN_HOMEDIR, this->dataPtr->temp->Path());
+
+  // Initialize Console
+  ignLogInit(common::joinPaths(this->dataPtr->temp->Path(), "test_logs"),
+      this->dataPtr->logFilename);
+
```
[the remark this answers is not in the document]
mjcarroll: I think ASSERT_TRUE is an issue (because it changes the return type), but an EXPECT_TRUE probably makes sense here.

mjcarroll:
#337 should fix the Windows CI

testing/src/TestPaths.cc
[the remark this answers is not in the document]
mjcarroll: [046cf5b](https://github.com/ignitionrobotics/ign-common/pull/314/commits/046cf5b2534d827aaacf887e3cc798fb9e6c0ba0)

testing/include/ignition/common/testing/detail/AutoLogFixture.hh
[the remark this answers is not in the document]
mjcarroll: [9c3b6dc](https://github.com/ignitionrobotics/ign-common/pull/314/commits/9c3b6dc7580e7d4638cb65f4a4dbb835e768bbf1)

testing/include/ignition/common/testing/Utils.hh
mjcarroll: [44cce83](https://github.com/ignitionrobotics/ign-common/pull/314/commits/44cce83d96a8a943132d4127258480d8b9e794fd)

testing/include/ignition/common/testing/TestPaths.hh
```diff
@@ -0,0 +1,191 @@
...
+//////////////////////////////////////////////////
+/// \brief Helper interface to generate path information to support
+/// test access to source/data files
+///
+/// It is intended that there is an implementation of this interface for
+/// each relevant buildsystem.
+class TestPaths
+{
+  /// \brief Constructor
+  public: explicit TestPaths(const std::string &_projectSourcePath =
```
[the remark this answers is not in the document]
mjcarroll: [bab3599](https://github.com/ignitionrobotics/ign-common/pull/314/commits/bab3599167ba1bb42ccab8ce97c53a7c4f3bfa02)

src/TempDirectory.cc
```diff
@@ -161,13 +161,22 @@ class ignition::common::TempDirectory::Implementation
...
+                             const std::string &_subDir,
+                             bool _cleanup):
   dataPtr(ignition::utils::MakeUniqueImpl<Implementation>())
 {
 
   this->dataPtr->oldPath = common::cwd();
   this->dataPtr->doCleanup = _cleanup;
 
-  auto tempPath = common::tempDirectoryPath();
+  auto tempPath = _root;
```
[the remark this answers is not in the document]
mjcarroll: [b2f7480](https://github.com/ignitionrobotics/ign-common/pull/314/commits/b2f7480bf1c8840d81bc12d411ce1f4fc2de2283)

nkoenig (approved):

mjcarroll:
> we need to update the ign-common5-release repo with this new component, similar to https://github.com/ignition-release/ign-common5-release/pull/4

Will do, thanks

chapulina:
* PR adding release metadata: https://github.com/ignition-release/ign-common5-release/pull/10

PR #315, opened 2022-02-23
Test updates for common::testing module.

mjcarroll:
Update ignition-common to make use of ign-common::testing

* Builds on top of #314


src/Timer_TEST.cc
```diff
@@ -41,6 +41,7 @@ TEST(Timer_TEST, Sequence)
 
   t.Stop();
   EXPECT_FALSE(t.Running());
+  std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
nkoenig, line 44: Is this addition needed?

nkoenig (approved):
One minor comment

src/Timer_TEST.cc
mjcarroll: No, I believe the flakiness was addressed via https://github.com/ignitionrobotics/ign-common/pull/343

Removed.

PR #319, opened 2022-03-03
Fix edge case handling for SubMesh::MaterialIndex

### LINKED ISSUE #317: Inconsistent signed/unsigned integer type in SubMesh MaterialIndex
As noted in https://github.com/ignitionrobotics/sdformat/pull/831/files#r816364549, the [SubMesh::MaterialIndex()](https://github.com/ignitionrobotics/ign-common/blob/ignition-common4_4.5.0/graphics/include/ignition/common/SubMesh.hh#L287-L289) accessor method and corresponding setter method both use `unsigned int` for the material index, but the value is stored internally as an `int` in [Material.cc](https://github.com/ignitionrobotics/ign-common/blob/ignition-common4_4.5.0/graphics/src/SubMesh.cc#L55) and explicitly uses a value of `-1` to represent an unsigned value.

We should consider using `std::optional<unsigned int>` for this API in ign-common5.

adlarkin:
Signed-off-by: Ashton Larkin <42042756+adlarkin@users.noreply.github.com>

# 🦟 Bug fix

Fixes #317

## Summary
As mentioned in #317, `SubMesh::MaterialIndex` returns an `unsigned int`, but internally, this data is stored in an `int` with an initial value of `-1` to represent a submesh having no material index applied to it.

To avoid implicit casts, I've created a new method named `GetMaterialIndex` that replaces the current `MaterialIndex` method. This new method returns a `std::optional<unsigned int>` to indicate whether the submesh has a material index assigned to it or not. The current `MaterialIndex` method is now deprecated as a result.

Also, for clarity/completeness, I updated `MaterialIndex` to explicitly return `std::numeric_limits<unsigned int>::max()` instead of implictly casting `-1` to an `unsigned int` if a submesh has no material index.

## Checklist
- [x] Added tests


graphics/include/ignition/common/SubMesh.hh
```diff
@@ -284,8 +285,15 @@ namespace ignition
...
-      public: unsigned int MaterialIndex() const;
+      /// \return The assigned material index. If no material index is assigned
+      /// to this submesh, std::numeric_limits<unsigned int>::max() is returned.
+      /// \note This method is deprecated, use GetMaterialIndex instead
+      public: unsigned int IGN_DEPRECATED(5) MaterialIndex() const;
+
+      /// \brief Get the material index
+      /// \return The assigned material index. Nullopt is returned if the
+      /// submesh has no assigned material index
+      public: std::optional<unsigned int> GetMaterialIndex() const;
```
ahcorde, line 296: We are not using the prefix `Get` to define the getters

ahcorde (changes requested):

graphics/include/ignition/common/SubMesh.hh
scpeters: I believe we need a new name for the method in order to change the return value, so GetMaterialIndex seems a reasonable choice, though it does break our naming convention.

adlarkin: > I believe we need a new name for the method in order to change the return value

Yes, that's right - see https://github.com/ignitionrobotics/ign-common/issues/317#issuecomment-1055965199. This code won't compile if `MaterialIndex` is used since it would result in two `MaterialIndex` methods with different return values, which is ambiguous. If we don't want to use the `Get` prefix here, what should we use instead?

adlarkin: As a follow-up to this, a workaround to avoid using the `Get` prefix is to make this new method take a mutable input parameter that the method can write to. This is what's being done for `gazebo::EntityComponentManager::State` to differentiate between the `msgs::SerializedState` and `msgs::SerializedStateMap` types:
* https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo6_6.7.0/include/ignition/gazebo/EntityComponentManager.hh#L554-L564
* https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo6_6.7.0/include/ignition/gazebo/EntityComponentManager.hh#L609-L624

So, for this PR, the new method would look like this:
```c++
MaterialIndex(std::optional<unsigned int> &_idx) const;
```

What do others think about this approach?

scpeters: that also works; either way is fine in my opinion

it's slightly easier to use with a `return` value instead of a mutable input

~~~
audo index = submesh->GetMaterialIndex();
// vs
std::optional<unsigned int> index;
submesh->MaterialIndex(index);
~~~

but our pattern is not to use `Get` in method names

adlarkin: > it's slightly easier to use with a return value instead of a mutable input

I agree; okay, I will leave this as-is unless anyone else has another idea. I don't think there's a way to get around the `Get` prefix for this scenario if we decide to take the return value approach (unless someone else has an idea for a different method name).

adlarkin:
@ahcorde we are going to stick with the `Get` prefix based on https://github.com/ignitionrobotics/ign-common/pull/319#discussion_r827431739, would you mind giving this another look?

ahcorde (approved):

chapulina:
Someone needs to open a PR against `ign-rendering` using the new API and fix deprecation warnings. CC @adlarkin

adlarkin:
> Someone needs to open a PR against `ign-rendering` and `ign-gazebo` using the new API and fix deprecation warnings.

I'll take care of this!

adlarkin:
> Someone needs to open a PR against `ign-rendering` and `ign-gazebo` using the new API and fix deprecation warnings.

Here are PRs that should take care of deprecation warnings:
* https://github.com/ignitionrobotics/ign-gazebo/pull/1436
* https://github.com/ignitionrobotics/ign-rendering/pull/606

PR #321, opened 2022-03-15
DEM improvements to nodata, buffer and other corner cases

chapulina:
# 🦟 Bug fix

## Summary

Various small fixes to work with some DEMs I have:

#### Buffer to min height

Set buffer height to minimum height value. Non-square DEMs are automatically padded to become a square. The original logic is using `height == 0` for the padding, which is very awkward for DEMs below ground. This PR changes the `bufferVal` to match the minimum height on the heightmap because a cliff down should be better than a cliff up, which is visible from far away.

Example:

Before (cliff up) | After (cliff down)
--- | ---
portuguese_ledge | pt_ledge

#### nodata comparisons

* All comparisons to NaN return false, which caused the `if (d > noDataValue)` check to prevent calculating the min / max values. I fixed that by guarding explicitly for NaNs.
* Another assumption in the current code is that the nodata value is smaller than all heights on the DEM. That's not the case for some DEMs I have here, which have very high values for nodata. So instead of checking `if (d > noDataValue)` I'm doing `math::equal(d, noDataValue)`.

#### Dem::LoadData indentation

The entire function was too indented, so I fixed that as part of this PR. I recommend hiding whitespace changes to see what's actually changed.

#### GetProjectionRef and GetGeoTransform

These calls were returning values for my DEMs which weren't being well-handled. I added a guard and reduced an error to a debug message because the DEMs still work without these.

#### NetCDF

I also went ahead and used a NetCDF file (`.nc`) in a test to make sure we keep supporting this format.

## Checklist
- [x] Added tests


geospatial/src/Dem.cc
```diff
@@ -184,21 +187,31 @@ int Dem::Load(const std::string &_filename)
...
-  double noDataValue = this->dataPtr->band->GetNoDataValue(&validNoData);
+  float noDataValue = this->dataPtr->band->GetNoDataValue(&validNoData);
+
   if (validNoData <= 0)
     noDataValue = defaultNoDataValue;
 
   double min = ignition::math::MAX_D;
   double max = -ignition::math::MAX_D;
-  for (auto d : this->dataPtr->demData)
+  for (auto &d : this->dataPtr->demData)
```
mjcarroll, line 197:
```suggestion
  for (const auto &d : this->dataPtr->demData)
```

mjcarroll (approved):
One place for `const`, LGTM.

PR #348, opened 2022-04-28
ign -> gz Header Migration : ign-common

methylDragon:
See https://github.com/ignition-tooling/release-tools/issues/711


methylDragon:
This is pending the merge and nightly of `ign-math`

geospatial/include/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_subdirectory(gz)
+install(DIRECTORY ignition DESTINATION ${IGN_INCLUDE_INSTALL_DIR_FULL})
```
chapulina, line 2: `geospatial` is a new component from Garden, so we don't need to add `ignition` headers and can go straight to `gz`, because no downstream users should be using it yet anyway.

One caveat is that our libraries are already using it, so one strategy we could take would be to keep `ignition/geospatial` while we're updating all libraries, and then remove it afterwards.

include/gz/common/Plugin.hh
```diff
@@ -0,0 +1,196 @@
+/*
```
chapulina, line 1: We're deprecating all the `*Plugin*` files from `ign-common` on Garden, in favor of using the `ign-plugin` library. So I think we don't need to add `gz` counterparts to it.

I'd be ok doing this in a separate PR, we don't even have `IGN_DEPRECATED` notices on them anyway.

include/ignition/common/detail/Plugin.hh
```diff
@@ -15,98 +15,4 @@
  *
  */
```
chapulina, line 17: Since the `detail` directories aren't part of the public API, I think we can just remove them. 

If `include/ignition/common/Plugin.hh` isn't including `include/ignition/common/detail/Plugin.hh`, no one else is.

This is valid for all libraries.

profiler/CMakeLists.txt
```diff
@@ -1,7 +1,10 @@
-
 add_subdirectory(src)
+# add_subdirectory(include)
```
chapulina, line 2: Should we do this instead of installing it below?

```suggestion
add_subdirectory(gz)
```

testing/include/ignition/common/testing/AutoLogFixture.hh
```diff
@@ -1,59 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#include <gz/common/testing/AutoLogFixture.hh>
```
chapulina, line 18: The `testing` component is also new on Garden, just like `geospatial`. So we don't need to provide `ignition/testing` headers.

chapulina (approved):
LGTM with green CI. My comments below can be addressed in follow-up PRs if that's easier.

profiler/CMakeLists.txt
methylDragon: I tried doing it that way at first, but I was getting an error:
```
Developer error: You are missing the file
  [.../ign_ws/src/ign-common/profiler/include/gz/common/config.hh.in]!
```

I suspect it's from the placement of the CMakeLists file or the fact that there's no further components inside. I didn't want to move the folder hierarchy around since there was a baze file as well :X

include/ignition/common/detail/Plugin.hh
methylDragon: Can I double check if this means that I delete all detail redirection headers across all libraries?

include/gz/common/Plugin.hh
methylDragon: Let's do this in a separate PR, so it can be tracked independently.

I'll track this in the issue here: https://github.com/ignitionrobotics/ign-common/issues/349
But I suspect I'll be focused on the other libraries.

geospatial/include/CMakeLists.txt
methylDragon: Might be neater to remove it during the namespace (or maybe the source) migration? I'm not sure :o

av/include/ignition/common/HWEncoder.hh
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Martin Pecka
+ * Copyright (C) 2022 Open Source Robotics Foundation
```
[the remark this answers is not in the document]
methylDragon: https://github.com/ignitionrobotics/ign-common/pull/348/commits/df2ddf011ef7e669c81d6921b629a8a7723f919a

av/include/ignition/common/HWVideo.hh
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Martin Pecka
+ * Copyright (C) 2022 Open Source Robotics Foundation
```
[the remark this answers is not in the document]
methylDragon: https://github.com/ignitionrobotics/ign-common/pull/348/commits/df2ddf011ef7e669c81d6921b629a8a7723f919a

include/ignition/common/detail/Plugin.hh
scpeters: > Since the `detail` directories aren't part of the public API, I think we can just remove them.
> 
> If `include/ignition/common/Plugin.hh` isn't including `include/ignition/common/detail/Plugin.hh`, no one else is.
> 
> This is valid for all libraries.

I disagree. These are installed header files, so they are part of the public API, and they are included from the header files not in the `detail` folder, though often from the bottom of the file so it is easy to miss:

* https://github.com/ignitionrobotics/ign-math/blob/ignition-math6_6.10.0/include/ignition/math/Box.hh#L224
* https://github.com/ignitionrobotics/ign-common/blob/ignition-common4_4.5.0/include/ignition/common/Plugin.hh#L194

I think it would be best to keep the redirection headers for all the contents of the `detail`  subfolders

chapulina: > These are installed header files, so they are part of the public API,

I understand that we've never explicitly declared our public API, and we should. But the reason that the `detail` directories exist is that they're not meant to be directly used by downstream. They're installed because of technicalities of the language which require those templates to be installed, but that's an implementation _detail_, not our API. And as such, we don't need to tick-tock them.

ROS QL1 packages [explicitly call this out](https://github.com/ros2/rclcpp/blob/master/rclcpp/QUALITY_DECLARATION.md#public-api-declaration-1iii). I ticketed this issue so we document this properly for Gazebo: 

* https://github.com/ignitionrobotics/docs/issues/248

> I think it would be best to keep the redirection headers for all the contents of the detail subfolders

Having them there wouldn't break anything, but those should be dead files that no one is using. I'd say that keeping them there sends the wrong message that those are part of our public API and subject to tick-tock.

geospatial/include/CMakeLists.txt
chapulina: I think any time after the header migration for all libraries should work, because none of the libraries should be using them anymore. So I'm fine either way. It's probably good to do it as a standalone PR for visibility. We just shouldn't forget to do it.

include/ignition/common/detail/Plugin.hh
scpeters: I'm getting a very different impression from your first and most recent comments. I think your points are reasonable, but the way they were initially phrased was a bit alarming to me.

> Since the `detail` directories aren't part of the public API, I think we can just remove them.

I've gathered from your subsequent comment that the context is specifically about redirection headers in a `detail`  subfolder, but on first glance it sounded to me like you were saying we can delete `detail` headers at any time from anywhere with no impact on API, which is absolutely not true. I now think you meant we don't need to tick-tock them, which is reasonable.

> If `include/ignition/common/Plugin.hh` isn't including `include/ignition/common/detail/Plugin.hh`, no one else is.

Again, I now gather that this context is about redirection headers, not public headers in general, since `gz/common/Plugin.hh` does include `gz/common/detail/Plugin.hh`

> This is valid for all libraries.

Nothing specific here but it added to my alarm! 😅

chapulina: Ah sorry I should have been clearer. Yes, what I meant is that we can remove the `ignition/*/detail/*.hh` headers that just include `gz/*/detail/*.hh` ✅

scpeters: I think downstream users are less likely to need the `ignition/*/detail/*` redirection headers. I would lean towards redirecting all headers for now until we've migrated headers for all the ignition libraries and then consider removing the `detail` redirects when we add the deprecation warnings to the primary redirection headers.

chapulina: > I think downstream users are less likely to need the ignition/*/detail/* redirection headers

If they're including them directly, that's a mistake on their part.

> I would lean towards redirecting all headers for now until we've migrated headers

I'm ok with that if it makes things easier. We already have a long list of things to go back to, and long PRs to review. If the script could be updated to skip the `detail` folders automatically I think that would make our lives easier...

methylDragon: Done! 7b17e475504d5ad308b943e3e5f848577b65d2de

PR #355, opened 2022-05-17
Assimp Loader with Recursion

onurtore:
Hi,
I believe you can use some part of this code to further improve this branch. The existing structure lacks recursive loading, but this snippet  can be utilized to implement that future.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+    "files.associations": {
+        "string": "cpp",
+        "string_view": "cpp"
+    }
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,6 +122,12 @@ ign_find_package(AVCODEC REQUIRED_BY av PRETTY libavcodec)
 ign_find_package(AVUTIL REQUIRED_BY av PRETTY libavutil)
 
 
+#------------------------------------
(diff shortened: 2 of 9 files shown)


onurtore:
@luca-della-vedova

luca-della-vedova:
Hi @onurtore ! This is super helpful, indeed the meshes should be built in a recursive way and not iterative as I was doing in a first quick draft, the limitations became evident when parsing scenes with complex hierarchies where each transform is derived from the parent transform (i.e. bones for animations).
I think this is a very large and complex task and help is definitely welcome, there are a lot of TODOs scattered around (and to be honest I put "sandbox" in  the name because the code is very :fire: right now, this is still in a feasibility study state for me, where I want to see if we can get all the basic features out of assimp before refactoring and polishing the code).
I was hoping assimp could be a full replacement for all the loaders and was hoping to prove that "it works" by merging all the unit tests in our loaders into a single `AssimpLoader_TEST.cc` file and making sure they work, I started this and documented the first few items not working so they can be a good target for fixes / documentation where needed (i.e. shininess, number of materials on top of my head).
I was also hoping that we could either remove hardcoded supported mesh formats or greatly increase their number since assimp supports a whole variety of them.

onurtore:
@luca-della-vedova,
Perfect

PR #350, opened 2022-05-03
Deprecate plugin in favor of gz-plugin

chapulina:
* Closes #349 

## Summary

* 0e4a0f5de63f47daf8170404a5a3fdc7802353ec: Move plugin headers back to `ignition` folder because users shouldn't migrate to `gz/common/*Plugin*`, but migrate to `gz/plugin` instead.
* 7090764daa6afefd4812fe39aafaf32e317e57be: Use the correct headers
* e0f0e216a48082d310ef30ef86df793813fbb67d: Add deprecation warnings

## Test it

CI should pass without warnings.

My local cppcheck had some `unknownMacro` warnings, but that may be due to my version. I'll see what CI says and suppress those warnings as needed.

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [x] ~~Added example and/or tutorial~~
- [x] Updated documentation (as needed)
- [x] Updated migration guide (as needed)
- [x] ~~Consider updating Python bindings (if the library has them)~~
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸

PR #356, opened 2022-05-18
ign -> gz Namespace Migration : gz-common

methylDragon:
See https://github.com/gazebo-tooling/release-tools/issues/711


include/gz/common/Console.hh
```diff
@@ -27,24 +27,30 @@
 #include <gz/common/Export.hh>
 #include <gz/utils/SuppressWarning.hh>
 
-namespace ignition
+namespace gz
 {
   namespace common
   {
+    // TODO(CH3): Remove all the ign macros on tock
```
chapulina, line 34: I like to leave a comment that has the word "deprecated" on it so it's easier to find when performing the tock. Otherwise this may get lost in the pool of TODOs

Migration.md
```diff
@@ -5,13 +5,16 @@ Deprecated code produces compile-time warnings. These warning serve as
...
+## Gazebo Common 4.X to 5.X
 
 ### Deprecations
 
 1. `Submesh::MaterialIndex` is deprecated. `SubMesh::GetMaterialIndex` should
    be used instead, which properly handles submeshes having no material index
    applied to them.
+2. The `ignition` namespace is deprecated and will be removed in future versions.  Use `gz` instead.
+3. Header files under `ignition/...` are deprecated and will be removed in future versions.
+   Use `gz/...` instead.
```
chapulina, line 17: We should add a note about the console deprecations (`ignerr` etc)

test/plugins/BadPluginAlign.cc
```diff
@@ -15,20 +15,20 @@
  *
 */
 
-#include "ignition/common/PluginMacros.hh"
+#include "gz/common/PluginMacros.hh"
```
chapulina, line 18: Ahh I'm undoing the plugin header migration on 

* https://github.com/gazebosim/gz-common/pull/350

We can either merge that first and update this PR, or the other way around. I think it may be easier to do the former so we have less changes to update here. I need to address reviewer feedback on that PR.

chapulina (commented):
LGTM besides the comments below.

test/plugins/BadPluginAlign.cc
methylDragon: Let's do the former! I'll be updating this one in the meantime.

Migration.md
methylDragon: Done! https://github.com/gazebosim/gz-common/pull/356/commits/dcc40dce6630288ba9ed67b08a0bd08499be3404

include/gz/common/Console.hh
methylDragon: Done! https://github.com/gazebosim/gz-common/pull/356/commits/dcc40dce6630288ba9ed67b08a0bd08499be3404 

I also did so in `gz-sim`

test/plugins/BadPluginAlign.cc
methylDragon: Rebased after the merge of the PR!

The rebase was not nice to do, and I had to do some manual additions and changes, might want to take another look at the changes:
- https://github.com/gazebosim/gz-common/pull/356/commits/d6211d1739f0736818a08777ddde6cd26fabb94b
- https://github.com/gazebosim/gz-common/pull/356/commits/c6c7825a2c5e7c0d6bb3c721e368108453c02afe

src/Plugin.cc
```diff
@@ -23,7 +25,7 @@ IGN_UTILS_WARN_IGNORE__DEPRECATED_DECLARATION
 #include "ignition/common/PluginInfo.hh"
 #include "PluginUtils.hh"
 
-namespace ignition
+namespace gz
```
chapulina, line 28: I think this broke the build, the namespace should remain `ignition` for plugin stuff. Maybe that makes it harder to do full sweeps of the code 😕

methylDragon: https://github.com/gazebosim/gz-common/pull/356/commits/0a80e8ef969400f71429c7545ed1d3a722ce61c8

I'm surprised the namespace redirection didn't help fix this. Let's migrate fully to `gz` (since we have the namespace redirections in `config.hh` anyway)

chapulina: ok, fair enough, that should have pretty much the same result to users. The only downside is that they will get warnings saying that they should use `gz` instead of `ignition`, but `gz/Plugin.hh` doesn't exist.

include/ignition/common/config.hh
```diff
@@ -15,4 +15,30 @@
...
+#define IGNITION_COMMON_VERSION GZ_COMMON_VERSION
+#define IGNITION_COMMON_VERSION_FULL GZ_COMMON_VERSION_FULL
+
+#define IGNITION_COMMON_VERSION_HEADER GZ_COMMON_VERSION_HEADER
+
+namespace gz
+{
+}
+
+namespace gz
```
chapulina, line 36:
```suggestion
namespace ignition
```

right?

chapulina (approved):
LGTM with 🟢 CI!

include/ignition/common/Plugin.hh
```diff
@@ -15,18 +15,18 @@
...
 #include <memory>
 #include <map>
 #include <string>
 
+#include <ignition/common/config.hh>
 #include <gz/common/Export.hh>
 #include <gz/utils/SuppressWarning.hh>
 
-namespace ignition
+namespace gz
```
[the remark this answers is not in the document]
chapulina: Yeah it's not ideal but there are some advantages to it, see this discussion: https://github.com/gazebosim/gz-common/pull/356#discussion_r883134423

PR #359, opened 2022-05-22
Fix: Assimp generates normals for all the meshes.

onurtore:
Some mesh files does not have normals and this creates exceptions when we try to load them into memory. Using the generate normal flag fixes this issue.


onurtore:
@luca-della-vedova

graphics/src/MeshManager.cc
```diff
@@ -146,19 +146,23 @@ const Mesh *MeshManager::Load(const std::string &_filename)
...
     else if (extension == "dae")
       loader = &this->dataPtr->colladaLoader;
     else if (extension == "obj")
-      loader = &this->dataPtr->objLoader;
+      //loader = &this->dataPtr->objLoader;
+      loader = &this->dataPtr->assimpLoader;
+    else if (extension == "gltf")
+      loader = &this->dataPtr->assimpLoader;
+    else if (extension == "fbx")
+      loader = &this->dataPtr->assimpLoader;
```
luca-della-vedova, line 160: Any specific reason for this change? My hope (if possible!) was to be able to substitute all the loaders with the assimp loader and I already got most static DAE assets, as well as some animated ones, to work fine. We can fallback later to the other loaders if we find an issue that can't be handled by assimp if that makes sense?

graphics/src/AssimpLoader.cc
```diff
@@ -307,6 +306,7 @@ Mesh *AssimpLoader::Load(const std::string &_filename)
       aiProcess_RemoveRedundantMaterials |
       aiProcess_SortByPType |
       aiProcess_Triangulate |
+      aiProcess_GenNormals |
```
luca-della-vedova, line 309: Do you have an example asset where this happens?

onurtore: Adding it below: 
[airbot.zip](https://github.com/gazebosim/gz-common/files/8751580/airbot.zip)

graphics/src/MeshManager.cc
onurtore: Yes sure, should I git force push to remove the commit, or does GitHub gives you option to select commits to merge?

luca-della-vedova: I can't really do that it's either merge all or nothing!

onurtore: Okay, gonna fix it later today.

onurtore: @luca-della-vedova  fixed.

onurtore:
@chapulina,
Can I add labels? Its looks little bit exhausting for you to add label for each PR.

PR #365, opened 2022-06-04
ign -> gz Environment Variable Migration : gz-common

methylDragon:
See: https://github.com/gazebo-tooling/release-tools/issues/734

### DIFF
--- a/src/SystemPaths.cc
+++ b/src/SystemPaths.cc
@@ -122,7 +122,7 @@ SystemPaths::SystemPaths()
   // environment variable.
 
   if (this->dataPtr->filePathEnv.empty() &&
-      !this->dataPtr->filePathEnv.empty())
+      !this->dataPtr->filePathEnvDeprecated.empty())
   {
     gzwarn << "Setting file path using deprecated environment variable ["
            <<  this->dataPtr->filePathEnvDeprecated
@@ -146,6 +146,26 @@ std::string SystemPaths::LogPath() const
 void SystemPaths::SetPluginPathEnv(const std::string &_env)
 {
   this->dataPtr->pluginPathEnv = _env;
+
+  // TODO(CH3): Deprecated. Remove on tock.
+  std::string result;
+  if (!this->dataPtr->pluginPathEnv.empty())
+  {
+    if (env(this->dataPtr->pluginPathEnv, result))
+    {
+      // TODO(CH3): Deprecated. Remove on tock.
+      std::string ign_prefix = "IGN_";
+
+      // Emit warning if env starts with IGN_
+      if (env_.compare(0, ign_prefix.length(), ign_prefix) == 0)
+      {
+        gzwarn << "Finding plugins using deprecated IGN_ prefixed environment "
+               << "variable ["
+               << _env << "]. Please use the GZ_ prefix instead"
+               << std::endl;
+      }
+    }
+  }
 }
 
 /////////////////////////////////////////////////
@@ -159,7 +179,7 @@ const std::list<std::string> &SystemPaths::PluginPaths()
       this->AddPluginPaths(result);
     }
     // TODO(CH3): Deprecated. Remove on tock.
-    else if (env(this->dataPtr->pluginPathEnvDeprecated, result))
+    if (env(this->dataPtr->pluginPathEnvDeprecated, result))
     {
       this->AddPluginPaths(result);
       gzwarn << "Finding plugins using deprecated environment variable "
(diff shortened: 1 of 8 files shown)


src/SystemPaths.cc
```diff
@@ -100,11 +117,23 @@ SystemPaths::SystemPaths()
...
   }
 
-
   this->dataPtr->logPath = fullPath;
   // Populate this->dataPtr->filePaths with values from the default
   // environment variable.
-  this->SetFilePathEnv(this->dataPtr->filePathEnv);
+
+  if (this->dataPtr->filePathEnv.empty() &&
+      !this->dataPtr->filePathEnv.empty())
```
chapulina, line 125: Humm can this ever be true? I think some variables got mixed up here

src/SystemPaths.cc
```diff
@@ -129,6 +158,15 @@ const std::list<std::string> &SystemPaths::PluginPaths()
     {
       this->AddPluginPaths(result);
     }
+    // TODO(CH3): Deprecated. Remove on tock.
+    else if (env(this->dataPtr->pluginPathEnvDeprecated, result))
```
chapulina, line 162: Since this is adding paths, I think we could support a mixed setup where users may have the old var as well as the new var

```suggestion
    if (env(this->dataPtr->pluginPathEnvDeprecated, result))
```

src/SystemPaths.cc
```diff
@@ -165,20 +203,43 @@ std::string SystemPaths::FindSharedLibrary(const std::string &_libName)
...
   {
     this->ClearFilePaths();
-    std::string result;
     if (env(this->dataPtr->filePathEnv, result))
     {
       this->AddFilePaths(result);
     }
   }
+  // TODO(CH3): Deprecated. Remove on tock.
+  else if (!this->dataPtr->filePathEnvDeprecated.empty())
```
chapulina, line 217: This doesn't look right, we've never assigned

```cpp
this->dataPtr->filePathEnvDeprecated = _env;
```

src/SystemPaths.cc
```diff
@@ -165,20 +203,43 @@ std::string SystemPaths::FindSharedLibrary(const std::string &_libName)
...
+  }
 }
 
 /////////////////////////////////////////////////
 std::string SystemPaths::FilePathEnv() const
 {
+  // TODO(CH3): Deprecated. Remove on tock.
+  if (this->dataPtr->filePathEnv.empty())
+  {
+    gzwarn << "Returning deprecated file path environment variable "
```
chapulina, line 237: This warning will come up if you have no env var set at all, which we don't want. 

I'm wondering if we can ever get to this situation though. The only way to change `filePathEnv` is through `SetFilePathEnv`. From the moment the user calls that, I think we could just empty `filePathEnvDeprecated`, because the user is using a custom variable and doesn't care about the defaults, deprecated or not.

chapulina (changes requested):

src/SystemPaths.cc
methylDragon: Oops

src/SystemPaths.cc
methylDragon: Unfortunately `SetFilePathEnv()` is called internally, so it's hard to distinguish a user-originating call.

But removing this warning makes sense!

src/SystemPaths.cc
methylDragon: Right, I'll emit a warning using string comparison

src/SystemPaths.cc
methylDragon: https://github.com/gazebosim/gz-common/pull/365/commits/cb5f31e2958f927c0bcab1db07114fd24df82668

src/SystemPaths.cc
methylDragon: https://github.com/gazebosim/gz-common/pull/365/commits/cb5f31e2958f927c0bcab1db07114fd24df82668

src/SystemPaths.cc
methylDragon: https://github.com/gazebosim/gz-common/pull/365/commits/cb5f31e2958f927c0bcab1db07114fd24df82668

src/SystemPaths.cc
methylDragon: https://github.com/gazebosim/gz-common/pull/365/commits/cb5f31e2958f927c0bcab1db07114fd24df82668

src/SystemPaths.cc
```diff
@@ -165,12 +223,25 @@ std::string SystemPaths::FindSharedLibrary(const std::string &_libName)
...
+  std::string result;
+
   if (!this->dataPtr->filePathEnv.empty())
   {
     this->ClearFilePaths();
-    std::string result;
     if (env(this->dataPtr->filePathEnv, result))
     {
+      // TODO(CH3): Deprecated. Remove on tock.
+      std::string ign_prefix = "IGN_";
```
chapulina, line 234: Style nit

```suggestion
      std::string ignPrefix = "IGN_";
```

src/SystemPaths.cc
```diff
@@ -117,6 +146,26 @@ std::string SystemPaths::LogPath() const
...
+  std::string result;
+  if (!this->dataPtr->pluginPathEnv.empty())
+  {
+    if (env(this->dataPtr->pluginPathEnv, result))
+    {
+      // TODO(CH3): Deprecated. Remove on tock.
+      std::string ign_prefix = "IGN_";
+
+      // Emit warning if env starts with IGN_
+      if (env_.compare(0, ign_prefix.length(), ign_prefix) == 0)
```
chapulina, line 160: Compilation is failing here:

```
/home/jenkins/workspace/ignition_common-ci-pr_any-ubuntu_auto-amd64/ign-common/src/SystemPaths.cc:160:11: error: ‘env_’ was not declared in this scope
  160 |       if (env_.compare(0, ign_prefix.length(), ign_prefix) == 0)
      |           ^~~~
```

chapulina (changes requested):

src/SystemPaths.cc
methylDragon: `env_` -> `_env` :man_facepalming: 
https://github.com/gazebosim/gz-common/pull/365/commits/7222e6b81141e3be809cd89463d81432278cc7df

src/SystemPaths.cc
methylDragon: https://github.com/gazebosim/gz-common/pull/365/commits/7222e6b81141e3be809cd89463d81432278cc7df

src/SystemPaths.cc
methylDragon: I didn't fix the other one, oops..

https://github.com/gazebosim/gz-common/pull/365/commits/26f7f5b5115f36eaaf221d05b97e2fa32d3d9143

chapulina (approved):

PR #367, opened 2022-06-07
Fix segfault when destroying Event objects

azeey:
# 🦟 Bug fix

## Summary

The segfault is caused when attempting to polymorphically destroy an `EventT` object that was instantiated inside a plugin (or any shared library loaded via dlopen) and the plugin has since been unloaded. Since `EventT` is a template, it's destructor is instantated in the translation unit that is part of the plugin and when that plugin is unloaded, the destructor is removed from memory as well. Gazebo stores pointers of the base class `Event`  (not `EventT`) in a container. Since the destructor of `Event` is virtual, the system will try to call the destructor of `EventT` when destroying that container, but that function is no longer available.

The solution is to ensure the destructor of `EventT` is never called. In this PR, this is accomplished by making `Event`'s destructor non-virtual and moving all data members to the base class. I believe it's safe to not call `EventT`s destructor as long as the destructor is trivial and `EventT` has no data members, but I would appreciate any feedback if this is not the case. 

Note, this will not fix the issue if an application stores `EventT` objects (instead of `Event` objects) instantiated by plugins and the plugins get unloaded. 

To test, run `ign gazebo shapes.sdf` and exit using `ctrl-c`.

## Checklist
- [x] Added tests

### DIFF
--- a/events/include/gz/common/Event.hh
+++ b/events/include/gz/common/Event.hh
@@ -17,6 +17,7 @@
 #ifndef GZ_COMMON_EVENT_HH_
 #define GZ_COMMON_EVENT_HH_
 
+#include <any>
 #include <atomic>
 #include <functional>
 #include <list>
@@ -41,11 +42,21 @@ namespace gz
       public: Event();
 
       /// \brief Destructor
-      public: virtual ~Event();
+      // WARNING: This should never be changed to virtual. See the class
+      // documentation for EventT. Otherwise, segfaults may occur in downstream
+      // libraries.
+      public: ~Event();
 
-      /// \brief Disconnect
-      /// \param[in] _id Integer ID of a connection
-      public: virtual void Disconnect(int _id) = 0;
+      /// \brief Disconnect a callback to this event.
+      /// \param[in] _id The id of the connection to disconnect.
+      public: virtual void Disconnect(int _id);
+
+      /// \brief Get the number of connections.
+      /// \return Number of connections.
+      unsigned int ConnectionCount() const
+      {
+        return this->connections.size();
+      }
 
       /// \brief Get whether this event has been signaled.
       /// \return True if the event has been signaled.
@@ -55,8 +66,53 @@ namespace gz
       /// \param[in] _sig True if the event has been signaled.
       public: void SetSignaled(bool _sig);
 
+      /// \internal
+      /// \brief Removes queued connections.
+      /// We assume that this function is called from a Signal function.
+      protected: void Cleanup();
+
       /// \brief True if the event has been signaled.
       private: bool signaled;
+
+      /// \brief A private helper class used in maintaining connections.
+      protected: class EventConnection
+      {
+        /// \brief Constructor
+       public: EventConnection(bool _on, const std::any &_cb,
+                        const ConnectionPtr &_publicConn)
+            : callback(_cb), publicConnection(_publicConn)
+        {
+          // Windows Visual Studio 2012 does not have atomic_bool constructor,
+          // so we have to set "on" using operator=
+          this->on = _on;
+        }
+
+        /// \brief On/off value for the event callback
+        public: std::atomic_bool on;
+
+        /// \brief Callback function
+        public: std::any callback;
+
+        /// \brief A weak pointer to the Connection pointer returned by Connect.
+        /// This is used to clear the Connection's Event pointer during
+        /// destruction of an Event.
+        public: std::weak_ptr<Connection> publicConnection;
+      };
+
+      /// \def EvtConnectionMap
+      /// \brief Event Connection map typedef.
+      protected: typedef std::map<int, std::unique_ptr<EventConnection>>
+                    EvtConnectionMap;
+
+      /// \brief Array of connection callbacks.
+      protected: EvtConnectionMap connections;
+
+      /// \brief A thread lock.
+      private: std::mutex mutex;
+
+      /// \brief List of connections to remove
+      private: std::list<typename EvtConnectionMap::const_iterator>
+              connectionsToRemove;
     };
 
     /// \brief A class that encapsulates a connection.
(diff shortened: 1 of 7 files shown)


azeey:
I've suppressed windows C4251 warnings in https://github.com/gazebosim/gz-common/pull/367/commits/5a6d88d9be5647829a488e2dd409f10c3b1b657e although some of the data members are protected and maybe accessed by derived classes (I think the suppression is meant for when the members are private https://github.com/gazebosim/gz-utils/blob/cdd48b33405fe42f400f8ca1238115c0bcd265e3/include/gz/utils/SuppressWarning.hh#L63-L65). If this is a concern, we can make them private and create accessor member functions.

events/include/gz/common/Event.hh
```diff
@@ -93,39 +151,37 @@ namespace gz
...
+    // WARNING: This class should not have any data members. Instead all data
+    // members should go into the base class `Event`. This is to ensure that
+    // this class behaves as expected when used with shared libraries (plugins)
+    // that get unloaded after emitting an event.
+    //
+    // Explanation: TODO
+    //
     template<typename T, typename N = void>
-    class EventT : public Event
+    class EventT final : public Event
```
mjcarroll, line 170: Are we sure that no one is currently inheriting from `EventT<T>`?  I can't recall anywhere in our libraries that we are doing it.

events/include/gz/common/Event.hh
```diff
@@ -93,39 +151,37 @@ namespace gz
...
       public: template<typename T, typename N> friend class EventT;
+      public: friend class Event;
     };
 
     /// \brief A class for event processing.
     /// \tparam T function event callback function signature
     /// \tparam N optional additional type to disambiguate events with same
     ///   function signature
+    //
+    // WARNING: This class should not have any data members. Instead all data
```
mjcarroll, line 162: Are we sure that no one is currently doing this?

events/include/gz/common/Event.hh
```diff
@@ -55,8 +67,54 @@ namespace gz
...
+        /// \brief A weak pointer to the Connection pointer returned by Connect.
+        /// This is used to clear the Connection's Event pointer during
+        /// destruction of an Event.
+        public: std::weak_ptr<Connection> publicConnection;
+      };
+IGN_UTILS_WARN_IGNORE__DLL_INTERFACE_MISSING
+      /// \def EvtConnectionMap
+      /// \brief Event Connection map typedef.
+      protected: typedef std::map<int, std::unique_ptr<EventConnection>>
+                    EvtConnectionMap;
```
mjcarroll, line 106: Prefer `using`.  I think we can probably switch to `unordered_map` here as well?

```suggestion
      protected: using EvtConnectionMap = std::map<int, std::unique_ptr<EventConnection>>;
```

events/include/gz/common/Event.hh
```diff
@@ -55,8 +67,54 @@ namespace gz
...
+          // Windows Visual Studio 2012 does not have atomic_bool constructor,
+          // so we have to set "on" using operator=
+          this->on = _on;
+        }
+
+        /// \brief On/off value for the event callback
+        public: std::atomic_bool on;
+
+        /// \brief Callback function
+        public: std::any callback;
```
mjcarroll, line 95: Why the change from std::function to any?

mjcarroll (commented):
The approach seems sound.  I just want to make sure that we aren't counting on current behaviors downstream (at least in our implementations).

events/include/gz/common/Event.hh
azeey: I don't know. As in my previous comment, I made it `final` to prevent users from inheriting from this and adding data members. We can drop the `final` and just warn users not to add data members if they inherit from `EventT`, but that doesn't seem reliable.

events/include/gz/common/Event.hh
azeey: Yes, I was able to compile all gz libraries without any issues. But as I mentioned in the test, users that call `delete` through a base pointer might get a compiler warning saying that `Event` doesn't have a virtual destructor. This warning doesn't occur when using a smart pointer though.

azeey (commented):
I've added a test in 0c03f6b02364f3227b29463b1d0b84bf4a302c87.

azeey:
The solution in this PR relies on an undefined behavior (deleting through a base pointer even though it's not virtual) and puts an unusual constraint on users of the EventT class (no data members). So, I explored a different solution using `std::shared_ptr` and `std::weak_ptr` where `EventT` objects created by plugins are stored in the `Connection` returned by `EventT::Connect` as shared pointers. In addition, in gz-sim's `EventManager`, we store weak pointers to the events instead of (unique_ptrs)(https://github.com/gazebosim/gz-sim/blob/bfbe8688ff94bd4e10a49ad94c28b36f8db38ddd/include/gz/sim/EventManager.hh#L72). The hope was that when a plugin gets unloaded, it deletes its shared pointer and if the event is no longer being used, it will be deleted. Subsequently, when `EventManager` tries to lock the weak pointer, it would fail, at which point, it will create a new `EventT` object that resides in a different plugin/shared library (this would happen when `EventManager::Connect` is called from a plugin). 

That seems to fix the segfault when exiting gz-sim, but upon further testing in a more narrowly scoped test, I've found that even the destruction of the weak pointer stored in `EventManger` can attempt to access memory that has been unloaded when a plugin got unloaded. So, unfortunately, I don't think this is a good solution either. 

The only solution I'm left with is to provide a new API in gz-plugin that allows loading plugins with `RTLD_NODELETE` set. This ensures that the shared librares associated with plugins are never actually deleted when `dlclose` is called. I've proved to myself that this works and fixes the segfault issue, but it warrants further discussion with the team.

mjcarroll:
Superceded by https://github.com/gazebosim/gz-plugin/pull/102 and friends.

Issue #344, opened 2022-04-16
Analysis of external mesh loading libraries and comparison with the existing approach


onurtore:
## Additional context

Gazebo's mesh loading library does not support glTF2 and FBX mesh files. There are two ways to extend the capabilities of the Gazebo's mesh loading system; (A) Using an external open-source library or (B) Writing glTF2 and FBX mesh loaders in a consistent way with the current approach.

Analysis is required to decide which approach is better. I am planning to generate such analysis in the next couple of days, so any idea is welcome. The main aspect of the analysis will be around:

1.Maturity Analysis
This section covers external library’s supported file types, and development history.

2. Complexity Analysis
This section mainly covers how hard it is to understand the design structure of candidate libraries.

3.Implementation/Integration & Testing Analysis

This section covers the required implementation and integration steps for each candidate library. Additionally covers an analysis of both internal and external possible testing scenarios for unit and integration tests.

I created an additional document for this analysis in [here](https://docs.google.com/document/d/1bs38PMkIK1B6EcN0l4knN2lFnWhiU4hVGH9SBK0zUMU/edit?usp=sharing). I will start my analysis with [assimp](https://github.com/assimp/assimp).

chapulina:
Thank you for the ticket and for looking into it, @onurtore . It would be great to add this capability to Gazebo. Looking forward to seeing the results of your analysis.

onurtore:
small update: Very basic transformation algorithm written (Just considers, vertices,indices and normals like the STLLoader.), now going to integrate Assimp to Gazebo to create speed and size metrics. If possible I will write this as plugin.

luca-della-vedova:
Hi! I thought it would be interesting to try out assimp and am in the process of writing an assimp loader class following the collada / obj / stl loader classes to load assets through assimp.
My hope is to be able to both parse all the asset formats supported by assimp (including but not limited to fbx and gltf) and reduce the size of the gz-common codebase by removing all the custom loaders.
It's been going reasonably smoothly so far, keeping a checklist of tasks and things remaining below:

- [ x ] Meshes (vertices, normals, faces)
- [ x ] Materials - external textures
- [ ] Materials - embedded textures (i.e. GLB)
- [ ] Skeletons
- [ ] Animations

For now I'll be focusing on parsing obj, dae and glb since those are the formats I personally use the most, I'll also try to unify all the unit tests in the collada / stl / obj loader into a single unit test file for assimp to try keep coverage high.
[...]

PR #368, opened 2022-06-09
Support compressed images and RGBA data in Image class

luca-della-vedova:
# 🎉 New feature

Work towards #344 , specifically https://github.com/gazebosim/gz-common/issues/344#issuecomment-1141869020 to allow loading embedded textures.
Main user of this API would be gz-rendering, discussion on downstream API taking place in the [gz-rendering PR](https://github.com/gazebosim/gz-rendering/pull/642) 

## Summary
This PR adds support for loading compressed images from in memory binary blobs, this is achieved through a new `SetFromCompressedData` function that takes as an input a byte array and a format and loads the image internally. For now only PNG is supported but other formats can be added trivially by adding a new case to the if statement.
Since images are provided as binary blobs, height and width are not known when the function is called. An alternative approach could be to use `SetFromData` with the byte buffer being passed as an image of `1 * len(buffer)` size, but I thought that wasn't too clean so went for another function.

It also adds a `RGBAData` function that returns the image in RGBA8 format. The main use case for this is that loading a PNG without an alpha channel through `FreeImage` would return an RGB8 24 bit image. This format might be tricky to use for downstream users because it's not aligned to the 32 bit boundary. This is specifically an issue in OGRE2 which is unable to transfer 24 bit textures to GPU (ref [here](https://github.com/OGRECave/ogre-next/blob/master/OgreMain/include/OgrePixelFormatGpu.h#L187-L191)).
Hence the convenience function is added for downstream users to make sure the image is always in a 32 bit RGBA8 format.  

Note, I'm personally not a big fan of the API for the `Data` functions where a user has to remember to manually delete[] the pointer they passed as soon as they are done with it, it would probably be safer to either return a `std::vector` of bytes (at the risk of expensive copies happening if people are not careful) or a `std::shared_ptr` to a `std::vector` of bytes, which would make sure copies only increase the reference count and memory is still correctly deallocated.
However having this API would have required changing the behavior of the `DataImpl` function that is used in other parts of the library so I didn't quite know if this was a good idea.

## Test it
Unit tests have been added, manual test is significantly more complex since it requires using the branches in #344 and the [gz-rendering PR](https://github.com/gazebosim/gz-rendering/pull/642), having a GLB asset and making sure the asset is correctly displayed.

## Checklist
- [x] Added tests


graphics/src/Image.cc
```diff
@@ -242,6 +242,28 @@ void Image::SetFromData(const unsigned char *_data,
...
+                                  unsigned int _size,
+                                  Image::PixelFormatType _format)
+{
+  if (this->dataPtr->bitmap)
+    FreeImage_Unload(this->dataPtr->bitmap);
+  this->dataPtr->bitmap = nullptr;
+
+  if (_format == COMPRESSED_PNG)
+  {
+    FIMEMORY *fiMem = FreeImage_OpenMemory(const_cast<unsigned char*>(_data), _size);
```
mjcarroll, line 256: `const_cast` is a little concerning here.  Can we change the API to reflect that `const`-ness may be violated?

luca-della-vedova: Done https://github.com/gazebosim/gz-common/pull/368/commits/8054b0d78433da3de936ffc9b81d87295207a26f

mjcarroll:
> However having this API would have required changing the behavior of the DataImpl function that is used in other parts of the library so I didn't quite know if this was a good idea.

I agree completely.  Fixing this API has been on my someday/maybe list for a while.  While it is out of the scope for this PR, if anyone on your side has some spare cycles, I would really appreciate a cleaner API before Garden.

mjcarroll:
Windows failing is expected.

luca-della-vedova:
> > However having this API would have required changing the behavior of the DataImpl function that is used in other parts of the library so I didn't quite know if this was a good idea.
> 
> I agree completely. Fixing this API has been on my someday/maybe list for a while. While it is out of the scope for this PR, if anyone on your side has some spare cycles, I would really appreciate a cleaner API before Garden.

I'm happy to look at it, I ticketed an issue https://github.com/gazebosim/gz-common/issues/369 and can start working on it if the idea sounds reasonable to you.

mjcarroll:
> I'm happy to look at it, I ticketed an issue https://github.com/gazebosim/gz-common/issues/369 and can start working on it if the idea sounds reasonable to you.

Great, thanks!  I think improving this API will both make it more reasonable for users as well as make it generally more efficient as we are doing a lot of reallocations and copies.

mjcarroll (approved):
LGTM, I just added two more quick test cases so we can get :100: coverage on the diff.

PR #372, opened 2022-06-21
Add API for getting Image bytes as std::vector

### LINKED ISSUE #369: Improving Image Data() API to reduce chance of memory leaks
While working on a [PR](https://github.com/gazebosim/gz-common/pull/368) to add support for compressed images to gz-common, the non-ideal behavior of the current API, that is quite prone to memory leaks, came up.

## Desired behavior
It should be safe to use the `Image.Data()` functions without a risk of segfault or memory leaks.

Right now the function takes as an input an `unsigned char **_data` and a `unsigned int& _count` that will be used as output parameters.
There are several vulnerabilities in this API:

`_data` will be freed internally if not equal to nullptr. If the user was to declare but not initialize the variable, i.e.:

```
unsigned char *data;
unsigned int count;
image.Data(&data, count);
```

The program would segfault.

Another issue is that the user needs to know to free manually the data variable after they are done, and again this is quite prone to error.
Finally, the API is slightly confusing since one argument is passed by reference, one by a pointer, and all the parameters are output parameters, it would probably be a bit clearer if it just had a return value.

## Alternatives considered
A possibility could be to create a new Data function that takes no parameters and returns either;
* `std::vector<unsigned char>`: It's great because it can be used out of the box but users have to be a bit careful of copies since the structure containing possibly multiple megabytes of data it can go out of control quite quickly.
* `std::shared_ptr<std::vector<unsigned char>>`: Slightly more work since there is dynamic allocation to do but still memory safe and copies now only change the reference count instead of copying the whole image.

Wildcard for completeness, it seems there is a std::byte [type in C++-17](https://en.cppreference.com/w/cpp/types/byte) which sounds in theory perfect for the purpose, however since downstream users will likely do mathematical operations (i.e. pixelwise operations) and std::byte seems not to allow them, it might be counterproductive.

## Implementation suggestion

My instinct would be to go for the `shared_ptr` approach, returning the vector itself sounds like it would be the cleanest API but when working with downstream packages I noticed that clones and copies happen quite a lot (i.e. before a texture gets rendered by gz-rendering it gets copied at least twice, although I didn't keep track of the destructors to see if it's always cleared).

So my recommendation would be to add new Data() - RGBData() - RGBAData() functions:

```
std::shared_ptr<std::vector<unsigned char>> Data() const;
```

And mark the current

```
void Data(unsigned char **_data, unsigned int &_count) const;
```

as deprecated to be removed in the next release. Then we can refactor (or probably duplicate and deprecate in the same way?) the DataImpl function to be able to return a `shared_ptr`.
We can rely on C++ figuring out from the parameters what function is being called so this behavior should, in theory, allow a full tick/tock cycle without breaking existing behavior and allow users to migrate before Gazebo H.

luca-della-vedova:
# 🎉 New feature

Closes #369 

## Summary
This PR adds a new API for `common::Image` to get the data as an `std::vector<unsigned char>` and deprecates the API where the user is expected to pass a pointer, a length and free the memory manually.

There is a fair bit of code duplication, I was trying to come up with a way to reuse some code but it would have probably made deprecation harder in the future.
For example I thought of having the std::vector version of the function just copy the raw array returned by the `Data` functions but that would have required refactoring when they are removed. In alternative I could have the old `Data` functions just allocate a C array and copy the std::vector's data, but now their performance would get quite a bit worse (need two allocations instead of one)
"Thanks to" the code duplication all the deprecated Data / DataImpl functions can be safely deleted and no visible behavior change should be exposed to the users.

## Test it
All the unit tests with the old API have been converted to use the new API and are still successful. I also added a new test case for the deprecated functions to make sure the behavior between the old and the new ones is exactly the same (all the bytes match). The deprecated tests have deprecation warning silencing to avoid cluttering the build output and are in a separate test case so they can be easily removed in the future.

## Checklist
- [x] Added tests


luca-della-vedova:
> There's a new deprecation warning, we should update `ImageHeightmap` to use the new API.
> 
> ```
> /Users/jenkins/workspace/ignition_common-ci-pr_any-homebrew-amd64/ign-common/geospatial/src/ImageHeightmap.cc:63:13: warning: 'Data' is deprecated [-Wdeprecated-declarations]
>   this->img.Data(&data, count);
>             ^
> ```

Done thanks!
Happy to see that this is a case where the new API indeed prevents a memory leak, if the heightmap was in an unsupported format we would [return the function](https://github.com/gazebosim/gz-common/pull/372/commits/c822fc056d1a2debe3f8f4f4afe9e85fd4f6431a#diff-6e6a9808b8db8350a59b86dad4ecada1b18b80b693bf0e7685652b559508389dR91) early and `delete[]` wouldn't be called, leaking the memory

mjcarroll:
I'm happy with this, but I think we should fix usage up the stack before merging to avoid introducing a whole bunch of deprecation warnings.

I would propose fixing those first, or if you are constrained for time, removing the deprecations until we get cycles to fix the warnings.

luca-della-vedova:
> I'm happy with this, but I think we should fix usage up the stack before merging to avoid introducing a whole bunch of deprecation warnings.
> 
> I would propose fixing those first, or if you are constrained for time, removing the deprecations until we get cycles to fix the warnings.

Agreed, that's pretty trivial and I actually prepared all the branches a few days ago, I didn't open the PRs to avoid making too much noise until this was in a "good enough state".
Sounds like we reached that point so here they are:

[gz-gui](https://github.com/gazebosim/gz-gui/pull/430)
[gz-sim](https://github.com/gazebosim/gz-sim/pull/1563)

mjcarroll:
Getting this in for nightlies so that others can merge tomorrow am.

PR #375, opened 2022-06-22
Fix: Reversed models when skeletons exist

onurtore:
Creating skeleton from Assimp nodes ends up with reversed models/weird arms, because some of the nodes does not have corresponding bones. Correspondence problems creates bones without inverse transform. While utilizing removeEmptyBones flag to false solves the issue of weird arm problem, the reversed model problem still exists. However, this problem does not originate from Assimp, the problem arise because we are trying to add nodes that are not related to bones. These nodes also not defined as joints in the original dae files so Assimp not loading them as bone is logical and correct behaviour.

To solve this issue, this code traverses the bone structure, and stores the name of the bones. These bone names used during skeleton creation to pass the nodes without corresponding bones while preserving the possible transformations. Finally, scene node removed from skeleton.

3_cowboys


onurtore:
@luca-della-vedova,
I am creating a test bed for this.

graphics/src/AssimpLoader.cc
```diff
@@ -489,6 +530,9 @@ Mesh *AssimpLoader::Load(const std::string &_filename)
       // TODO parse different skeletons and merge them
       this->dataPtr->RecursiveSkeletonCreate(rootNode->mChildren[childIdx], rootSkelNode, rootTransform);
     }
+    rootSkelNode = rootSkelNode->Child(0); //We dont need scene node and adding will create inverse model
```
onurtore, line 533: Does not look safe to me? But if this is not the case, then whole logic is wrong.

onurtore:
@luca-della-vedova , 
Depending on your input, I will move into the next problem, animation interpolation.

graphics/src/AssimpLoader.cc
luca-della-vedova: Agree, I think this is a good time to start investigating the TODO above about meshes with multiple skeletons.
A classic example we have been using internally is the one below, where the human and the cane have two separate skeletons that are merged in the "legacy" ColladaLoader.

[PatientWalkingCane.zip](https://github.com/gazebosim/gz-common/files/8963136/PatientWalkingCane.zip)

This model segfaults both with this PR and without and by looking at the stdout debug output from `assimp` it seems it might actually be an assimp parsing issue since it reports a lot of errors about bones not being found.
It seems then that the fix might once again be on the assimp side, then if this approach works for that edge case I think we can keep this approach until we find a case for which it doesn't work.

graphics/src/AssimpLoader.cc
```diff
@@ -177,16 +182,50 @@ void AssimpLoader::Implementation::RecursiveCreate(const aiScene* _scene, const
...
+  for (unsigned meshIdx = 0; meshIdx < _node->mNumMeshes; ++meshIdx)
+  {
+    auto assimpMeshIdx = _node->mMeshes[meshIdx];
+    auto assimpMesh = _scene->mMeshes[assimpMeshIdx];
+    for (unsigned boneIdx = 0; boneIdx < assimpMesh->mNumBones; ++boneIdx)
+    {
+      auto bone = assimpMesh->mBones[boneIdx];
+      auto boneName = std::string(bone->mName.C_Str());
+      gzdbg << "\t" << "RecursiveStoreBoneNames: Bone name: " << boneName << std::endl;
+      BoneNames.push_back(boneName);
```
luca-della-vedova, line 199: Nit, I think the usual pattern is doing `this->` to access variables to make it clear that they are member variables

graphics/src/AssimpLoader.cc
```diff
@@ -49,6 +49,9 @@ class AssimpLoader::Implementation
   /// Convert a color from assimp implementation to Ignition common
   public: math::Color ConvertColor(aiColor4D& _color);
 
+  // Stores the loaded bone names 
+  public: std::vector<std::string> BoneNames;
```
luca-della-vedova, line 53: Since the only operations we do on this structure are insertion and lookup by name I'd recommend using a `std::unordered_set<std::string>` so lookup can be in (approximately) constant time. Using a vector forces us to use `std::find` that will have to traverse the whole structure to find a key.

Also variable names usually have the `lowerCamelCase` convention

graphics/src/AssimpLoader.cc
```diff
@@ -177,16 +182,50 @@ void AssimpLoader::Implementation::RecursiveCreate(const aiScene* _scene, const
   }
 }
 
+void AssimpLoader::Implementation::RecursiveStoreBoneNames(const aiScene *_scene, const aiNode *_node)
```
luca-della-vedova, line 185: I guess I can see how it would be nicer (but significantly more work, can probably postpone to a "if we have time") to have the new skeleton container in assimp to be populated also in Collada (or even GLB) assets

graphics/src/AssimpLoader.cc
```diff
@@ -177,16 +182,50 @@ void AssimpLoader::Implementation::RecursiveCreate(const aiScene* _scene, const
...
 {
   // First explore this node
   auto nodeName = ToString(_node->mName);
   // TODO check if node or joint?
-  auto skelNode = new SkeletonNode(_parent, nodeName, nodeName, SkeletonNode::JOINT);
-  // Calculate transform
+  auto boneExist = std::find(BoneNames.begin(), BoneNames.end(), nodeName) != BoneNames.end();
   auto nodeTrans = this->ConvertTransform(_node->mTransformation);
-  skelNode->SetTransform(nodeTrans);
+  auto skelNode = _parent;
```
luca-della-vedova, line 219: Just thought I'd check whether the logic (and my understanding) are correct:

* If the bone exists, same workflow as before, create a node then recurse through its children setting the current node as parent.
* If the bone does not exist, don't create the node and recurse through the children using the current node and its transformation as a parent

So I guess we will just "skip" ahead while keeping the transforms? Again this sounds reasonable to me but just thought I'd check if my understanding is correct

graphics/src/AssimpLoader.cc
```diff
@@ -177,16 +182,50 @@ void AssimpLoader::Implementation::RecursiveCreate(const aiScene* _scene, const
...
+    return;
+
+  for (unsigned meshIdx = 0; meshIdx < _node->mNumMeshes; ++meshIdx)
+  {
+    auto assimpMeshIdx = _node->mMeshes[meshIdx];
+    auto assimpMesh = _scene->mMeshes[assimpMeshIdx];
+    for (unsigned boneIdx = 0; boneIdx < assimpMesh->mNumBones; ++boneIdx)
+    {
+      auto bone = assimpMesh->mBones[boneIdx];
+      auto boneName = std::string(bone->mName.C_Str());
```
luca-della-vedova, line 197: Nit, I added a `ToString` function since this conversion is done all the time throughout the loader

luca-della-vedova (commented):
Tested locally and indeed `actor.sdf` looks good and other actors work as well! As I mentioned in the comments there probably remains a last tricky edge case of multiple skeletons (with an example mesh that shows the behavior). We are getting there!

onurtore:
Hi @ahcorde,
Just a progress update.

graphics/src/AssimpLoader.cc
onurtore: Done.

graphics/src/AssimpLoader.cc
onurtore: Done, thanks for heads up.

graphics/src/AssimpLoader.cc
onurtore: I will put "Use the new skeleton structure" to my TO-DO list. This will remove the need for this func.

graphics/src/AssimpLoader.cc
onurtore: Yes exactly.

graphics/src/AssimpLoader.cc
onurtore: Done.

graphics/src/AssimpLoader.cc
onurtore: Hi @luca-della-vedova,
Currently in my branch I cant see the reports related to bones not being found. I had seen it one time, but I think I lost it after some branch changes, can you share me the assimp, and gz-common branches you use? Prefabably after you merged this one.

onurtore:
Assimp PR I created for the missing bones merged.

I might not be too effective this weekend (I have graduation ceremony) but continue afterwards to fix this problem.

graphics/src/AssimpLoader.cc
onurtore: I can change the solution from removing the parent node to adding an identity transformation matrix if there is non.

onurtore:
Also *sometimes* the Gazebo stucks like below, is this happening to you?
 


When try to click buttons it says:

`[GUI] [Wrn] [Application.cc:811] [QT] file::/Gazebo/GazeboDrawer.qml:175:3: QML Dialog: Binding loop detected for property "implicitWidth"`

graphics/src/AssimpLoader.cc
onurtore: I can change the solution from removing the parent node to adding an identity transformation matrix if there is non exist?

onurtore:
Also looks like with this PR, frame interpolation error gone as well. I compared the _actor.sdf_ with the default loader, and checked the animations and I did not see any difference. 

Maybe after fixing the meshes with multiple skeletons it is time to ask OSRF' internal team to use this branch for test their meshes or share their meshes with us to test ? We can create a flag for using assimp loader during compilation for the initial one. Furthermore, maybe we can start doing memory and speed test?

graphics/src/AssimpLoader.cc
luca-della-vedova: > When I try to load the mesh with the assimp PR you created, the whole Gazebo justs stucks and prints malloc(): invalid size (unsorted)), and if I dont use that PR, then we see this errors related to bones not being found. Furthermore, when I use the old colladaLoader, then again I see the Malloc error as well... So maybe it is an error not related to loaders? Should we find another dae file with multiple skeletons?

That's a bit strange, the model works (or at least used to work under Fortress) with the default collada loader, I wonder if the issue is somewhere else? Can you check that it works with a binary Fortress installation just to be safe?

>  I can change the solution from removing the parent node to adding an identity transformation matrix if there is non exist?

That makes me feel a bit better, but I thought the transformation is always populated and is an identity for `Scene`? Is that not the case?

graphics/src/AssimpLoader.cc
luca-della-vedova: I saw positive feedback on your PR in `assimp` on this, I wonder if we can make it happen "soon enough"

luca-della-vedova:
> @luca-della-vedova , Also looks like with this PR, frame interpolation error gone as well. I compared the _actor.sdf_ with the default loader, and checked the animations and I did not see any difference.
> 
> Maybe after fixing the meshes with multiple skeletons it is time to ask OSRF' internal team to use this branch for test their meshes or share their meshes with us to test ? We can create a flag for using assimp loader during compilation for the initial one. Furthermore, maybe we can start doing memory and speed test?

I'd be happy to see some profiling results!

I think a way to really push the confidence in the implementation would be to add unit tests for new formats (for the minimum implementation of just using `assimp` to import GLB / FBX) and port the unit tests from the other loaders to use assimp and make sure they work (if we wanted to deprecate the internal loaders).
There is only so far manual testing can go and it would be good to have some automated CI to make sure the importing library is doing something sensible and avoid small errors / regressions.

Up to you which task you want to tackle first!

graphics/src/AssimpLoader.cc
luca-della-vedova: I see, then I guess setting the inverse bind transform to an identity sounds like a reasonable approach to me

graphics/src/AssimpLoader.cc
luca-della-vedova: None really, just remove duplicated code all over the place

graphics/src/AssimpLoader.cc
luca-della-vedova: Friendly ping for this @onurtore, I think we can merge it after this change and start to look into opening a PR against `main`

onurtore: Hi @luca-della-vedova ,
Adding inverse bind transform to the scene brings back the animation segmentation fault. So now I am not sure whether should we add that or not.

Btw, mesh works with Fortress however, I am porting everthing to Fortress to find whether assimp has problems or not.

onurtore: Hi @luca-della-vedova, 
Good news, this mesh also loaded with assimp in Fortress. The only problem was related to texture as you can see below. 


 
I am in favour of merge, but the next think I will do is the divide the CreateActor() function into subparts. I tested the createActor function with the static analysis tools and dividing into subparts looks like good idea.

Cyclomatic Complexity: 36
#Lines: 302

onurtore: Assigning child and setting parent null creates segmentation error in Edifice demo.

luca-della-vedova: Yea I'm not sure how stable this solution is. The issue is not in adding a child and setting a parent to null but rather accessing the hardcoded element at index 0 in cases in which the root node does not have any children (which happens very often for simple meshes).
I added the "if" statement to only execute this block when there are animations in the scene which mitigates the issue, since scene with animations will most likely have a number of nodes, but it would be good to remove the hardcoded value or make it safer

PR #385, opened 2022-07-02
Feature: Inits units tests for fbx and glTF2

onurtore:
AssimpLoader_TEST.cc updated to include basic unit tests for fbx and glb mesh files. New mesh files uploaded.

PR #393, opened 2022-07-13
Use assimp for loading meshes

### LINKED ISSUE #374: [Meta-ticket][GSoC22] Improved mesh support for Gazebo
This meta-ticket tracks the GSoC22 project named 'Improving Gazebo’s Mesh Format Support' by including relevant PRs, Issues, and Comments for each subtask.

 ## Done
***
- [X] Basic mesh support.
    * #354 
    * #355 
    * #358 
    * #359 
    * Assimp PR for fixing for removing empty flags. [#4585](https://github.com/assimp/assimp/pull/4585)

## To-Do
***
- [ ] Skeleton Support
       #344
- [ ] Animation Support
     *  Frame Interpolation
 - [ ] Runtime analysis  & Fallback system for Assimp
    * #360 

## Documentation Improvements
***
* [Adds a note about out of memory errors and provides a way to fix that](https://github.com/gazebosim/docs/pull/253)
* #347 
* [Adds note about how to activate debug symbols during source installation](https://github.com/gazebosim/docs/pull/246)

luca-della-vedova:
# 🎉 New feature

Closes #374 

## Summary
This PR introduces using the `assimp` library for parsing models and, with it, support for GLB, GLTF and FBX formats. 
For now it overrides all the custom loaders, there is no known regression with obj and stl, while there are a few regression with collada files.
Most of the regressions have been handled upstream at the `assimp` library level, however the fixes didn't make it to the released Debian package so they might not reflect unless `assimp` is built from source or a new release is triggered upstream.
A regression in `actor.sdf` that segfaults is known and currently under investigation.
Imo a nice stretch goal would be to deprecate as many loaders as possible and only maintain one but this can be a scary change so the alternative could be to just add yet another loader for "all other formats" and keep using the internal ones for obj, stl and dae. However, obj and stl seem to be simple and reliable enough that I believe they _could_ be removed.

Depends on a [gz-cmake PR](https://github.com/gazebosim/gz-cmake/pull/278) and release to add support for finding `assimp` in `gz_find_package` for CI to be green.

## Test it
All the unit tests from STL, Collada and OBJ loaders have been ported to the assimp loader and updated in cases where the behavior changed.
New unit tests for the new formats have been merged in #385 by @onurtore.
You will need a few glb assets to test this PR, I provided a few in the matching `gz-rendering` [PR](https://github.com/gazebosim/gz-rendering/pull/642#pullrequestreview-1047474661).

## Checklist
- [x] Added tests


graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,704 @@
...
+
+#include <queue>
+#include <unordered_set>
+
+#include <assimp/Logger.hpp>
+#include <assimp/DefaultLogger.hpp>
+#include <assimp/GltfMaterial.h>    // GLTF specific material properties
+#include <assimp/Importer.hpp>      // C++ importer interface
+#include <assimp/scene.h>           // Output data structure
+#include <assimp/postprocess.h>     // Post processing flags
```
ahcorde, line 38: alphabetixe

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,704 @@
...
+  /// \return the converted math::Matrix4d
+  public: math::Matrix4d ConvertTransform(const aiMatrix4x4& _matrix);
+
+  /// \brief Convert from assimp to gz::common::Material
+  /// \param[in] _scene the assimp scene
+  /// \param[in] _matIdx index of the material in the scene
+  /// \param[in] _path path where the mesh is located
+  /// \return pointer to the converted common::Material
+  public: MaterialPtr CreateMaterial(const aiScene* _scene, unsigned _matIdx,
+                                     const std::string& _path);
```
ahcorde, line 69:
```suggestion
  public: MaterialPtr CreateMaterial(const aiScene* _scene, 
                                     unsigned _matIdx,
                                     const std::string& _path);
```

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,704 @@
...
+  /// \param[in] _type the type of texture (i.e. Diffuse, Metal)
+  /// \return the generated texture name
+  public: std::string GenerateTextureName(const aiScene* _scene,
+                                          const aiMaterial*  _mat,
+                                          const std::string& _type);
+
+  /// \brief Function to parse texture information and load it if embedded
+  /// \param[in] _scene the assimp scene
+  /// \param[in] _texturePath the path where the texture is located
+  /// \param[in] _texturePath the name of the texture
```
ahcorde, line 90:
```suggestion
  /// \param[in] _textureName the name of the texture
```

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,704 @@
...
+  /// \param[in] _assimpMesh the assimp mesh to load
+  /// \param[in] _transform the node transform for the mesh
+  /// \return the converted common::Submesh
+  public: SubMesh CreateSubMesh(const aiMesh* _assimpMesh,
+                                const math::Matrix4d& _transform);
+
+  /// \brief Recursively create submeshes scene starting from the root node
+  /// \param[in] _scene the assimp scene
+  /// \param[in] _node the node being processed
+  /// \param[in] _node the transform of the node being processed
```
ahcorde, line 116:
```suggestion
  /// \param[in] _transform the transform of the node being processed
```

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,704 @@
...
+      _sm.a1, _sm.a2, _sm.a3, _sm.a4,
+      _sm.b1, _sm.b2, _sm.b3, _sm.b4,
+      _sm.c1, _sm.c2, _sm.c3, _sm.c4,
+      _sm.d1, _sm.d2, _sm.d3, _sm.d4);
+}
+
+void AssimpLoader::Implementation::RecursiveCreate(const aiScene* _scene,
+    const aiNode* _node, const math::Matrix4d& _transform, Mesh* _mesh)
+{
+  if (!_node)
```
ahcorde, line 179: check `_scene` too?

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,704 @@
...
+    // Finally recursive call to explore subnode
+    this->RecursiveCreate(_scene, child_node, nodeTrans, _mesh);
+  }
+}
+
+void AssimpLoader::Implementation::RecursiveStoreBoneNames(
+    const aiScene *_scene, const aiNode *_node,
+    std::unordered_set<std::string>& _boneNames)
+{
+  if (!_node)
```
ahcorde, line 248: check `_scene` too?

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,704 @@
...
+  public: void ApplyInvBindTransform(SkeletonPtr _skeleton);
+};
+
+// Utility function to convert to std::string from aiString
+static std::string ToString(const aiString& str)
+{
+  return std::string(str.C_Str());
+}
+
+math::Color AssimpLoader::Implementation::ConvertColor(aiColor4D& _color)
```
ahcorde, line 160: Add this in all method/function implementations
```suggestion
/////////////////////////////////////////////////
math::Color AssimpLoader::Implementation::ConvertColor(aiColor4D& _color)
```

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,704 @@
...
+    // Finally recursive call to explore subnode
+    this->RecursiveStoreBoneNames(_scene, child_node, _boneNames);
+  }
+}
+
+void AssimpLoader::Implementation::RecursiveSkeletonCreate(const aiNode* _node,
+    SkeletonNode* _parent, const math::Matrix4d& _transform,
+    const std::unordered_set<std::string> &_boneNames)
+{
+  // First explore this node
```
ahcorde, line 275: check `_node` nad `_parent` ?

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,704 @@
...
+  mat->SetPbrMaterial(pbr);
+  return mat;
+}
+
+std::pair<std::string, ImagePtr> AssimpLoader::Implementation::LoadTexture(
+    const aiScene* _scene,
+    const aiString& _texturePath,
+    const std::string& _textureName)
+{
+  std::pair<std::string, ImagePtr> ret;
```
ahcorde, line 436: check `_scene` ?

graphics/src/AssimpLoader_TEST.cc
```diff
@@ -0,0 +1,609 @@
+/*
+ * Copyright (C) 2016 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2022 Open Source Robotics Foundation
```

graphics/src/AssimpLoader_TEST.cc
```diff
@@ -0,0 +1,609 @@
...
+ *
+*/
+#include <gtest/gtest.h>
+
+#include "gz/common/Mesh.hh"
+#include "gz/common/SubMesh.hh"
+#include "gz/common/Material.hh"
+#include "gz/common/Skeleton.hh"
+#include "gz/common/SkeletonAnimation.hh"
+#include "gz/common/AssimpLoader.hh"
```
ahcorde, line 24: alphabetize

graphics/src/AssimpLoader_TEST.cc
```diff
@@ -0,0 +1,609 @@
...
+}
+
+/////////////////////////////////////////////////
+TEST_F(AssimpLoader, LoadZeroCount)
+{
+  common::AssimpLoader loader;
+  common::Mesh *mesh = loader.Load(
+      common::testing::TestFile("data", "zero_count.dae"));
+  ASSERT_TRUE(mesh);
+  // This is custom logging messages, not implemented in Assimp
```
ahcorde, line 167: remove ?

graphics/src/CMakeLists.txt
```diff
@@ -27,13 +28,17 @@ gz_build_tests(
     gz-common${GZ_COMMON_VER}-testing
 )
 
+message ("ASSIMP INCLUDE IS ${ASSIMP_INCLUDE_DIRS}")
```
ahcorde, line 31: remove ?

graphics/src/MeshManager.cc
```diff
@@ -138,6 +147,8 @@ const Mesh *MeshManager::Load(const std::string &_filename)
         extension.begin(), ::tolower);
     MeshLoader *loader = nullptr;
 
+    loader = &this->dataPtr->assimpLoader;
```
ahcorde, line 150: we discussed at the Gazebo meeting that colladaLoader and objLoader might have some adhoc tags or internal modifications (in the daes or objs). We are thinking about adding a cmd flag to be able to use assimp with an argument but keep the current Collada and objloader as the default behaviour.

FYI @onurtore

ahcorde (dismissed):
@onurtore do you mind to review and test this PR ?

there are some TODOs in the code, maybe you can work in some of them

graphics/src/MeshManager.cc
luca-della-vedova: OK! That makes development on this PR considerably simpler since there won't be any risk to break existing behavior. Just to confirm the intention is also to keep STLLoader right? Therefore assimp will be used only for the new formats?

graphics/src/AssimpLoader.cc
luca-della-vedova: `_scene` is guaranteed to be non empty, the check for it being null is the first operation being done after the object creation and if it is there is an error printout, early return and nothing else is executed, specifically [here](https://github.com/gazebosim/gz-common/blob/luca/assimp_sandbox/graphics/src/AssimpLoader.cc#L593-L597)

luca-della-vedova:
Two high level items that I'd like some feedback on:

If there is no intention to replace the default loaders, should I remove all the unit tests for OBJ / Collada / STL? If we only use the new loader for new formats they are not that relevant and might make migrations tricky / tests flaky if the implementation of `assimp` changes between versions. On the other hand if we want to offer it through a flag they should probably still be there.

This PR technically includes two somewhat separate features, the main one is using assimp to load models, however some of these models (i.e. GLB) might have embedded textures so there is a need for an API for passing textures as binary blobs between gz libraries, instead of file names.
Discussion on the API mostly took place in [gz-rendering](https://github.com/gazebosim/gz-rendering/pull/642) since it's the primary user of it.
If it would make review easier I could split this PR to have the material API in a new PR, which would remove all the changes in the Material / Pbr files, I didn't do it earlier because technically speaking these APIs have no use without this loader so if they end up merged without the loader they would just bloat the code unnecessarily

graphics/src/AssimpLoader.cc
luca-della-vedova: https://github.com/gazebosim/gz-common/pull/393/commits/d647967caf54768f45b9e9bd296c858c85ec4487

graphics/src/AssimpLoader.cc
luca-della-vedova: https://github.com/gazebosim/gz-common/pull/393/commits/d647967caf54768f45b9e9bd296c858c85ec4487

graphics/src/AssimpLoader.cc
luca-della-vedova: https://github.com/gazebosim/gz-common/pull/393/commits/d647967caf54768f45b9e9bd296c858c85ec4487

graphics/src/AssimpLoader.cc
luca-della-vedova: https://github.com/gazebosim/gz-common/pull/393/commits/d647967caf54768f45b9e9bd296c858c85ec4487

graphics/src/AssimpLoader.cc
luca-della-vedova: https://github.com/gazebosim/gz-common/pull/393/commits/d647967caf54768f45b9e9bd296c858c85ec4487

graphics/src/AssimpLoader_TEST.cc
luca-della-vedova: https://github.com/gazebosim/gz-common/pull/393/commits/d647967caf54768f45b9e9bd296c858c85ec4487

graphics/src/AssimpLoader_TEST.cc
luca-della-vedova: https://github.com/gazebosim/gz-common/pull/393/commits/d647967caf54768f45b9e9bd296c858c85ec4487

graphics/src/CMakeLists.txt
luca-della-vedova: https://github.com/gazebosim/gz-common/pull/393/commits/d647967caf54768f45b9e9bd296c858c85ec4487

graphics/src/AssimpLoader.cc
luca-della-vedova: https://github.com/gazebosim/gz-common/pull/393/commits/d647967caf54768f45b9e9bd296c858c85ec4487

graphics/src/MeshManager.cc
ahcorde: I did't remember nothing about STLs

> Therefore assimp will be used only for the new formats?

Yes, that's the idea for now, or maybe add a CMD flag to allow to use ASSIMP. @onurtore how do you feel about openning a PR on top of this ?

ahcorde:
> If there is no intention to replace the default loaders, should I remove all the unit tests for OBJ / Collada / STL? If we only use the new loader for new formats they are not that relevant and might make migrations tricky / tests flaky if the implementation of assimp changes between versions. On the other hand if we want to offer it through a flag they should probably still be there.

For now we want to keep OBJ and Collada loaders, these two loaders might contain adhoc code. We were only thinking about adding a flag to allow to use ASSIMP in all cases. This way we can provide a transition period.

> If it would make review easier I could split this PR to have the material API in a new PR, which would remove all the changes in the Material / Pbr files, I didn't do it earlier because technically speaking these APIs have no use without this loader so if they end up merged without the loader they would just bloat the code unnecessarily

you can keep it like it's

luca-della-vedova:
OK I think we are in reasonably good state!

I moved away from the source build of assimp and made the PR compatible with both Ubuntu 20.04 (assimp 5.0.x) and Ubuntu 22.04 (assimp 5.2.x). Compatibility is handled through a define flag that disables features / tests that are known not to work in the older assimp version.
I ran tests locally on both my 20.04 and 22.04 installation and they work well.

Support for 20.04 is decent but not as good, specifically PBR support for GLTF is not present, 22.04 has all the features.
I also reverted the change of using assimp for everything, sadly we rely on a bunch of merged but not released upstream PRs  and one unmerged PR, hence the behavior on binary builds even in Ubuntu 22.04 for collada files wouldn't be great.

There are unit tests for both the collada / stl / obj (where appropriate), as well as basic cubes for both FBX and GLB.

luca-della-vedova:
Another note is that in order to support materials with textures stored as binary blobs (i.e. glb or base64 gltf) it was necessary to add an API to set and retrieve texture data for normal and PBR textures (or more accurately the subset of PBR that is supported in GLB).
This uses the `shared_ptr` approach (i.e. set a texture through `std::string name,  shared_ptr<common::Image> data = nullptr` to avoid having to do multiple MB sized allocations when copying materials between libraries but it does make the API slightly more clunky.
If the performance drawback of setting and copies vector of bytes is OK I'm happy to change it to just be a `gz::common::Image`.

I think all the functionality is ready, the only part that I'll try to work with in the coming days is increasing coverage by having glb assets with more features for unit tests (i.e. with embedded PBR textures to test that branch of the code that is currently not covered).

Migration.md
```diff
@@ -43,6 +43,8 @@ release will remove the deprecated code.
    10. `IGN_CREATE_SPEC_INTERFACE`
    11. `IGN_DUMMY_PLUGIN_PATH`
    12. `IGNITION_UNITTEST_SPECIALIZED_PLUGIN_ACCESS`
+8. The `Image::Data(unsigned char**, unsigned int&)` functions that accept a pointer and a size and internally allocate memory are deprecated and will be removed in future versions.
+   Use the new `Data` functions that return a `std::vector<unsigned char>` to have automatic memory management.
```
luca-della-vedova, line 47: I merged locally https://github.com/gazebosim/gz-common/pull/372 to make testing the downstream gz-rendering PR easier. The diff will get shorter once it is merged and we merge main back into this branch

graphics/src/Pbr_TEST.cc
```diff
@@ -28,10 +29,14 @@ TEST(Pbr, BasicAPI)
...
   EXPECT_EQ(std::string(), pbr.AlbedoMap());
   EXPECT_EQ(std::string(), pbr.NormalMap());
+  EXPECT_EQ(nullptr, pbr.NormalMapData());
   EXPECT_EQ(gz::common::NormalMapSpace::TANGENT, pbr.NormalMapType());
   EXPECT_EQ(std::string(), pbr.RoughnessMap());
+  EXPECT_EQ(nullptr, pbr.RoughnessMapData());
   EXPECT_EQ(std::string(), pbr.MetalnessMap());
+  EXPECT_EQ(nullptr, pbr.MetalnessMapData());
   EXPECT_EQ(std::string(), pbr.EmissiveMap());
+  EXPECT_EQ(nullptr, pbr.EmissiveMapData());
```
luca-della-vedova, line 39: I didn't go too deep in editing all the tests for the PBR object, it still has a lot of test code for testing all the copy, assignment, copy-assignment, move operators etc. that are a leftover from when the class was not using the `gz-utils` implementation pointer and all the operators were implemented manually, i.e. [here](https://github.com/gazebosim/gz-common/pull/163/files#diff-60e733fab278ef8fb8a50a97bd9270374f2cf2cc67b2bafad190eb0ea0e47f46L60-L80). I believe all the tests are now unnecessary and can be removed but that's out of the scope of this (already too wide!) PR.

graphics/include/gz/common/AssimpLoader.hh
```diff
@@ -0,0 +1,51 @@
...
+#include <gz/common/MeshLoader.hh>
+
+#include <gz/utils/ImplPtr.hh>
+
+namespace gz
+{
+  namespace common
+  {
+    /// \class AssimpLoader AssimpLoader.hh gz/common/AssimpLoader.hh
+    /// \brief Class used to load Assimp mesh files
```
chapulina, line 31:
```suggestion
    /// \brief Class used to load mesh files using the Assimp library.
```

graphics/src/CMakeLists.txt
```diff
@@ -27,6 +28,17 @@ gz_build_tests(
     gz-common${GZ_COMMON_VER}-testing
 )
 
+# Assimp doesn't offer preprocessor version, use cmake to set a compatibility
+# mode for versions below 5.2.0
+if(${GzAssimp_VERSION} STRLESS "5.2.0")
+  message("Warning, assimp below 5.2.0 detected, setting compatibility mode")
+  target_compile_definitions(${graphics_target} PRIVATE ASSIMP_COMPATIBILITY)
```
chapulina, line 35: The name `ASSIMP_COMPATIBILITY` isn't very informative. How about something like `GZ_ASSIMP_PRE_5_2_0`? That's following the pattern of [GZ_ZMQ_POST_4_3_1](https://github.com/gazebosim/gz-transport/blob/012a19dfb5ddbfe31411a909e18703b5a536b58a/include/gz/transport/Helpers.hh#L35).

graphics/src/AssimpLoader_TEST.cc
```diff
@@ -0,0 +1,607 @@
...
+
+  // check all vertices are unique
+  for (unsigned int i = 0; i < mesh->SubMeshCount(); ++i)
+  {
+    const std::shared_ptr<common::SubMesh> subMesh =
+      mesh->SubMeshByIndex(i).lock();
+    for (unsigned int j = 0; j < subMesh->VertexCount(); ++j)
+    {
+      gz::math::Vector3d v = subMesh->Vertex(j);
+      gz::math::Vector3d n = subMesh->Normal(j);
```
chapulina, line 146: Avoid using the `gz::` or `ignition::` namespaces when already using those namespaces, this helps reduce the diff between Fortress and Garden (and beyond)

chapulina:
We should make sure all the commits are signed for DCO before merging.

I've retriggered the macOS and Windows jobs now that https://github.com/gazebosim/gz-cmake/pull/278 is in. I'll retrigger the Linux ones after the `gz-cmake3` nightlies are done:

* Jammy [Build Status](https://build.osrfoundation.org/job/ign-cmake3-debbuilder/278/)
* Focal [Build Status](https://build.osrfoundation.org/job/ign-cmake3-debbuilder/279/)

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,722 @@
...
+  {
+    pbr.SetRoughness(value);
+  }
+#endif
+  mat->SetPbrMaterial(pbr);
+  return mat;
+}
+
+//////////////////////////////////////////////////
+std::pair<std::string, ImagePtr> AssimpLoader::Implementation::LoadTexture(
```
chapulina, line 448: The test coverage report is in. The entire file is 77.6% covered, which is not bad, but we're aiming for >90%. This function isn't covered at all, can we test it?

chapulina: The coverage report shows up better to me when I'm logged into codecov:

https://app.codecov.io/gh/gazebosim/gz-common/pull/393

graphics/include/gz/common/Material.hh
```diff
@@ -102,15 +103,22 @@ namespace gz
 
       /// \brief Set a texture image
       /// \param[in] _tex The name of the texture, which must be in the
-      /// resource path
-      public: void SetTextureImage(const std::string &_tex);
+      /// resource path or its name if _img is provided
+      /// \param[in] _img The image containing the texture if image has been
+      /// loaded in memory
+      public: void SetTextureImage(const std::string &_tex,
+                  const std::shared_ptr<Image> &_img = nullptr);
```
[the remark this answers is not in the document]
luca-della-vedova: The main reason is in https://github.com/gazebosim/gz-rendering/pull/642 under "other considerations". When working at the gz-rendering level I noticed that the image gets copied at least twice, once to convert between `gz::common::Image` and `gz::rendering::Image` and once in a gz::rendering::Material::Clone() function, this means that blobs of memory that for each memory that could be several megabytes we will need to do three allocations which might be not great from a performance point of view.
I guess it's a performance vs API clarity issue, I'm really on the fence on this, on one side it is good to avoid premature optimization and having a more obscure API on the other several multi MB copies every time a new model is loaded are not really negligible.

luca-della-vedova:
> We should make sure all the commits are signed for DCO before merging.

I'm having a lot of trouble with this since the history is so complex, most of the failing commits are by @onurtore, I tried an interactive rebase and adding his sign-off manually but after doing so I got failed DCO issues from a lot of the commits introduced by the regular main merges that are completely unrelated to this PR, i.e. below:

image

Not sure what to do here, unless I was to actually rewrite all the previous ones from other people that are now suddenly failing :man_shrugging:

graphics/src/AssimpLoader.cc
luca-della-vedova: I added a large test to get pretty much all the features tested other than the stray nullptr check that would require a fair bit of "creating invalid assets manually and make sure nothing goes badly".

Codecov still shows a large chunk of code not being tested, for example the whole `SplitMetallicRoughnessMap` function.
This function is private and is actually used and tested in a unit test that passes if the assimp version is above 5.0.0, specifically [here](https://github.com/gazebosim/gz-common/pull/393/files#diff-0fd2d29168662b4fdd425dcafa2a6d90a35ee01c90f533e4126fe509b4390a89R684-R687), however from the looks of the report it seems that codecov is running an older version of Assimp (maybe Ubuntu 20.04?) hence the whole chunk is `#ifndef`d away and not tested.
I could add an `#ifndef` to the function to remove dead code in Ubuntu 20.04 and increase the percentage of coverage but that would be pretty much only for the purpose of increasing coverage.

luca-della-vedova:
Windows is failing because `assimp` is not installed, I see that dependencies seem to be manually installed in the Windows CI, is there anything we should update?

chapulina:
> Windows is failing because assimp is not installed, I see that dependencies seem to be manually installed in the Windows CI, is there anything we should update?

* https://github.com/gazebo-tooling/release-tools/pull/774

graphics/src/AssimpLoader.cc
chapulina: Thanks, 88% is ok for now. We should look into either generating covarege from Jammy (moving `codecov_enabled` [here](https://github.com/gazebosim/gz-common/blob/main/.github/workflows/ci.yml)) or if it's something we're explicitly not testing, ignore it with `// LCOV_EXCL_START` and `// LCOV_EXCL_STOP`

onurtore:
> > We should make sure all the commits are signed for DCO before merging.
> 
> I'm having a lot of trouble with this since the history is so complex, most of the failing commits are by @onurtore, I tried an interactive rebase and adding his sign-off manually but after doing so I got failed DCO issues from a lot of the commits introduced by the regular main merges that are completely unrelated to this PR, i.e. below:
> 
> image
> 
> Not sure what to do here, unless I was to actually rewrite all the previous ones from other people that are now suddenly failing man_shrugging

I can modify commits without DCO using rebase and then force-push them. But this sounds problematic, any other ways?

luca-della-vedova:
> Assimp is now installed on Windows:
> CI log
> 
> But it's still not found:
> 
> ```
> -- Could NOT find GzAssimp (missing: GzAssimp_FOUND) 
> -- Looking for assimp - not found
> ```
> 
> Maybe something to be fixed at the `gz-cmake` level?

It ended up being a quite long debugging session, I tried to refactor the `gz-cmake` function in [a branch ](https://github.com/gazebosim/gz-cmake/blob/d1842c6bfbee8363114e6e7ae28a25fe92f24eeb/cmake/FindGzAssimp.cmake), this works locally on my 20.04 and Windows workstation, is there a way to tell CI to use that branch or do we have to necessarily merge and release? I don't have a Mac or 22.04 machine with me to test that it doesn't break behavior there

mjcarroll:
> It ended up being a quite long debugging session, I tried to refactor the gz-cmake function in [a branch ](https://github.com/gazebosim/gz-cmake/blob/d1842c6bfbee8363114e6e7ae28a25fe92f24eeb/cmake/FindGzAssimp.cmake), this works locally on my 20.04 and Windows workstation, is there a way to tell CI to use that branch or do we have to necessarily merge and release? I don't have a Mac or 22.04 machine with me to test that it doesn't break behavior there

You should be able to mostly copy what I did here to test branches: https://github.com/gazebosim/gz-common/pull/407

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,733 @@
...
+  }
+  else
+  {
+    ret.first = ToString(_texturePath);
+  }
+  return ret;
+}
+
+std::pair<ImagePtr, ImagePtr>
+    AssimpLoader::Implementation::SplitMetallicRoughnessMap(
```
mjcarroll, line 490: Nit: This appears to be completely uncovered by tests.  If it is unused, can we remove it, otherwise, I would appreciate a follow-up to correct that.

chapulina (approved):
LGTM. I'll second what @mjcarroll said about test coverage.

graphics/src/AssimpLoader.cc
luca-della-vedova: I think this is due to Codecov running on Ubuntu 20.04 rather than Ubuntu 22.04, specifically:

* The function is called [here](https://github.com/gazebosim/gz-common/blob/2c0773159e0299d0ed506b1a0f6ee541cb0bdd44/graphics/src/AssimpLoader.cc#L398), which is under an `#ifndef` block that doesn't get compiled in Assimp < 5.2.x (Ubuntu 20.04). The function populates metallic and roughness maps.
* The unit test for this functionality is [here](https://github.com/gazebosim/gz-common/blob/2c0773159e0299d0ed506b1a0f6ee541cb0bdd44/graphics/src/AssimpLoader_TEST.cc#L682-L686), I tested it locally on my Ubuntu 22.04 machine and indeed had to fix a few issues when I ran it, but again this test is `#ifndef`d away on Ubuntu 20.04 since the old assimp library doesn't support metallic roughness maps and it would fail. It runs on github actions for Jammy but not on Focal and is not picked up by codecov.

An example of an action from some time ago while I was iterating is [here](https://github.com/gazebosim/gz-common/actions/runs/2747512033), specifically Focal passes because the block is not compiled but Jammy fails [here](https://github.com/gazebosim/gz-common/runs/7542761772?check_suite_focus=true#step:4:4663) because of an issue that I fixed since.

If we want to increase coverage I can put the whole code block that is effectively dead in a `#ifndef` block as well, otherwise I don't know if there is a way (or it is desirable) to have codecov run on a newer version of Ubuntu.

That being said there is a small code path that is never tested, which is assets (such as gltf) that have an external texture, I'll add this in an upcoming PR.
This is specifically the path going through [this](https://app.codecov.io/gh/gazebosim/gz-common/compare/393/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%20comments&utm_term=gazebosim#D4L363) and [this](https://app.codecov.io/gh/gazebosim/gz-common/compare/393/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%20comments&utm_term=gazebosim#D4L484) line.

graphics/src/MeshManager.cc
```diff
@@ -97,9 +105,13 @@ MeshManager::MeshManager()
...
-  this->dataPtr->fileExtensions.push_back("stl");
-  this->dataPtr->fileExtensions.push_back("dae");
-  this->dataPtr->fileExtensions.push_back("obj");
+  this->dataPtr->fileExtensions.insert("stl");
+  this->dataPtr->fileExtensions.insert("dae");
+  this->dataPtr->fileExtensions.insert("obj");
+  this->dataPtr->fileExtensions.insert("gltf");
+  this->dataPtr->fileExtensions.insert("glb");
+  this->dataPtr->fileExtensions.insert("fbx");
+
```
onurtore, line 114: Hi @luca-della-vedova,
Should we remove this newline?

onurtore (commented):
Tested locally, looks good to me

onurtore:
Hi @luca-della-vedova,

The reason why actor.sdf seg faults is due two reason, one is solved in [this](https://github.com/gazebosim/gz-common/pull/411) PR, second is bvh animation in actor.sdf. Before bvh animation are added to skeleton in [here](https://github.com/gazebosim/gz-sim/blob/8c9489dd01f44cfe0a66d6aaea458025e87d8c48/src/rendering/SceneManager.cc#L2280) we are checking whether loaded bvh skeleton count are same with the skeleton loaded through mesh manager. If we use scene node in mesh manager, then those numbers are not same (The code is [here](https://github.com/gazebosim/gz-common/blob/1fed836b28d2e209c549695b2b7420e749beaaee/graphics/src/Skeleton.cc#L340)). Actually, even if we use assimp for loading bvh file, which I tried, then the rootNode will not be scene, bvh loader in assimp do not utilize scene node. 

So there are two multiple things we can do, if  [this](https://github.com/gazebosim/gz-common/pull/411) PR, and [this](https://github.com/gazebosim/gz-common/pull/410) PR are merged, then seg fault problem will dissappear. Otherwise, we have to modify the code during Bvhloading in SkeletonAnimation. I dont think that will be hard, and even we might deprecate the bvhloader and use assimp there as well.

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,733 @@
...
+  {
+    auto assimpMeshIdx = _node->mMeshes[meshIdx];
+    auto& assimpMesh = _scene->mMeshes[assimpMeshIdx];
+    auto nodeName = ToString(_node->mName);
+    auto subMesh = this->CreateSubMesh(assimpMesh, _transform);
+    subMesh.SetName(nodeName);
+    // Now add the bones to the skeleton
+    if (assimpMesh->HasBones() && _scene->HasAnimations())
+    {
+      // TODO(luca) merging skeletons here
```
onurtore, line 202: Hi @luca-della-vedova,
Can you eloborate on this TODO little bit more, I want to implement it, but I did not quite understand it.

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,733 @@
...
+      auto skeleton = _mesh->MeshSkeleton();
+      // TODO(luca) Append to existing skeleton if multiple submeshes?
+      skeleton->SetNumVertAttached(subMesh.VertexCount());
```
onurtore, line 205: Hi @luca-della-vedova,
Why we are setting this value in each iteration?

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,733 @@
...
+    subMesh.AddVertex(vertex);
+    subMesh.AddNormal(normal);
+    // Iterate over sets of texture coordinates
+    int uvIdx = 0;
+    while(_assimpMesh->HasTextureCoords(uvIdx))
+    {
+      math::Vector3d texcoords;
+      texcoords.X(_assimpMesh->mTextureCoords[uvIdx][vertexIdx].x);
+      texcoords.Y(_assimpMesh->mTextureCoords[uvIdx][vertexIdx].y);
+      // TODO(luca) why do we need 1.0 - Y?
```
onurtore, line 582: Looks like aiProcess_FlipUVs flag solves this issue, should we use that?

graphics/src/AssimpLoader.cc
```diff
@@ -0,0 +1,733 @@
...
+  ret = assimpMat->GetTexture(aiTextureType_DIFFUSE, 0, &texturePath);
+  // TODO(luca) check other arguments,
+  // type of mappings to be UV, uv index, blend mode
+  if (ret == AI_SUCCESS)
+  {
+    // Check if the texture is embedded or not
+    auto [texName, texData] = this->LoadTexture(_scene,
+        texturePath, this->GenerateTextureName(_scene, assimpMat, "Diffuse"));
+    if (texData != nullptr)
+      mat->SetTextureImage(texName, texData);
```
onurtore, line 361: Hi  @luca-della-vedova,
Do you know how we can combine multiple diffuse in here? I want to implement it but my knowledge is weak.

PR #401, opened 2022-07-25
Utilizes assimp flag in the mesh manager

onurtore:
Related to pr #393 and https://github.com/gazebosim/gz-sim/pull/1600

### DIFF
--- a/graphics/include/gz/common/MeshManager.hh
+++ b/graphics/include/gz/common/MeshManager.hh
@@ -240,6 +240,10 @@ namespace gz
                                const gz::math::Vector2d &_segments,
                                const gz::math::Vector2d &_uvTile);
 
+      /// \brief Set the assimp mode
+      /// \param[in] _useAssimp Set to true to enable assimp mode.
+      public: void SetUseAssimp(const bool _useAssimp);
+
       /// \brief Tesselate a 2D mesh
       ///
       /// Makes a zigzag pattern compatible with strips
--- a/graphics/src/MeshManager.cc
+++ b/graphics/src/MeshManager.cc
@@ -148,6 +148,7 @@ const Mesh *MeshManager::Load(const std::string &_filename)
     std::transform(extension.begin(), extension.end(),
         extension.begin(), ::tolower);
     MeshLoader *loader = nullptr;
+
     if (extension == "stl" || extension == "stlb" || extension == "stla")
       if (this->dataPtr->useAssimp)
         loader = &this->dataPtr->assimpLoader;
--- a/graphics/src/MeshManager.cc
+++ b/graphics/src/MeshManager.cc
@@ -148,7 +148,6 @@ const Mesh *MeshManager::Load(const std::string &_filename)
     std::transform(extension.begin(), extension.end(),
         extension.begin(), ::tolower);
     MeshLoader *loader = nullptr;
-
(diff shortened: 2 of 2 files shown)


onurtore:
@luca-della-vedova

graphics/include/gz/common/MeshManager.hh
```diff
@@ -240,6 +240,10 @@ namespace gz
                                const gz::math::Vector2d &_segments,
                                const gz::math::Vector2d &_uvTile);
 
+      /// \brief Set the assimp mode
+      /// \param[in] _useAssimp Set to true to enable assimp mode.
+      public: void SetUseAssimp(const bool _useAssimp);
```
ahcorde, line 245: More descriptive doc
```suggestion
      /// \brief Set who is going to read the mesh file. This only work for Collada and OBJ files
      /// If True ASSIMP will load the file, otherwise the internal loader will do it
      /// \param[in] _useAssimp Set to true to load the mesh with ASSIMP.
      public: void SetUseAssimp(const bool _useAssimp);
```

graphics/src/MeshManager.cc
```diff
@@ -70,6 +70,9 @@ class gz::common::MeshManager::Implementation
 
   /// \brief Mutex to protect the mesh map
   public: std::mutex mutex;
+
+  /// \brief True if assimp mode is enabled, false otherwise.
```
ahcorde, line 74: more descriptive

ahcorde (changes requested):

graphics/src/MeshManager.cc
```diff
@@ -146,20 +149,22 @@ const Mesh *MeshManager::Load(const std::string &_filename)
...
 
-    loader = &this->dataPtr->assimpLoader;
-    /*
     if (extension == "stl" || extension == "stlb" || extension == "stla")
-      loader = &this->dataPtr->stlLoader;
+      if (this->dataPtr->useAssimp)
+        loader = &this->dataPtr->assimpLoader;
+      else
+        loader = &this->dataPtr->stlLoader;
     else if (extension == "dae")
```
chapulina, line 157: Can we also use it for DAE and OBJ if the user requests so? This should make it easier to compare both loaders.

chapulina (commented):
Just a quick suggestion, it may be easier to expose this option through an environment variable, for example:

```cpp
      std::string useAssimpEnv;
      bool useAssimp{false};
      common::env("GZ_MESH_FORCE_ASSIMP", debugEnvEnv);
      if (useAssimpEnv == "true")
      {
        gzmsg << "Using assimp to load all mesh formats"  << std::endl;
        useAssimp = true;
      }

```

This way you don't need any changes in `gz-sim`. It also makes it easy to change it during runtime. For example:

1. Start `gz-sim`
2. Insert a COLLADA model (it will use our loader)
3. `export GZ_MESH_FORCE_ASSIMP=true`
4. Insert the same model again - it should be loaded with assimp now

graphics/src/MeshManager.cc
onurtore: @ahcorde, 
Should I close this PR and the PR related the server flag, and create a new one that uses environment variable? The ability the change the assimp flag during runtime sounds good.

onurtore:
I am closing this PR because now we are going to use environment variables set assimp mode.

PR #402, opened 2022-07-27
Add CSV parsing support

hidmic:
# 🎉 New feature

## Checklist
- [ ] Added tests

### DIFF
--- a/include/gz/common/DataFrame.hh
+++ b/include/gz/common/DataFrame.hh
@@ -1,0 +1,149 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#ifndef GZ_COMMON_DATAFRAME_HH_
+#define GZ_COMMON_DATAFRAME_HH_
+
+#include <array>
+#include <numeric>
+#include <sstream>
+#include <string>
+#include <stdexcept>
+#include <unordered_map>
+
+#include <gz/common/Io.hh>
+
+#include <gz/math/TimeVaryingVolumetricGrid.hh>
+
+namespace gz
+{
+  namespace common
+  {
+    template <typename K, typename V>
+    class DataFrame
+    {
+      public: bool Has(const K &_key) const
+      {
+        return this->storage.count(_key) > 0;
+      }
+
+      public: V &operator[](const K &_key)
+      {
+        return this->storage[_key];
+      }
+
+      public: const V &operator[](const K &_key) const
+      {
+        return this->storage.at(_key);
+      }
+
+      private: std::unordered_map<K, V> storage;
+    };
+
+    template <typename K, typename T, typename V>
+    struct IO<DataFrame<K, math::InMemoryTimeVaryingVolumetricGrid<T, V>>>
+    {
+      static DataFrame<K, math::InMemoryTimeVaryingVolumetricGrid<T, V>>
+      ReadFrom(const CSVFile &_file, const std::string &_timeColumn,
+               const std::array<std::string, 3> &_coordinateColumns)
+      {
+        const std::vector<std::string> &header = _file.Header();
+        if (header.empty())
+        {
+          throw std::invalid_argument(_file.Path() + " has no header");
+        }
+
+        auto it = std::find(header.begin(), header.end(), _timeColumn);
+        if (it == header.end())
+        {
+          std::stringstream sstream;
+          sstream << _file.Path() << " has no '"
+                  << _timeColumn << "' column";
+          throw std::invalid_argument(sstream.str());
+        }
+        const size_t timeIndex = it - header.begin();
+
+        std::array<size_t, 3> coordinateIndices;
+        for (size_t i = 0; i < _coordinateColumns.size(); ++i)
+        {
+          it = std::find(header.begin(), header.end(), _coordinateColumns[i]);
+          if (it == header.end())
+          {
+            std::stringstream sstream;
+            sstream << _file.Path() << " has no '"
+                    << _coordinateColumns[i] << "' column";
+            throw std::invalid_argument(sstream.str());
+          }
+          coordinateIndices[i] = it - header.begin();
+        }
+
+        return ReadFrom(_file, timeIndex, coordinateIndices);
+      }
+
+      static DataFrame<K, math::InMemoryTimeVaryingVolumetricGrid<T, V>>
+      ReadFrom(const CSVFile &_file, const size_t &_timeIndex = 0,
+               const std::array<size_t, 3> &_coordinateIndices = {1, 2, 3})
+      {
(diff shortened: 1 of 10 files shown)


chapulina:
Instead of maintaining a custom CSV parser, maybe we can depend on / vendor an existing one? This is the first one that came up on my search: https://github.com/d99kris/rapidcsv

hidmic:
> Instead of maintaining a custom CSV parser, maybe we can depend on / vendor an existing one?

@chapulina absolutely. There's no de-facto standard out there though, which is why I crafted a minimal implementation. 

> This is the first one that came up on my search: https://github.com/d99kris/rapidcsv

That one is header-only and source distributed, so we would have to vendor it. Code's simple enough.

hidmic:
@chapulina as per our discussion offline, I've updated this patch to rely on [`libcsv-dev`](https://packages.debian.org/sid/libdevel/libcsv-dev) for CSV parsing. I've also kept an iterator-based interface to it to avoid polluting the global namespace for `gz-common`. PTAL!

include/gz/common/DataFrame.hh
```diff
@@ -0,0 +1,202 @@
...
+#include <sstream>
+#include <string>
+#include <stdexcept>
+#include <unordered_map>
+#include <vector>
+
+#include <gz/common/CSVStreams.hh>
+#include <gz/common/Io.hh>
+
+#include <gz/math/TimeVaryingVolumetricGrid.hh>
```
scpeters, line 31: currently there are gz-common components that depend on gz-math, but the gz-common core library does not. We should consider moving the code in this PR to its own `csv` component or adding a math dependency to gz-common core

scpeters: this could also be added as a component of `gz-utils`

traversaro:
> @chapulina as per our discussion offline, I've updated this patch to rely on [`libcsv-dev`](https://packages.debian.org/sid/libdevel/libcsv-dev) for CSV parsing. I've also kept an iterator-based interface to it to avoid polluting the global namespace for `gz-common`. PTAL!


Totally ok for me, but just to double check: are you aware that this would add a LGPL dependency that cannot be skipped to `gz-common`?

chapulina:
> Totally ok for me, but just to double check: are you aware that this would add a LGPL dependency that cannot be skipped to gz-common?

Thanks for bringing it up, @traversaro , we hadn't noticed that. Still iterating.

chapulina:
Removed the `Breaking change` because the dependency on `libcsv` will be removed.

include/gz/common/DataFrame.hh
hidmic: Done in https://github.com/gazebosim/gz-common/pull/402/commits/84753af49393db913378e7aca02bf3f0d2778f61.

hidmic:
Alright, rolled back to a custom implementation in https://github.com/gazebosim/gz-common/pull/402/commits/150bb2657a5085a851cbd5883f446266af507233. Regex wasn't working that well so I went for a poor man's lexer/parser approach.

hidmic:
@scpeters @chapulina PTAL!

hidmic:
https://github.com/gazebosim/gz-math/pull/456 is merged. This should pass CI in some hours. 

@caguero mind to give it a pass?

chapulina:
Removed the `Garden` label because we're past feature freeze ❄️ .

Since this PR is backwards-compatible, it could be targeted at Citadel and merged forward from there.

io/include/gz/common/Io.hh
```diff
@@ -0,0 +1,60 @@
...
+#include <sstream>
+#include <string>
+
+namespace gz
+{
+  namespace common
+  {
+    template <typename T>
+    struct IO {
+      static T ReadFrom(std::istringstream &_istream)
```
caguero, line 29: Document these functions?

io/include/gz/common/CSVStreams.hh
```diff
@@ -0,0 +1,177 @@
...
+      char delimiter;
+
+      /// Row termination character.
+      char terminator;
+
+      /// Field quoting character.
+      char quote;
+
+      /// CSV dialect as expected by Unix tools.
+      static const CSVDialect Unix;
```
caguero, line 45: Does this parser work with Windows for example? If not I think we should mention it somewhere.

caguero:
Reminder: We'll need to update [gz-common5-release](https://github.com/gazebo-release/gz-common5-release/tree/main/focal/debian) with the new component. @j-rivero .

io/include/gz/common/CSVStreams.hh
hidmic: @caguero I'll start by saying I didn't try this on Windows, BUT my understanding is that Windows will translate `\r\n` to `\n` when reading a text file (see [here](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170#generic-text-routine-mappings)) so this dialect should work just fine there too.

io/include/gz/common/DataFrame.hh
```diff
@@ -0,0 +1,201 @@
...
+  {
+    /// \brief An abstract data frame.
+    ///
+    /// \tparam K Column key type
+    /// \tparam V Column value type
+    template <typename K, typename V>
+    class DataFrame
+    {
+      /// \brief Check if column key is present.
+      public: bool Has(const K &_key) const
```
caguero, line 45: Document `_key` and return value.

io/include/gz/common/DataFrame.hh
```diff
@@ -0,0 +1,201 @@
...
+      {
+        return this->storage.count(_key) > 0;
+      }
+
+      /// \brief Fetch mutable reference to column
+      public: V &operator[](const K &_key)
```
caguero, line 51: Document `_key` and return value.

io/include/gz/common/DataFrame.hh
```diff
@@ -0,0 +1,201 @@
...
+      {
+        return this->storage[_key];
+      }
+
+      /// \brief Fetch immutable reference to column
+      public: const V &operator[](const K &_key) const
```
caguero, line 57: Document `_key` and return value.

caguero (approved):
Minor documentation comments but looks good to me!

io/include/gz/common/DataFrame.hh
hidmic: Done in https://github.com/gazebosim/gz-common/pull/402/commits/9a3c84d6d1d6d76a62b35fd9ab04d31a0ca8178b.

io/include/gz/common/DataFrame.hh
hidmic: Done in https://github.com/gazebosim/gz-common/pull/402/commits/9a3c84d6d1d6d76a62b35fd9ab04d31a0ca8178b.

io/include/gz/common/DataFrame.hh
hidmic: Done in https://github.com/gazebosim/gz-common/pull/402/commits/9a3c84d6d1d6d76a62b35fd9ab04d31a0ca8178b.

hidmic:
@caguero the Homebrew PR job is failing for unrelated reasons, see [here](https://build.osrfoundation.org/job/ignition_common-ci-pr_any-homebrew-amd64/1742/clang/).

BTW how do we go about validating the Bazel BUILD file?

scpeters:
> @caguero the Homebrew PR job is failing for unrelated reasons, see [here](https://build.osrfoundation.org/job/ignition_common-ci-pr_any-homebrew-amd64/1742/clang/).

I believe that compiler warning has been fixed by https://github.com/gazebosim/gz-common/pull/416

mjcarroll:
> BTW how do we go about validating the Bazel BUILD file?

Don't worry about it for now, I have a batch of Bazel updates coming.

hidmic:
OK, I'll merge this into `main`, then backport to `mbari/gz-common5`. FYI @caguero.

PR #403, opened 2022-07-27
Introduces GZ_MESH_FORCE_ASSIMP flag

onurtore:
These flags utilized to use assimp for mesh loading process. If useAssimp flag is true then STL, GLTF and FBX meshes are loaded using assimp library, otherwise old loaders will be used. forceAssimp flags used to load all supported mesh formats using assimp.


onurtore:
@chapulina,
What do you think?

graphics/src/MeshManager.cc
```diff
@@ -145,21 +151,30 @@ const Mesh *MeshManager::Load(const std::string &_filename)
...
+    if (this->dataPtr->forceAssimp)
+    {
+      loader = &this->dataPtr->assimpLoader;
+    }
     else
     {
-      gzerr << "Unsupported mesh format for file[" << _filename << "]\n";
-      return nullptr;
+      if (extension == "stl" || extension == "stlb" || extension == "stla")
+        if (this->dataPtr->useAssimp)
```
chapulina, line 162: It looks like `useAssimp` is only used to enable its use for STLs. Can't we just rely on `forceAssimp` for that, and reduce the number of environment variables?

chapulina (changes requested):
Thanks! Could you document the new environment variables? Maybe on the `MeshManager`'s header?

graphics/src/MeshManager.cc
onurtore: Hi @chapulina,
Is there a way to document these variables in the MeshManager's header without moving them there?

onurtore: Is [this](https://github.com/gazebosim/gz-common/pull/403/commits/94bf4c04771205c1846bc5e6055023ca2ed4eeaa) commit what you taught?

chapulina (approved):
Looks great, thanks! Since we're not checking extensions when forcing assimp, it would be nice to add a test where `GZ_MESH_FORCE_ASSIMP=true` and we try to load a mesh format that isn't supported by assimp (i.e. `mesh.invalid`), to make sure that doesn't cause a crash.

graphics/src/MeshManager.cc
onurtore: [Here](https://github.com/gazebosim/gz-common/pull/403/commits/27b8a2b154c9a3e014831391198eb38961cbc63d)

chapulina: Great, thanks!

onurtore:
> Looks great, thanks! Since we're not checking extensions when forcing assimp, it would be nice to add a test where `GZ_MESH_FORCE_ASSIMP=true` and we try to load a mesh format that isn't supported by assimp (i.e. `mesh.invalid`), to make sure that doesn't cause a crash.

Hi @chapulina,
I think _IsValidFilename_ function in _MeshManager_ checks regardless of the mesh loader, and assimp loader covers the previous formats as well.  

On the other hand, I am very open to ideas about possible test cases, so if you have anything in mind, I would love to implement it.

chapulina:
I added a simple test in https://github.com/gazebosim/gz-common/pull/393/commits/a1c0b9fbbd42997d62524933eeed079c856e958f as part of https://github.com/gazebosim/gz-common/pull/393, that loads a non-supported extension and verifies that a `nullptr` is returned. That's all I had in mind 😄

PR #407, opened 2022-07-29
CI Configuration for new dependency in #402

mjcarroll:
Testing CI for #402 

For actions, we override the `dependencies.yaml` to temporarily use the gz-cmake PR: https://github.com/gazebosim/gz-cmake/pull/290

For macOS, we add a dependency on `libcsv` in the homebrew formula with the same `ci_matching_branch` branch name: https://github.com/osrf/homebrew-simulation/pull/1976

Additionally for macOS, I duplicated the https://github.com/gazebosim/gz-cmake/pull/290 branch with the same `ci_matching_branch` branch name.

### DIFF
--- a/.github/ci/dependencies.yaml
+++ b/.github/ci/dependencies.yaml
@@ -2,4 +2,9 @@ repositories:
   gz-cmake:
     type: git
     url: https://github.com/gazebosim/gz-cmake
-    version: hidmic/libcsv-cmake-module
+    version: ci_matching_branch/hidmic/csv-data
+  gz-math:
+    type: git
+    url: https://github.com/gazebosim/gz-math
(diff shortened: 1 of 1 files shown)


mjcarroll:
Looks like we are also depending on https://github.com/gazebosim/gz-math/pull/456, here.

```
  [ 19%] Building CXX object src/CMakeFiles/UNIT_DataFrame_TEST.dir/DataFrame_TEST.cc.o
  /github/workspace/src/DataFrame_TEST.cc:23:10: fatal error: gz/math/TimeVaryingVolumetricGrid.hh: No such file or directory
     23 | #include <gz/math/TimeVaryingVolumetricGrid.hh>
```

chapulina:
We decided not to depend on `libcsv` because it's LGPL.

PR #410, opened 2022-08-03
Hotfix for removing scene node

onurtore:
🦟 Bug fix

## Summary

Directly removing scene node from the skeleton creates problem for meshes without childs. This bugfix creates a dummy skeleton for this type of meshes. 

## Checklist
- [ ] Added tests


onurtore:
Hi @luca-della-vedova, 
I want to learn your opinion on this version. Should we directly move into new skeleton system of assimp, or should I focus on creating a better patch than this hotfix.

onurtore:
Hi @luca-della-vedova,

This PR results with failing MergeBoxWithDoubleSkeleton test. However, the test looks for the artifical 'Scene' node which is not originally should be there. 

I am looking for a mesh where there are two children right after Scene Node to represent this test, however I can not find such mesh. Even PatientWalkingCane.dae, which you mentioned have two skeletons, only have one child. So maybe we should delete MergeBoxWithDoubleSkeleton test? 

Secondly, this is currently a hotfix, I can spend time to create a better patch, or we can update our code to work with the new skeleton. What do you think is the best way?

luca-della-vedova:
Do you have any details of which type of meshes exhibit wrong behavior when the scene node is not removed and works when this patch is added?

luca-della-vedova:
I did a bit more investigation and I guess this goes back to the issues with actor.sdf segfaulting?

I wonder if instead of removing the root and hardcoding a "root is equal to the first child" logic we could just fix the scene node by assigning it a dummy animation that is just an identity (zero translation, unit quaternion) for the whole time, so children are preserved and their animations unchanged.

I'm not familiar enough with actor animations and exporting software to know whether the hardcoding 0 would have any side effects or not but it feels a bit dangerous if a root scene node was to have more than one child for any reason.

onurtore:
> I did a bit more investigation and I guess this goes back to the issues with actor.sdf segfaulting?
> 
> I wonder if instead of removing the root and hardcoding a "root is equal to the first child" logic we could just fix the scene node by assigning it a dummy animation that is just an identity (zero translation, unit quaternion) for the whole time, so children are preserved and their animations unchanged.
> 
> I'm not familiar enough with actor animations and exporting software to know whether the hardcoding 0 would have any side effects or not but it feels a bit dangerous if a root scene node was to have more than one child for any reason.

Yes it is related to actor.sdf.

This idea might work if [this](https://github.com/gazebosim/gz-sim/pull/1655) is also merged.

onurtore:
@luca-della-vedova,
Something like this maybe: #429

onurtore:
Hi @luca-della-vedova, 
I have implemented the dummy animation as well ([here](https://github.com/gazebosim/gz-common/pull/429)), but I dont think thats a good approach. So I am still voting for removing the auxilary 'scene' node from the system.

azeey:
@luca-della-vedova can we close this PR? There hasn't been any progress in a long time.

PR #411, opened 2022-08-04
Implement XDisplacement function for checking x displacement for all nodes

onurtore:
# 🦟 Bug fix

Related to #393 

## Summary

During actor's trajectory creation, createActor function only looks for the root node's x displacement while setting the trajectory interpolation flag. 
If root node does not have any animation, this results with a segmentation fault. While with the previous mesh loaders, this seg fault never happened, recent introduction of assimp loader create cases where root node does not have animation and results with segmentation fault during actor creation.

This patch introduces new function called _XDisplacement_ to check all nodes inside the skelAnim for possible X displacement. This function will be utilized during the actor's trajectory creation.

## Checklist
- [ ] Added tests


ahcorde (changes requested):
add some test

graphics/src/AssimpLoader_TEST.cc
```diff
@@ -697,3 +697,20 @@ TEST_F(AssimpLoader, LoadGlbPbrAsset)
...
+  common::AssimpLoader loader;
+  common::Mesh *mesh = loader.Load(common::testing::TestFile("data",
+        "walk.dae"));
+  auto meshSkel =  mesh->MeshSkeleton();
+  std::string rootNodeName = meshSkel->RootNode()->Name();
+  common::SkeletonAnimation *skelAnim = meshSkel->Animation(0);
+  common::NodeAnimation *rootNode = skelAnim->NodeAnimationByName(rootNodeName);
+  EXPECT_EQ(nullptr, rootNode);
+  auto xDisplacement = skelAnim->XDisplacement();
+  ASSERT_TRUE(xDisplacement);
```
ahcorde, line 715: can we add another example below with that return xDisplacement=false?

ahcorde (changes requested):

graphics/src/AssimpLoader_TEST.cc
onurtore: This might take a while, I need to understand how to remove animations from collada.

onurtore: @ahcorde,
Can I create just an empty animation and test the function with that?

ahcorde: yes,

onurtore: Done, added new test file for skeletonAnimation because we are not using any mesh for the new test.

onurtore: @ahcorde ,
One question, I added new test file and re-compile everything using colcon, when I tried to make test, make didnt see any test, I had to remove gz-common build and re-compile everything. Took me 30min, is there a easier way? Refreshing cache for test?

graphics/src/SkeletonAnimation_TEST.cc
```diff
@@ -0,0 +1,35 @@
...
+using namespace gz;
+class SkeletonAnimation : public common::testing::AutoLogFixture { };
+
+/////////////////////////////////////////////////
+TEST_F(SkeletonAnimation, CheckNoXDisplacement)
+{
+  common::SkeletonAnimation* skelAnim =
+    new common::SkeletonAnimation("emptyAnimation");
+  auto xDisplacement = skelAnim->XDisplacement();
+  ASSERT_TRUE(!xDisplacement);
```
ahcorde, line 34:
```suggestion
  ASSERT_FALSE(xDisplacement);
```

ahcorde (changes requested):

graphics/src/SkeletonAnimation_TEST.cc
onurtore: Done.

ahcorde:
@osrf-jenkins run tests please!

ahcorde (approved):

ahcorde:
Closed and reopened again, Github actions were stuck

ahcorde:
@j-rivero @mjcarroll something is broken in the ABI checker

j-rivero:
> @j-rivero @mjcarroll something is broken in the ABI checker

Log of the ABI checker is:
```
...
ERROR: some errors occurred when compiling headers
ERROR: see log for details:
  /home/jenkins/workspace/ignition_common-abichecker-any_to_any-ubuntu_auto-amd64/abi_checker/logs/ign-common/X/log.txt

ERROR: can't compile header(s)
```

Note that the file described there is available via Jenkins workspace at https://build.osrfoundation.org/job/ignition_common-abichecker-any_to_any-ubuntu_auto-amd64/ws/abi_checker/logs/ign-common/X/log.txt/*view*/

```
The GCC parameters:
  gcc -fdump-lang-raw -fkeep-inline-functions -c -x c++ -fpermissive -w  -std=c++17 "/tmp/lCfTAc7aXw/dump1.h"  -I/usr/local/destination_branch/include/gz/common5 -I/usr/include/gz/utils2 -I/usr/include/gz/math7

In file included from /usr/local/destination_branch/include/gz/common5/gz/common/testing.hh:23,
                 from /tmp/lCfTAc7aXw/dump1.h:83:
/usr/local/destination_branch/include/gz/common5/gz/common/testing/AutoLogFixture.hh:20:10: fatal error: gtest/gtest.h: No such file or directory
   20 | #include <gtest/gtest.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```

Something seems to have broken the compilation in the `gz-common5` branch. Would you mind creating an issue for it please?

luca-della-vedova:
Did you try to use this function on an actual animated model that has no X displacement (i.e. [this](https://app.gazebosim.org/OpenRobotics/fuel/models/MaleVisitorPhone))?
My concern is that this change will return much more positives. For example if there is even one node that has a pose which has a X in its last keyframe that is slightly different from the first keyframe (i.e. the hand bone for a person looking at their phone while being on the spot) this function might return `true` while previously we would have considered that `false` since the root was not moving.

onurtore:
@luca-della-vedova,
What is the alternative that will works with assimp?  Accesing second node? And what if the root node does have two child node? Then which one we should we use?

onurtore:
Also can you share a code snippet that breaks the animation due to wrong interpolation flag (if a root node doesnt have positive x , and non root node have positive x displacement.)

onurtore:
@luca-della-vedova,
Is there a specific reason we only check for root node and if there is, what is the reason.

onurtore:
Hi @iche033,
I think you wrote the code for checking x interpolation for animations during actor creations. The current code checks only for root node, is there a specific reason for that?

PR #414, opened 2022-08-09
ign -> gz : Remove redundant namespace references

methylDragon:
Part of: https://github.com/gazebo-tooling/release-tools/issues/784

### DIFF
--- a/src/Console_TEST.cc
+++ b/src/Console_TEST.cc
@@ -40,12 +40,12 @@ class Console_TEST : public ::testing::Test {
   /// \brief Clear out all the directories we produced during this test.
   public: virtual ~Console_TEST()
   {
-    EXPECT_TRUE(ignition::common::unsetenv(IGN_HOMEDIR));
+    EXPECT_TRUE(common::unsetenv(IGN_HOMEDIR));
 
-    if (ignition::common::isDirectory(this->logBasePath))
+    if (common::isDirectory(this->logBasePath))
     {
       ignLogClose();
-      EXPECT_TRUE(ignition::common::removeAll(this->logBasePath));
+      EXPECT_TRUE(common::removeAll(this->logBasePath));
     }
   }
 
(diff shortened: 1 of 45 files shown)


src/PluginUtils_TEST.cc
```diff
@@ -30,7 +30,7 @@ TEST(PluginUtils, NormalizeName)
   EXPECT_EQ("::ignition", common::NormalizeName("ignition"));
   EXPECT_EQ("::ignition", common::NormalizeName("::ignition"));
 
-  EXPECT_EQ("::ignition::math", common::NormalizeName("ignition::math"));
+  EXPECT_EQ("::ignition::math", common::NormalizeName("math"));
```
chapulina, line 33: I think this should stay as it is

chapulina (changes requested):
Compilation is failing, we need to make sure the namespace is used where we remove it:

```
create_and_switch_to_temp_dir(std::__cxx11::string&)':
  /github/workspace/src/Filesystem_TEST.cc:38:8: error: 'env' was not declared in this scope
     if (!env("TMPDIR", tmppath))
```

src/PluginUtils_TEST.cc
methylDragon: I've edited the script now to exclude any namespace cleanups occurring with a leading `"`, so this has been reverted!

I'm not sure what's up with the compilation error, I'll get to it if CI explodes again

src/Console_TEST.cc
```diff
@@ -40,12 +40,12 @@ class Console_TEST : public ::testing::Test {
   /// \brief Clear out all the directories we produced during this test.
   public: virtual ~Console_TEST()
   {
-    EXPECT_TRUE(ignition::common::unsetenv(IGN_HOMEDIR));
+    EXPECT_TRUE(unsetenv(IGN_HOMEDIR));
```
chapulina, line 43: This is one of the places where removing the namespace is actually changing what function is called. I believe the tests are failing because this is calling [this](https://linux.die.net/man/3/unsetenv) instead of our cross-platform function.

chapulina (changes requested):

src/Console_TEST.cc
methylDragon: Hmm.. Let me add the following excludes then:
- `env`
- `unsetenv`
- `joinPaths`
- `removeAll`
- `uuid`
- `isDirectory`

This will result in just `ignition::` being removed, where applicable

methylDragon: Tests pass locally for me

methylDragon:
Tests now pass, but with a deprecation warning on homebrew (that I think is unrelated?)

chapulina:
> deprecation warning on homebrew (that I think is unrelated?)

* https://github.com/gazebosim/gz-common/issues/415

It looks like @mjcarroll is on it

chapulina (approved):

PR #424, opened 2022-08-17
Introduces remove scene node env var, renames func 

onurtore:
# 🦟 Bug fix

Fixes #393

## Summary

Assimp creates additional scene node for the skeleton, this commit introduces a flag for removing that. Users also
allowed to set this flag through environment variable. Furthermore renames a previous func to fix its naming.

This commit does not implement the actual operation but prepares the necessary env var. The actual base implementation already inside #410, and will move from there.

## Checklist
- [ ] Added tests


graphics/include/gz/common/MeshManager.hh
```diff
@@ -247,7 +247,7 @@ namespace gz
       /// \brief Sets the forceAssimp flag by reading the GZ_MESH_FORCE_ASSIMP
       /// environment variable. If forceAssimp true, MeshManager uses Assimp
       /// for loading all mesh formats, otherwise only for GLTF and FBX.
-      public: void SetAssimpEnvs();
+      public: void SetAssimpEnv();
```
ahcorde, line 250: I recommend not changing the signature of this method, otherwise we need to tick-tock the method name to deprecate this one and create the new one.

onurtore: Okay, updating them.

onurtore: Done.

mjcarroll:
@onurtore and @luca-della-vedova what is the status of this (and the other few Assimp-related PRs?

onurtore:
Hi @mjcarroll,
There is one major integration problem, and the whole process kinda stuck. The problem arises as  segmentation fault during loading of certain sdf files with actors (One example is actor.sdf) with Assimp.  This happens because:

Current integration of the Assimp library creates additional root node for the Gazebo skeleton. This root node generated by Assimp and can not be found in the file that user trying to load and creates segmentation faults during loading of actor.sdf. I will refer to this node as root node, and the actual root node that is the root node defined in the file as actual root node.

Segmentation fault happens during the XDisplacement check in src/rendering/SceneManager.cc during actor creation. Related code snippet is [here](https://github.com/gazebosim/gz-sim/blob/2d19f9312c87752f50a6b9792bccc46519ed940d/src/rendering/SceneManager.cc#L2383).  What happens is, Gazebo tries to access rootNode animation, however the rootNode does not have any animation because it is not the actual root node, it is a dummy root node generated by assimp.

To solve this, I suggest to remove this node by just assigning root node to it is child (PR [here](https://github.com/gazebosim/gz-common/pull/410). This is I am aware that sound dangrerous, as it stated by @luca-della-vedova below: 

> I wonder if instead of removing the root and hardcoding a "root is equal to the first child" logic we could just fix the scene node by assigning it a dummy animation that is just an identity (zero translation, unit quaternion) for the whole time, so children are preserved and their animations unchanged.
> 
> I'm not familiar enough with actor animations and exporting software to know whether the hardcoding 0 would have any side effects or not but it feels a bit dangerous if a root scene node was to have more than one child for any reason.
> 

Luca suggest adding dummy animation because if root node have more than one actual root node assigning root to one of its actual root node might be problematic as we disregard some critical information. In that case, dummy animation needs to have the actual roots node properties, otherwise XDisplacement check condition inside the actor creation will be wrong. However, if there is a possibility of more than one child, then which child we are going to use for the dummy animation, that is a question...

As an alternative solution, I created a PR where we check not just the root node, but all the skeleton nodes for XDisplacement ([PR](https://github.com/gazebosim/gz-common/pull/411)), This way we can sure whether there is XDisplacement or not by checking all the nodes.  However, this might increase the number of positives as it stated by @Luca:

> Did you try to use [this](https://app.gazebosim.org/OpenRobotics/fuel/models/MaleVisitorPhone) function on an actual animated model that has no X displacement (i.e. this)?
> My concern is that this change will return much more positives. For example if there is even one node that has a pose which has a X in its last keyframe that is slightly different from the first keyframe (i.e. the hand bone for a person looking at their phone while being on the spot) this function might return true while previously we would have considered that false since the root was not moving.

So with this new information, I do not have any alternative. I suggest to assign the root node to its child, because I do not see any cases where dummy root node have more than one real child. However creating a PR for this functionality will make at least one test fails because  we created a specific test for checking dummy root node. And no one accepts former intern's PR with failed test cases :D.

onurtore:
Hi @mjcarroll,
Friendly ping. An answer would be really good.

azeey:
@luca-della-vedova after reading your comment, my understanding is that we should close this PR and keep using the old Collada loader for all dae files. Is that the case? If so, do we need to revert/modify https://github.com/gazebosim/gz-common/pull/393?

luca-della-vedova:
> @luca-della-vedova after reading your comment, my understanding is that we should close this PR and keep using the old Collada loader for all dae files. Is that the case? If so, do we need to revert/modify #393?

Hi!

Yea agree with the approach. #393 does not override the previous behavior and we still use the collada loader for all dae files so there is no need to revert anything. I'll close this then

PR #427, opened 2022-08-19
Fix ABI checker with testing module

mjcarroll:

# 🦟 Bug fix

Fixes issue brought up in comment here: https://github.com/gazebosim/gz-common/pull/411#issuecomment-1220892591

## Summary

The header `AutoLogFixture.hh` is available to be used as part of a library or executable that links against `gtest`.  We don't want to introduce a hard dependency on `gtest` in `common::testing` for just this header, but it seems to annoy the ABI checker.

~~This adds a private include so that the gtest header may be resolved, even though downstream users of the `testing` component will never see it.~~

This puts the AutoLogFixture within a define guard `GTEST_API` to guarantee that gtest is available before using the autologfixture.

## Checklist
- [ ] Added tests

### DIFF
--- a/testing/include/gz/common/testing/AutoLogFixture.hh
+++ b/testing/include/gz/common/testing/AutoLogFixture.hh
@@ -1,5 +1,5 @@
 /*
-* Copyright (C) 2022 Open Source Robotics Foundation
+* Copyright (C) 2022 Open Source Robotics FoundationO
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -17,7 +17,8 @@
 #ifndef GZ_COMMON_TESTING_AUTOLOGFIXTURE_HH_
 #define GZ_COMMON_TESTING_AUTOLOGFIXTURE_HH_
 
-#include <gtest/gtest.h>
+/// Protect to guarantee that gtest is included before this header.
+#ifdef GTEST_API_
 
 #include <memory>
 #include <string>
@@ -55,5 +56,9 @@ class AutoLogFixture : public ::testing::Test
 }  // namespace gz::common::testing
 
 #include <gz/common/testing/detail/AutoLogFixture.hh>
+
+#else
+#warning "AutoLogFixture needs <gtest/gtest.h> to be included in order to work"
+#endif  // GTEST_API_
 
 #endif  // GZ_COMMON_TESTING_AUTOLOGFIXTURE_HH_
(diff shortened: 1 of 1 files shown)


mjcarroll:
@j-rivero so it seems I can't manipulate the ABI checker via CMake. 

The situation here is that the `AutoLogFixture.hh` header is intended to be used only by people who already have linkage/includes against `gtest`.

Is there a way to exclude this from the ABI checker?  Or is there a better way to indicate this?

testing/src/CMakeLists.txt
```diff
@@ -13,5 +13,9 @@ set(test_sources
 
 gz_add_component(testing SOURCES ${sources} GET_TARGET_NAME testing_target)
 
+# In general, this shouldn't be necessary as the downstream test is pulling in the 
+# gtest includes.  We do this here to keep the ABI checker happy
```
ahcorde, line 17:
```suggestion
# gtest includes. We do this here to keep the ABI checker happy
```

mjcarroll: I ended up removing the CMakeLists part, as it has no impact on the ABI checker.

mjcarroll:

testing/include/gz/common/testing/AutoLogFixture.hh
```diff
@@ -1,5 +1,5 @@
 /*
-* Copyright (C) 2022 Open Source Robotics Foundation
+* Copyright (C) 2022 Open Source Robotics FoundationO
```
jennuine, line 2:
```suggestion
* Copyright (C) 2022 Open Source Robotics Foundation
```

mjcarroll:
The abi checker is actually having issues with the _destination_ branch in this case.  This PR would need to be merged so that the destination branch (`gz-common5`) is fixed.  I would expect the ABI checker to fail on this PR but succeed on future PRs.

mjcarroll:
The error in question:

```
The GCC parameters:
  gcc -fdump-lang-raw -fkeep-inline-functions -c -x c++ -fpermissive -w  -std=c++17 "/tmp/4wmT6ajeTj/dump1.h"  -I/usr/local/destination_branch/include/gz/common5 -I/usr/include/gz/utils2 -I/usr/include/gz/math7

In file included from /usr/local/destination_branch/include/gz/common5/gz/common/testing.hh:23,
                 from /tmp/4wmT6ajeTj/dump1.h:83:
/usr/local/destination_branch/include/gz/common5/gz/common/testing/AutoLogFixture.hh:20:10: fatal error: gtest/gtest.h: No such file or directory
   20 | #include <gtest/gtest.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```

chapulina (approved):

PR #429, opened 2022-08-22
Adds dummy animation for the root scene node

onurtore:
# 🦟 Bug fix

Fixes #393 

## Summary
Adds dummy animation for the root scene.

## Checklist
- [ ] Added tests


azeey:
@luca-della-vedova would you be able to give this a look?

luca-della-vedova:
I can't remember the exact details of this PR but iirc it was to fix a corner case when parsing one specific collada actor with the Assimp loader, which in the end is also hidden behind a custom environment variable, we can revisit these corner cases if we were to remove our custom parser and just use assimp for every mesh, otherwise this is extremely niche and not sure about the implications on other models

Issue #434, opened 2022-09-02
DataFrame picks up spurious all zeros row


hidmic:
## Environment
* OS Version: Ubuntu 20.04
* Source build, on [`main`](https://github.com/gazebosim/gz-common/tree/main) (@ https://github.com/gazebosim/gz-common/commit/d847f39ab8fb272d47b4208e15a6e0531579b4da)
  
  

## Description
* Expected behavior: reading `gz::common::DataFrame` instances from CSV streams picks up all file data.   
* Actual behavior: reading `gz::common::DataFrame` instances from CSV streams picks up an extra row at the beginning, filled with zeros.

azeey:
Fixed by #435.

PR #435, opened 2022-09-02
Skip CSV header when reading DataFrame

hidmic:
# 🦟 Bug fix

Fixes https://github.com/gazebosim/gz-common/issues/434

## Summary

This patch fixes a minor issue introduced in #402. Namely, header row shouldn't be read in as numeric data. It also updates tests to identify this regression.

## Checklist
- [x] Added tests
- [x] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))

### DIFF
--- a/io/include/gz/common/DataFrame.hh
+++ b/io/include/gz/common/DataFrame.hh
@@ -179,7 +179,8 @@ namespace gz
         using FactoryT =
             math::InMemoryTimeVaryingVolumetricGridFactory<T, V>;
         std::vector<FactoryT> factories(dataColumnIndices.size());
-        for (auto it = _begin; it != _end; ++it)
+        auto it = _begin;
+        while (++it != _end)
         {
           const T time = IO<T>::ReadFrom(it->at(_timeColumnIndex));
           const math::Vector3<P> position{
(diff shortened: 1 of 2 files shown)


io/include/gz/common/DataFrame.hh
```diff
@@ -179,7 +179,8 @@ namespace gz
         using FactoryT =
             math::InMemoryTimeVaryingVolumetricGridFactory<T, V>;
         std::vector<FactoryT> factories(dataColumnIndices.size());
-        for (auto it = _begin; it != _end; ++it)
+        auto it = _begin;
+        while (++it != _end)
```
arjo129, line 183: Knit: This works although it isn't immediately obvious what's going on. It may be cleaner to use `for (auto it = std::next(_begin); it != _end; ++it)` since we already check if `_begin` is valid before.

arjo129 (approved):
Works for me.

io/include/gz/common/DataFrame.hh
hidmic: @arjo129 that's way cleaner. Applied in https://github.com/gazebosim/gz-common/pull/435/commits/b478475964af847ef04aa6920883170051ca3974.

hidmic:
Alright, PR approved, CI's happy. Going in.

PR #446, opened 2022-10-04
Adds an API to retrieve keys.

arjo129:
This is useful for when we perform visualization.

Signed-off-by: Arjo Chakravarty <arjo@openrobotics.org>


ahcorde (changes requested):
Can you add a test?

Issue #470, opened 2022-10-26
gz-common io - include nested too deeply


hidmic:
## Environment
* OS Version: Mac OS, Windows
* Built from [`gz-common5`](56b82ad501b64a0c51b06a79fd68aeec2766f86e) source  

## Description
* Expected behavior: downstream builds against `gz-common5` succeed
* Actual behavior: downstream builds against `gz-common5` fail with: 

  ```sh
  /usr/local/include/gz/common5/gz/common/Io.hh:24:10: error: #include nested too deeply
  ```

  E.g. https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-homebrew-amd64/9405, https://build.osrfoundation.org/job/ign_gazebo-pr-win/5169.

## Steps to reproduce

1. Check https://github.com/gazebosim/gz-sim/pull/1616 (first `gz-common::io` component use).

mjcarroll:
Fixed by #471

PR #471, opened 2022-10-26
Avoid Io.hh header name clash

hidmic:
# 🦟 Bug fix

Fixes https://github.com/gazebosim/gz-common/issues/470.

## Summary

More of a hunch than a fix. I suspect neither Mac OS nor Windows do case sensitive header look up, which creates a clash between the `Io.hh` file and the generated `io.hh` file.

## Checklist
- [ ] Added tests


mjcarroll:
Yep, this is very likely the fix.

mjcarroll (approved):

hidmic:
Would you mind releasing this @mjcarroll ?

mjcarroll:
Ah yeah, in order for it to be released, we need to target something other than `main`.  Can this land on garden?

scpeters:
an alternative is to nest the headers of components in a subfolder like `gz/common/io/Io.hh`

hidmic:
There was no fundamental reason to name that header `Io.hh`, so no biggie.

PR #474, opened 2022-10-30
Expose Vertex & Index raw ptrs for efficient reading

darksylinc:
# 🎉 New feature

## Summary

This PR exposes vertex & index raw ptrs.

It is necessary to _significantly_ improve performance of Ogre2RayQuery.

It is up to caller to ensure this unsafe pointers are read properly and hence its usage is discouraged; only when necessary.

## Checklist
- [x] Added tests

### DIFF
--- a/graphics/include/gz/common/SubMesh.hh
+++ b/graphics/include/gz/common/SubMesh.hh
@@ -157,6 +157,12 @@ namespace gz
       /// \sa bool HasVertex(const unsigned int) const
       public: gz::math::Vector3d Vertex(const unsigned int _index) const;
 
+      /// \brief Get the raw vertex pointer. This is unsafe, it is the
+      /// caller's responsability to ensure it's not indexed out of bounds.
+      /// The valid range is [0; VertexCount())
+      /// \return Raw vertices
+      public: const gz::math::Vector3d* VertexPtr() const;
+
       /// \brief Set a vertex
       /// \param[in] _index Index of the vertex
       /// \param[in] _v The new vertex coordinate
@@ -216,6 +222,12 @@ namespace gz
       /// \param[in] _index Array index.
       /// \return The index, or -1 if the _index is out of bounds.
       public: int Index(const unsigned int _index) const;
+
+      /// \brief Get the raw index pointer. This is unsafe, it is the
+      /// caller's responsability to ensure it's not indexed out of bounds.
+      /// The valid range is [0; IndexCount())
+      /// \return Raw indices
+      public: const unsigned int* IndexPtr() const;
 
       /// \brief Set an index
       /// \param[in] _index Index of the indices
--- a/graphics/src/Mesh_TEST.cc
+++ b/graphics/src/Mesh_TEST.cc
@@ -95,11 +95,13 @@ TEST_F(MeshTest, Mesh)
   EXPECT_EQ(submesh.lock()->Vertex(0), v0 * scale);
   EXPECT_EQ(submesh.lock()->Normal(0), n0);
   EXPECT_EQ(submesh.lock()->TexCoord(0), uv0);
+  EXPECT_EQ(submesh.lock()->VertexPtr()[0], v0 * scale);
 
   mesh->Scale(scale);
   EXPECT_EQ(submesh.lock()->Vertex(0), v0 * scale * scale);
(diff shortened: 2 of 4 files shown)


graphics/include/gz/common/SubMesh.hh
```diff
@@ -157,6 +157,12 @@ namespace gz
       /// \sa bool HasVertex(const unsigned int) const
       public: gz::math::Vector3d Vertex(const unsigned int _index) const;
 
+      /// \brief Get the raw vertex pointer. This is unsafe, it is the
+      /// caller's responsability to ensure it's not indexed out of bounds.
+      /// The valid range is [0; VertexCount())
+      /// \return Raw vertices
+      public: const gz::math::Vector3d* VertexPtr() const;
```
mjcarroll, line 164: Is the intention for the consumer to be able to modify the contents?  I think that would break `const` correctness in this case.

I think we should either use `const gz::math::Vector3d const *` to indicate that both the pointer and the pointed-to data are const OR make this a non-const function as it is returning mutable data.

darksylinc: > Is the intention for the consumer to be able to modify the contents?

No. The specific use case this function is needed is to implement efficient raycasts (which only needs reading).

But even if we consider other use cases, exposing the pointers non-const precludes gazebo from sanitizing those writes (e.g. prevent writing an index that addresses an out-of-bounds to vertex).

Following [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), if in the future a case of vertex data manipulation arises that needs write access (that can't be moved inside SubMesh class), we can evaluate it then.

mjcarroll: That's what I was thinking. In that case, lets do const pointer to const to make things safer/sanitizable.

darksylinc: OK I was confused about this. But I wanted to confirm with the compiler.

The following are valid (I use printf to avoid invalid usage warning):

```cpp
#include "stdio.h"

class Foo
{
public:
	int const *test0() const;
	const int *test1() const;
};

int main()
{
	Foo f;

	const int *ta0 = f.test0();
	int const *tb0 = f.test0();
	const int *const tc0 = f.test0();

	const int *ta1 = f.test1();
	int const *tb1 = f.test1();
	const int *const tc1 = f.test1();

	printf( "%i,%i,%i,%i,%i,%i", *ta0, *tb0, *tc0, *ta1, *tb1, *tc1 );

	return 0;
}
```

The following raise warnings:

```cpp
class Foo
{
public:
	const int const *test0() const; // warning: duplicate 'const' declaration specifier [-Wduplicate-decl-specifier]
	const int *const test1() const; // warning: 'const' type qualifier on return type has no effect
};
```

Obviously when it comes to variables `const int *ta0` is not the same as `int const *tb0`.

 - tb0 can later be changed to point to another address, but is not allowed to change the contents of the addresses it holds.
 - ta0 cannot be changed to point to any other address, and it cannot change the contents of the address either.

However when it comes to return values; these are the same and writing `const int const *` raises warnings.

So which version do you prefer for **return** values? `const int*` or `int const *` ? Both do exactly the same in C++.

If the pointers were of type `int **` then that'd be a different story.

mjcarroll: Ah, my mistake, I didn't realize that it had no effect in return values like this.  I had to confirm for myself.  I think that my preference is for `const int*` for most clarity, which is what you have here.

mjcarroll:
@darksylinc I think at this point, if you want to retarget to `gz-common5` as @scpeters suggested, we should be good to go here.

iche033:
changes look good. Maybe add tests for the new APIs in [SubMesh_TEST.cc](https://github.com/gazebosim/gz-common/blob/gz-common5/graphics/src/SubMesh_TEST.cc)

darksylinc:
- Rebased and retargeted the PR to gz-common5
 - Added requested unit tests

darksylinc:
The submesh test if failing on macOS. Anyone has any idea why?

The detailed info doesn't appear.

darksylinc:
Test fixed. All tests sre green.

mjcarroll (approved):

PR #478, opened 2022-11-15
Fixed animation with wierd animation due bad times

ahcorde:
Signed-off-by: ahcorde <ahcorde@gmail.com>

# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
Sometimes the assimp rotation value returns a invalid `mTime`. This lines should check this

wrong behaviour
curtains_bad

Fixed
curtains

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #479, opened 2022-11-16
Fix out of bounds access to rotation array in assimp loader

luca-della-vedova:
# 🦟 Bug fix

Alternative to #478, it fixes the root cause of the issue that is undefined behavior due to access past the end of the array.

## Summary

When the parsing function was written, the assumption was that the position keys array and the rotation keys array would be the same length, this proved to not be the case since they can have different lengths.
The root cause of the issue addressed through #478 was that the rotation key array only had one element while the position key array was much longer, hence we were accessing values past the end of the array which is undefined behavior and can cause all sorts of trouble to downstream users.
Other cases with other animations can cause NaN exceptions in OGRE2 that crash the whole application.

With this PR we iterate over the longest of the arrays (to make sure no data is lost) and set the access to the position / rotation keys to be the minimum between the current index and the length of the array, to avoid undefined behavior.
This means that the last key will be duplicated in case the array is shorter, but that is acceptable behavior. It seems assimp returns a rotation array of length 1 to denote that rotation does not change throughout the animation.

The first one-line commit fixes the issue we are personally observing but I added https://github.com/gazebosim/gz-common/commit/f2cb2ce011b75966849b5b946c39bfa882682820 to fix the symmetrical case

## Checklist
- [ ] Added tests

### DIFF
--- a/graphics/src/AssimpLoader.cc
+++ b/graphics/src/AssimpLoader.cc
@@ -683,12 +683,19 @@ Mesh *AssimpLoader::Load(const std::string &_filename)
     {
       auto& animChan = anim->mChannels[chanIdx];
       auto chanName = ToString(animChan->mNodeName);
-      for (unsigned keyIdx = 0; keyIdx < animChan->mNumPositionKeys; ++keyIdx)
+      auto numKeys = std::max(
+          animChan->mNumPositionKeys, animChan->mNumRotationKeys);
+      // Position and rotation arrays might be different lengths,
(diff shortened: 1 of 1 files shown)


mxgrey:
I'm concerned that the approach of truncating the key frames to the array with a smaller number will throw away important animation information.

Instead I'd suggest fixing this by interpolating the smaller array to have the same number of key frames as the larger one, and then combining them in the poses used by the skeletal animation.

PR #489, opened 2023-02-13
Fix build error when using gz:: with ign-common4

scpeters:
# 🦟 Bug fix

Fixes a build error

## Summary

This doesn't yet fix the build error; it just illustrates it with some example code that fails to compile.

## Checklist
- [ ] Added tests


mjcarroll:
The issue here is that all of the headers in the `gz` folder still have the `ignition` namespace, when they should actually be in the `gz` namespace.  I will PR to address this.  It may have been an oversite, or the result of a bad merge?

scpeters:
I think the `gz/common/config.hh.in` is missing something:

* https://github.com/gazebosim/gz-math/blob/ign-math6/include/gz/math/config.hh.in#L19-L26
* https://github.com/gazebosim/gz-common/blob/ign-common4/include/gz/common/config.hh.in

mjcarroll:
> I think the gz/common/config.hh.in is missing something:


Added in [6f5c2ed](https://github.com/gazebosim/gz-common/pull/489/commits/6f5c2edaa1aa71a4998ad87c06f37a7289c0cef5)

scpeters:
> > I think the gz/common/config.hh.in is missing something:
> 
> Added in [6f5c2ed](https://github.com/gazebosim/gz-common/pull/489/commits/6f5c2edaa1aa71a4998ad87c06f37a7289c0cef5)

thanks!

mjcarroll (approved):

azeey (approved):
LGTM! My only question is: do we need to do the same for the headers in the components of gz-common?

mjcarroll:
> do we need to do the same for the headers in the components of gz-common?

Likely yes, I will go ahead and do that.  It may make sense to switch the tests to go ahead and use the `gz` namespace to guarantee that this is actually getting exercised, but I'm going to leave that to a follow-up.

azeey (approved):

PR #492, opened 2023-03-06
Merge ign-common4 ➡️  gz-common5

scpeters:
# ➡️  Forward port

Port `ign-common4 ` ➡️  `gz-common5`

Branch comparison: https://github.com/gazebosim/gz-common/compare/gz-common5...ign-common4

**Note to maintainers**: Remember to **Merge** with commit (not squash-merge or rebase)


mjcarroll:
There are some things failing here that I don't think were failing before.

graphics/src/Image_TEST.cc
```diff
@@ -93,6 +140,44 @@ TEST_F(ImageTest, ImageProperties)
       std::string::npos);
 }
 
+TEST_F(ImageTest, ImageSavePng)
+{
+  std::string pathOut;
+  ASSERT_TRUE(common::testing::TestTmpPath(pathOut));
```
scpeters, line 146: @mjcarroll `TestTmpPath` now has different implementations for Bazel and CMake in `gz-common5`. In order to forward-port this, should we split this test out into separate files for Bazel and CMake?

mjcarroll: Change to use this method: https://github.com/gazebosim/gz-common/blob/7b0a17cda12a6c0916787714995a4a1f094aeeaa/testing/include/gz/common/testing/TestPaths.hh#L184-L202

Which will let you get a file path in a temporary directory for testing, and automatically do the right thing (TM) between cmake and bazel.

For example `auto pathOut = gz::common::testing::TempPath("foo.png")`

scpeters: thanks!

scpeters:
CI is finally happy

mjcarroll (approved):

PR #511, opened 2023-05-18
Port: 4 to 5

mjcarroll:
# ➡️ Forward port

Port ign-common4 to gz-common5

Branch comparison: https://github.com/gazebosim/gz-common/compare/gz-common5...ign-common4

**Note to maintainers**: Remember to **Merge** with commit (not squash-merge or rebase)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -329,6 +329,44 @@
 
 ## Gazebo Common 4.x
 
+## Gazebo Common 4.7.0 (2023-05-17)
+
+1. Forward Ports: 3 -> 4 (#506)
+    * [Pull request #506](https://github.com/gazebosim/gz-common/pull/506)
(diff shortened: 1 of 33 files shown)


scpeters (commented):
I think the `gz/common/config.hh` includes are not universally needed in Garden. They were just needed in citadel and fortress. If we want to keep fewer includes, we could revert them from this merge-forward, but if we don't care then we can keep them

mjcarroll:
> I think the gz/common/config.hh includes are not universally needed in Garden. They were just needed in citadel and fortress. If we want to keep fewer includes, we could revert them from this merge-forward, but if we don't care then we can keep them

True, I will remove them

ahcorde (approved):

PR #515, opened 2023-05-22
Header was dropped in the forward port and breaks downstream

mjcarroll:


clalancette (commented):
This fixes the problem I was having building Gazebo Garden from source.  Thanks @mjcarroll !

azeey (approved):

PR #529, opened 2023-08-14
Remove deprecated std::iterator inheritance

mjcarroll:
# 🦟 Bug fix

Fixes #505

## Summary

Removes the use of the deprecated std::iterator template and replaces it with using the correct set of `using` statements.

I also added a `gzEnum` variadic template at clang's suggestion.  This should probably be preferred over the macro where relevant.

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #538, opened 2023-10-04
Prevent loading lightmaps if mesh is a glb file that has an occlusion-metallic-roughness texture

iche033:
# 🦟 Bug fix

## Summary

When loading this [water bottle glb file](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/WaterBottle), I manually saved the lightmap texture file (loaded by assimp) to examine what it looks like. It turns out the lightmap texture is just a copy of the [occlusion-metallic-roughness](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/WaterBottle/glTF/WaterBottle_occlusionRoughnessMetallic.png) texture, which is not correct.  There are issues reported upstream (see inline comments in code).

Currently if an occlusion-metallic-roughness texture exists, we extract the roughness texture map from the `g` channel and the metalness map is in the `b` channel of this texture.  I suspect that the right thing to do is to extract the `r` channel as a separate occlusion texture and use that as the lightmap (not quite the same but occlusion maps are not yet supported in gz-rendering). This needs to be investigated. For now, this PR prevents setting incorrect lightmap textures for glb meshes.

## Checklist
- [ ] Added tests

### DIFF
--- a/graphics/src/AssimpLoader.cc
+++ b/graphics/src/AssimpLoader.cc
@@ -425,6 +425,21 @@ MaterialPtr AssimpLoader::Implementation::CreateMaterial(
           this->GenerateTextureName(_scene, assimpMat, "Roughness"));
       pbr.SetRoughnessMap(texName, texData);
     }
+    // Load lightmap only if it is not a glb/glTF mesh that contains a
+    // MetallicRoughness texture
+    // It was found that lightmap field just stores the entire MetallicRoughness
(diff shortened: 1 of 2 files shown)


ahcorde (approved):

iche033:
the homebrew warning should be fixed in https://github.com/gazebosim/gz-common/pull/529

PR #542, opened 2023-10-12
Drop the internal vendored version of TinyXML2

mjcarroll:
All platforms have decently-supported TinyXML2 packages, so we have no need to vendor it anymore.


graphics/src/CMakeLists.txt
```diff
@@ -39,47 +31,14 @@ if(${GzAssimp_VERSION} STRLESS "5.2.0")
...
+target_link_libraries(${graphics_target}
+  PRIVATE
+    TINYXML2::TINYXML2)
 
-# define of tinxml2 major version >= 6
-# https://github.com/gazebosim/gz-common/issues/28
-if (NOT TINYXML2_VERSION VERSION_LESS "6.0.0")
-  message(STATUS "TINYXML2_VERSION ${TINYXML2_VERSION} >= 6.0.0")
-  target_compile_definitions(${graphics_target}
-    PRIVATE "TINYXML2_MAJOR_VERSION_GE_6")
```
scpeters, line 84: This macro is used in ifdefs to support versions of tinyxml2 before 6.0. I think most of our target platforms are already using newer versions than tinyxml2 6, so we could decide that we require at least version 6.0 of tinyxml2, but if we do so, I think we should give a cmake error if an older version is detected and remove the ifdefs around TINYXML2_MAJOR_VERSION_GE_6 in our code. If we don't want to do all that now, then I would suggest restoring this code block.

mjcarroll: I think Windows was the problem, it looks like vcpkg is well ahead of that now: https://github.com/microsoft/vcpkg/blob/ba823b1e1f330a2679273d29bda5b2ab6a603de8/ports/tinyxml2/vcpkg.json#L3

I will make sure we are searching for 6+ and remove the guards.

mjcarroll: Done in 2c535d9

graphics/src/CMakeLists.txt
```diff
@@ -39,47 +30,14 @@ if(${GzAssimp_VERSION} STRLESS "5.2.0")
...
-    # cannot link to those symbols and must generate its own.
-    target_sources(UNIT_ColladaExporter_TEST PRIVATE tinyxml2/tinyxml2.cpp)
-
-  endif()
-
-endif()
+# If we are using an external copy of tinyxml2, add its imported target
+target_link_libraries(${graphics_target}
+  PRIVATE
+    TINYXML2::TINYXML2)
```
scpeters, line 36: I would move this to the `target_link_libraries` call on line 14

graphics/src/CMakeLists.txt
```diff
@@ -39,47 +30,14 @@ if(${GzAssimp_VERSION} STRLESS "5.2.0")
...
 
-# define of tinxml2 major version >= 6
-# https://github.com/gazebosim/gz-common/issues/28
-if (NOT TINYXML2_VERSION VERSION_LESS "6.0.0")
-  message(STATUS "TINYXML2_VERSION ${TINYXML2_VERSION} >= 6.0.0")
-  target_compile_definitions(${graphics_target}
-    PRIVATE "TINYXML2_MAJOR_VERSION_GE_6")
+# The collada exporter test uses tinyxml2, so we must link it if we're using
+# an external copy. The graphics target considers tinyxml2 to be a private
```
scpeters, line 39:
```suggestion
# The collada exporter test uses tinyxml2, so we must link it.
# The graphics target considers tinyxml2 to be a private
```

graphics/src/CMakeLists.txt
mjcarroll: Done.

scpeters (approved):

PR #546, opened 2023-10-25
Fix segfault in case of no write access to log dir

jmachowinski:
# 🦟 Bug fix

Fixes a Segfault when building Ign common3 with pbuilder

## Summary
If there is no write access to the log directory, the logger will call itself recusively in an infinite loop and segfault.
Note, this bug is also present in the latest version of ign common.

### DIFF
--- a/src/Console.cc
+++ b/src/Console.cc
@@ -277,7 +277,7 @@ void FileLogger::Init(const std::string &_directory,
   // Create the directory if it doesn't exist.
   if(!createDirectories(logPath, std::cerr))
   {
-      std::cerr << "Failed to generate log directories."
+      std::cerr << "Failed to generate log directories. "
                 << "No log file will be generated.";
       return;
   }
--- a/src/Console.cc
+++ b/src/Console.cc
@@ -261,7 +261,7 @@ void FileLogger::Init(const std::string &_directory,
   {
     if (!env(IGN_HOMEDIR, logPath))
     {
-      std::cerr << "Missing HOME environment variable."
+      std::cerr << "Missing HOME environment variable. "
                 << "No log file will be generated.";
       return;
     }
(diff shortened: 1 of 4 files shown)


mjcarroll:
> Note, this bug is also present in the latest version of ign common.

once merged, this will be forward ported across versions.

src/Console.cc
```diff
@@ -261,8 +261,8 @@ void FileLogger::Init(const std::string &_directory,
   {
     if (!env(IGN_HOMEDIR, logPath))
     {
-      ignerr << "Missing HOME environment variable."
-        << "No log file will be generated.";
+      std::cerr << "Missing HOME environment variable."
```
iche033, line 264: space between the two messages or new line?

```suggestion
      std::cerr << "Missing HOME environment variable. "
```

src/Console.cc
```diff
@@ -275,7 +275,12 @@ void FileLogger::Init(const std::string &_directory,
   auto* buf = dynamic_cast<FileLogger::Buffer*>(this->rdbuf());
 
   // Create the directory if it doesn't exist.
-  createDirectories(logPath);
+  if(!createDirectories(logPath, std::cerr))
+  {
+      std::cerr << "Failed to generate log directories."
```
iche033, line 280:
```suggestion
      std::cerr << "Failed to generate log directories. "
```

jmachowinski:
ping ?

src/EnumIface_TEST.cc
```diff
@@ -41,7 +41,7 @@ IGN_ENUM(myTypeIface, MyType, MY_TYPE_BEGIN, MY_TYPE_END,
 /////////////////////////////////////////////////
 TEST_F(EnumIfaceTest, StringCoversion)
 {
-  MyType type;
+  MyType type = MyType::TYPE2;
```
mjcarroll, line 44: Is this change related to the rest of the pull request?

jmachowinski: No, this is a separate commit fixing a compile warning.

PR #555, opened 2023-11-09
Another visibility fix attempt

mjcarroll:


mjcarroll:
@j-rivero tests green, just a C4251 warning, which should be addressabl

PR #565, opened 2023-12-14
Backwards compatible assimp texture name fix

luca-della-vedova:
# 🦟 Bug fix

Supersedes #564 

## Summary
Apologies for the noise! Hopefully third time is the charm? When working on #564 and #563 I realized that we do have a way to ship changes that depend on the assimp version, by using a define.
This allows us to have the fix in Harmonic (LTS) without being blocked by the fact that `gz-common4` is also shipped in Garden (that is short term support anyway).
If it's desirable I can open a PR targeting main that gets rid of all the `ifdef`s since they are only needed for compatibility with assimp 5.0.x that was last shipped in Ubuntu 20.04.

## Checklist
- [ ] Added tests
- [x] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))

### DIFF
--- a/graphics/src/AssimpLoader.cc
+++ b/graphics/src/AssimpLoader.cc
@@ -584,7 +584,12 @@ ImagePtr AssimpLoader::Implementation::LoadEmbeddedTexture(
 std::string AssimpLoader::Implementation::GenerateTextureName(
     const aiScene* _scene, aiMaterial* _mat, const std::string& _type) const
 {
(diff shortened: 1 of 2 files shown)


ahcorde (approved):

mjcarroll:
macOS CI is unofficial and not needed here.  The other two jenkins jobs are artifacts of the previous job names and are safe to ignore.  (I'm removing them as required now).

mjcarroll (approved):

PR #573, opened 2024-01-20
Be louder when graphics is missing for geospatial

mjcarroll:
# 🦟 Bug fix

## Summary

Users building from source could get in a situation where dependencies of the `graphics` component weren't available, so that the component is not created.  In this case, the `geospatial` component still has a hard dependency on the `graphics` component, but it is still attempting to build.

This adds gating logic to prevent the `geospatial` component from building when the `graphics` component isn't build and adds a warning message for the user.

### DIFF
--- a/geospatial/src/CMakeLists.txt
+++ b/geospatial/src/CMakeLists.txt
@@ -1,24 +1,27 @@
-gz_get_libsources_and_unittests(sources gtest_sources)
+if (TARGET ${PROJECT_LIBRARY_TARGET_NAME}-graphics)
+  gz_get_libsources_and_unittests(sources gtest_sources)
+  gz_add_component(geospatial
+    SOURCES ${sources}
+    DEPENDS_ON_COMPONENTS graphics
+    GET_TARGET_NAME geospatial_target)
 
-gz_add_component(geospatial
-  SOURCES ${sources}
-  DEPENDS_ON_COMPONENTS graphics
-  GET_TARGET_NAME geospatial_target)
+  target_link_libraries(${geospatial_target}
+    PUBLIC
+      ${PROJECT_LIBRARY_TARGET_NAME}-graphics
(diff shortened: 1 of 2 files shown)


Ryanf55 (approved):
The warning is much more obvious and it skips building geospatial if graphics do not exist. Thanks!

image

I'll create another bug for the follow up (assuming it's optional) since it's a different problem.

geospatial/src/CMakeLists.txt
```diff
@@ -1,24 +1,27 @@
...
+if (TARGET ${PROJECT_LIBRARY_TARGET_NAME}-graphics)
+  gz_get_libsources_and_unittests(sources gtest_sources)
+  gz_add_component(geospatial
+    SOURCES ${sources}
+    DEPENDS_ON_COMPONENTS graphics
+    GET_TARGET_NAME geospatial_target)
 
-gz_add_component(geospatial
-  SOURCES ${sources}
-  DEPENDS_ON_COMPONENTS graphics
```
scpeters, line 5: I thought the `DEPENDS_ON_COMPONENTS` parameter would generate a complaint if the needed component wasn't available, but apparently it doesn't

scpeters:
an alternative would be to change every `find_package` call with `REQUIRED_BY graphics` to `REQUIRED_BY graphics geospatial`; not sure that it makes more sense though

scpeters (approved):

geospatial/src/CMakeLists.txt
mjcarroll: That was also my understanding, but does not seem to be the case.

mjcarroll:
> an alternative would be to change every find_package call with REQUIRED_BY graphics to REQUIRED_BY graphics geospatial; not sure that it makes more sense though

It would be the other way around, correct?  Anyone who requires geospatial also will require graphics.

bperseghetti (approved):

PR #578, opened 2024-03-01
Fix noise issue that appears in certain image textures

azeey:
# 🦟 Bug fix

## Summary
The issue is that if a pixel's color value in 8 bit format is close 1 (out of 255), when assigned to `math::Color`, `math::Color::Clamp` assumes it to be in the floating point range [0.0 - 1.0] instead of 8 bit unsigned range of [0 - 255]. The fix here is to convert colors to floating point at the source instead of relying on `Color::Clamp`.

Reference glTF file: [Damaged helmet](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/DamagedHelmet)

### Before
Screenshot 2024-03-01 17 12 39

### After
Screenshot 2024-03-01 17 11 08

## Test it

* Extract [DamagedHelmet.zip](https://github.com/gazebosim/gz-common/files/14466833/DamagedHelmet.zip) and run `gz sim test.sdf` from the same directory

## Checklist
- [ ] Added tests

### DIFF
--- a/graphics/src/Image.cc
+++ b/graphics/src/Image.cc
@@ -533,7 +533,8 @@ math::Color Image::Pixel(unsigned int _x, unsigned int _y) const
         << _x << " " << _y << "] \n";
       return clr;
     }
-    clr.Set(firgb.rgbRed, firgb.rgbGreen, firgb.rgbBlue);
+    clr.Set(firgb.rgbRed / 255.0f, firgb.rgbGreen / 255.0f,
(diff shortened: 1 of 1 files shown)


iche033 (approved):
ouch, I remember running into the same issue before and worked around it. We should probably change the way `math::Color::Clamp` works in `main`

ahcorde (approved):

PR #583, opened 2024-03-13
Adds new function in MeshManager for performing convex decomposition

iche033:
# 🎉 New feature

## Summary

Adds function for doing convex decomposition using the header only library [V-HACD](https://github.com/kmammou/v-hacd).

The new function take a submesh and returns a vector of decomposed submeshes. Currently a couple of parameters are exposed:
* `_maxConvexHulls`: max number of convex hulls to produce
    * value: 16 (default in V-HACD: 64)
* `_voxelResolution`:  resolution of voxels to use when decomposing the submesh
    * value: 200000 (default in V-HACD: 400000)
   
These default values for these parameters are lower than the default values provided in V-HACD for a faster convex decomposition process as the process with original values could take quite some time for complex meshes.

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #585, opened 2024-03-14
Adds new function in MeshManager for performing convex decomposition

iche033:
# 🎉 New feature

replaces #583 

## Summary

Adds function for doing convex decomposition using the header only library [V-HACD](https://github.com/kmammou/v-hacd).

The new function take a submesh and returns a vector of decomposed submeshes. Currently a couple of parameters are exposed:
* `_maxConvexHulls`: max number of convex hulls to produce
    * value: 16 (default in V-HACD: 64)
* `_voxelResolution`:  resolution of voxels to use when decomposing the submesh
    * value: 200000 (default in V-HACD: 400000)
   
These default values for these parameters are lower than the default values provided in V-HACD for a faster convex decomposition process as the process with original values could take quite some time for complex meshes.

## Checklist
- [x] Added tests


graphics/src/VHACD.h
azeey, line 1: I think it would be best if we put this in its own directory with other metadata such as a `LICENSE` file and the version hash used from upstream, etc.

graphics/include/gz/common/MeshManager.hh
```diff
@@ -280,6 +280,21 @@ namespace gz
           const Mesh *_m2, const int _operation,
           const gz::math::Pose3d &_offset = gz::math::Pose3d::Zero);
 
+      /// \brief Perform convex decomposition on a submesh.
+      /// The submesh is decomposed into multiple convex submeshes. The output
+      /// submeshes contain vertices and indices but texture coordinates
+      /// not preserved.
```
azeey, line 286:
```suggestion
      /// submeshes contain vertices and indices but texture coordinates
      /// are not preserved.
```

graphics/include/gz/common/MeshManager.hh
```diff
@@ -280,6 +280,21 @@ namespace gz
...
+      /// \param[in] _subMesh Input submesh to decompose.
+      /// \param[in] _maxConvexHulls Maximum number of convex hull submeshes to
+      /// produce.
+      /// \param[in] _voxelResolution Voxel resolution to use. Higher value
+      /// produces more accurate shapes.
+      /// \return A vector of decomposed submeshes.
+      public: std::vector<SubMesh> ConvexDecomposition(
+                  const common::SubMesh *_subMesh,
```
azeey, line 294: Any reason for using a pointer there instead of a const ref? IMO, it would be best for the caller to be responsible for checking that the submesh is not a nullptr.

graphics/src/MeshManager.cc
```diff
@@ -1648,3 +1660,82 @@ void MeshManager::SetAssimpEnvs()
...
+  std::vector<SubMesh> decomposed;
+
+  if (!_subMesh)
+    return decomposed;
+
+  auto vertexCount = _subMesh->VertexCount();
+  auto indexCount = _subMesh->IndexCount();
+  auto triangleCount = indexCount / 3u;
+
+  float *points = new float[vertexCount * 3u];
```
azeey, line 1679: Consider using `std::vector` or `std::unique_ptr`.

graphics/src/MeshManager.cc
```diff
@@ -1648,3 +1660,82 @@ void MeshManager::SetAssimpEnvs()
...
+  for (std::size_t i = 0; i < vertexCount; ++i)
+  {
+    std::size_t idx = i * 3u;
+    points[idx] = _subMesh->Vertex(i).X();
+    points[idx + 1] = _subMesh->Vertex(i).Y();
+    points[idx + 2] = _subMesh->Vertex(i).Z();
+  }
+
+  uint32_t *indices = new uint32_t[indexCount];
```
azeey, line 1688: Consider using `std::vector` or `std::unique_ptr`.

graphics/src/MeshManager.cc
```diff
@@ -1648,3 +1660,82 @@ void MeshManager::SetAssimpEnvs()
...
+  parameters.m_resolution = _voxelResolution;
+  parameters.m_asyncACD = true;
+  iface->Compute(points, vertexCount, indices, triangleCount, parameters);
+  while (!iface->IsReady())
+  {
+    std::this_thread::sleep_for(std::chrono::nanoseconds(10000));
+  }
+
+  if (!iface->GetNConvexHulls())
+    return decomposed;
```
azeey, line 1706: Should we add an error message here?

graphics/src/MeshManager.cc
```diff
@@ -1648,3 +1660,82 @@ void MeshManager::SetAssimpEnvs()
...
+  VHACD::IVHACD *iface = VHACD::CreateVHACD_ASYNC();
+  VHACD::IVHACD::Parameters parameters;
+  parameters.m_maxConvexHulls = _maxConvexHulls;
+  parameters.m_resolution = _voxelResolution;
+  parameters.m_asyncACD = true;
+  iface->Compute(points, vertexCount, indices, triangleCount, parameters);
+  while (!iface->IsReady())
+  {
+    std::this_thread::sleep_for(std::chrono::nanoseconds(10000));
+  }
```
azeey, line 1703: Should we have a timeout?

graphics/src/MeshManager.cc
```diff
@@ -1648,3 +1660,82 @@ void MeshManager::SetAssimpEnvs()
...
+    SubMesh convexMesh;
+    for (std::size_t i = 0u; i < ch.m_points.size(); ++i)
+    {
+      const VHACD::Vertex &p = ch.m_points[i];
+      gz::math::Vector3d vertex(p.mX, p.mY, p.mZ);
+      convexMesh.AddVertex(vertex);
+
+      // add dummy normal
+      auto norm = vertex;
+      norm.Normalize();
```
azeey, line 1722: If `vertex` is (0,0,0), would `Normalize()` work okay?

graphics/src/VHACD.h
traversaro: Having the header in a separate library would also simplify the eventual/future modifications to permit optional de-vendoring of the library.

graphics/src/MeshManager.cc
iche033: should be ok as there is a check for 0 in gz-math. 

We're not currently using normals for collisions so these should be good enough. If users do need valid normals, it can be done by calling `submesh.RecalculateNormals()`

azeey: If it's just a dummy normal, why not just use a constant `0, 0, 1` or something and avoid the Normalize call? Also, if anyone ever uses the normals, it would be obvious something is wrong if all the normals are the same than if they're seemingly random.

graphics/src/VHACD.h
iche033: moved header it into it's own directory with license file and a readme with commit hash. [a6c5a2f](https://github.com/gazebosim/gz-common/pull/585/commits/a6c5a2f6c565e4ec636c1f492631312b470a5696)

graphics/include/gz/common/MeshManager.hh
iche033: [a6c5a2f](https://github.com/gazebosim/gz-common/pull/585/commits/a6c5a2f6c565e4ec636c1f492631312b470a5696)

graphics/src/MeshManager.cc
iche033: changed to `std::vector`. [a6c5a2f](https://github.com/gazebosim/gz-common/pull/585/commits/a6c5a2f6c565e4ec636c1f492631312b470a5696)

graphics/src/MeshManager.cc
iche033: changed to `std::vector`. [a6c5a2f](https://github.com/gazebosim/gz-common/pull/585/commits/a6c5a2f6c565e4ec636c1f492631312b470a5696)

graphics/src/MeshManager.cc
iche033: added timeout of 5mins. [a6c5a2f](https://github.com/gazebosim/gz-common/pull/585/commits/a6c5a2f6c565e4ec636c1f492631312b470a5696)

graphics/include/gz/common/MeshManager.hh
iche033: mainly to be consistent with other functions in MeshManager which all take mesh pointers. I updated this to const ref. [a6c5a2f](https://github.com/gazebosim/gz-common/pull/585/commits/a6c5a2f6c565e4ec636c1f492631312b470a5696)

graphics/src/MeshManager.cc
iche033: > it would be obvious something is wrong if all the normals are the same

if all normals are the same, it could just mean it's a valid plane. To be safe, I added the `RecalculateNormals` call to now generate valid normals.

azeey (approved):
LGTM!

graphics/src/MeshManager.cc
```diff
@@ -16,11 +16,23 @@
...
+#include <cstdint>
 #include <mutex>
+#include <string>
 #include <unordered_map>
 #include <unordered_set>
-#include <cctype>
+#include <vector>
+
+// Suppress warnings for VHACD
+#pragma GCC diagnostic push
```
azeey, line 29: This is generating a windows warning.

azeey (changes requested):

graphics/src/MeshManager.cc
iche033: fxied in [d92687c](https://github.com/gazebosim/gz-common/pull/585/commits/d92687c4f150f2cdee48f21ff34bd0f6a884ee0b)

azeey (approved):

PR #603, opened 2024-05-03
Fix deprecation warnings

azeey:
# 🦟 Bug fix

## Summary
Fix deprecation warnings:
* Replace `avcodec_close` with `avcodec_free_context`
* Suppress deprecation of `std::iterator` since we can't change it without breaking ABI. I only did this for macOS, since that's the only platform that emits the warning.

## Checklist
- [ ] Added tests

### DIFF
--- a/av/src/AudioDecoder.cc
+++ b/av/src/AudioDecoder.cc
@@ -67,7 +67,13 @@ void AudioDecoder::Cleanup()
 {
   // Close the codec
   if (this->data->codecCtx)
+  {
+#if LIBAVFORMAT_VERSION_MAJOR < 59
+    avcodec_close(this->data->codecCtx);
+#else
     avcodec_free_context(&this->data->codecCtx);
+#endif
+  }
 
   // Close the audio file
   if (this->data->formatCtx)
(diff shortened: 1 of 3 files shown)


av/src/AudioDecoder.cc
```diff
@@ -67,7 +67,7 @@ void AudioDecoder::Cleanup()
 {
   // Close the codec
   if (this->data->codecCtx)
-    avcodec_close(this->data->codecCtx);
+    avcodec_free_context(&this->data->codecCtx);
```
scpeters, line 70: this change causes a test failure on 20.04; perhaps we can `ifdef` on the software version if we can figure out the difference in the version variables between 20.04 and 22.04?

azeey: I didn't do much investigation into whether we can simply replace `avcodec_close` with `avcodec_free_context`, but let me try it on a newer version of Ubuntu to see if it fails everywhere.

azeey: Looking at this more closely, I think it's more about how `codecCtx` was created in the first place. In https://github.com/gazebosim/gz-common/pull/603/commits/d38e4d354ff46bb22e2b874cfae735ddf8265bd8, I've used `avcodec_free_context` if `codeCtx` was created using `avcodec_alloc_context3`, which is inline with the [ffmpeg](https://ffmpeg.org/doxygen/6.0/group__lavc__core.html#gae80afec6f26df6607eaacf39b561c315) documentation

mjcarroll (approved):

PR #609, opened 2024-05-12
SubMesh::RecalculateNormals improvement

ntfshard:
# 🦟 Bug fix

Fixes #N/A

## Summary
Previous version of `RecalculateNormals` member function has a O(n^2) complexity because inner loop where we are looking for normal elements. But we already know indexes of this elements. So we can get rid of inner loop.
And seems we should check `indices.size() < 3u || indices.size() % 3u != 0` due to if it will be 4, main loop will start and on second iteration it will start reading out of array.
Also seems condition in the begin should check amount of indexes/vector, but not normals, due to we are resizing it later.
Maybe this function also should have a check of `enum PrimitiveType` of data, seems attempt to call it for `POINTS` and `LINES` is not legit (for rest 4 types maybe it's ok, not sure)

## Checklist
- [ ] Added tests

### DIFF
--- a/graphics/src/SubMesh.cc
+++ b/graphics/src/SubMesh.cc
@@ -644,54 +644,11 @@ void SubMesh::RecalculateNormals()
         this->dataPtr->vertices[this->dataPtr->indices[i+2]];
     gz::math::Vector3d n = gz::math::Vector3d::Normal(v1, v2, v3);
 
-#if 0
-  /*
-  i7-11800H @ Ubuntu22 VM; 2^14 triangles
-    SubMeshTest.NormalsRecalculation (1105 ms)
-    SubMeshTest.NormalsRecalculation (1158 ms)
-    SubMeshTest.NormalsRecalculation (1117 ms)
-    SubMeshTest.NormalsRecalculation (1137 ms)
-    SubMeshTest.NormalsRecalculation (1156 ms)
-  */
-    for (unsigned int j = 0; j < this->dataPtr->vertices.size(); ++j)
-    {
-      gz::math::Vector3d v = this->dataPtr->vertices[j];
-      if (v == v1 || v == v2 || v == v3)
-      {
-        this->dataPtr->normals[j] += n;
-      }
-    }
-#else
-  // same env, ~1ms
-  // FAILS: ASSERT_NE(submesh->Normal(0), submesh->Normal(1));
-    /*this->dataPtr->normals[this->dataPtr->indices[i]] += n;
-    this->dataPtr->normals[this->dataPtr->indices[i+1]] += n;
-    this->dataPtr->normals[this->dataPtr->indices[i+2]] += n;*/
-
-    for (const auto& point : {v1, v2, v3})
+    for (const auto &point : {v1, v2, v3})
       neighbors.Visit(point, [&](unsigned int index)
       {
         this->dataPtr->normals[index] += n;
       });
-
-    //auto& nv1 = neighbors[v1];
-    //nv1.insert(this->dataPtr->indices[i]);
-    /*for (const auto index: neighbors[v1])
-      this->dataPtr->normals[index] += n;*/
-
-    //neighbors[v2].insert(this->dataPtr->indices[i+1]);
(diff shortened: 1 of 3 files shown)


ntfshard:
I force-pushed commit due to previous one was not signed and I made boundary check more strict

iche033:
looks like there is still an issue with DCO, can you try fix it again?

iche033:
I was testing this with a capsule shape and I noticed that a seam is now visible after the changes:

Before:
cylinder_normal_before

After:
cylinder_normal_after

The `RecaculateNormals` function was ported from gazebo-classic so I went through it's history and found this: 
https://github.com/gazebosim/gazebo-classic/commit/ea54b563a82198fafc42744d5aad60ecad86e1d0

From the commit msg, the loop was actually implemented on purpose to make objects appear smoother. The effect of that loop is that more normals are updated instead of limiting to 3 normals. Not efficient but produces smoother results.

ntfshard:
I see. Thank you for testing and clarification. The main problem with current code that it could work about a minute if polygon has quite a lot of triangles even on a modern hardware. Maybe I can try to re-implement initial behaviour with unordered_map. I guess it should be ok to exchange execution time to a memory consumption spike

iche033:
> Maybe I can try to re-implement initial behaviour with unordered_map. I guess it should be ok to exchange execution time to a memory consumption spike

yes that would be good to try, thanks

graphics/src/SubMesh.cc
```diff
@@ -573,10 +574,54 @@ void SubMesh::FillArrays(double **_vertArr, int **_indArr) const
   }
 }
 
+namespace {
+// Simple way to finding neighbors by grouping all vertexes
```
skorobogatydmitry, line 578:
```suggestion
// Simple way to find neighbors by grouping all vertices
```

ntfshard:
Sorry for a delay. Here is an another solution. We are grouping all points by x coordinate and check only points with a same x coordinate (and adjacent groups, due to `operator==` for `Vector3d` is actually treat equal a coordinate in a plus/minus Epsilon range.)

For a relatively small sub-meshes it works slower(on my hardware, around 8 milliseconds) than previous one. It looks like a minimal running time constant. Not sure how critical is it. If it's a showstopper, I can dispatch algorithms according to a input size, but it will make code more complex(for a future readers), that's why I came with this version first.

graphics/src/SubMesh.cc
```diff
@@ -573,10 +573,56 @@ void SubMesh::FillArrays(double **_vertArr, int **_indArr) const
...
+};
+}  // namespace
+
 //////////////////////////////////////////////////
 void SubMesh::RecalculateNormals()
 {
-  if (this->dataPtr->normals.size() < 3u)
+  if (this->dataPtr->indices.empty()
+      || this->dataPtr->indices.size() % 3u != 0)
     return;
```
ntfshard, line 627: It feels like it also should check: `this->dataPtr->primitiveType == SubMesh::TRIANGLES` but not 100% sure that it's only one primitiveType

graphics/src/SubMesh.cc
```diff
@@ -573,10 +573,56 @@ void SubMesh::FillArrays(double **_vertArr, int **_indArr) const
...
+  {
+    for (unsigned int i = 0; i < _indices.size(); ++i)
+    {
+      const auto index = _indices[i];
+      this->neighbors[_vertices[index].X()].push_back(index);
+    }
+  }
+
+  template<typename Visitor>
+  void Visit(const gz::math::Vector3d &_point, Visitor _v) const
```
iche033, line 594: can you add some documentation to explain the logic in this function? I think it's to find and include all indices of vertices that are within the tolerance of the `math::equal` checks, so that their normals all count towards the final calculation?

iche033 (commented):
The new approach looks fine to me. I tested the code and the capsule renders fine without the seam. Just one minor comment on adding some doc to explain the logic.

iche033 (approved):
looks good, thanks for iterating

PR #627, opened 2024-08-28
Remove flushes and fix tests

azeey:
With https://github.com/gazebosim/gz-utils/pull/142, it's no longer necessary to flush manually. This removes the need for `_fileInitialize` and fixes the behavior of `auto_default.log`.

### DIFF
--- a/include/gz/common/Console.hh
+++ b/include/gz/common/Console.hh
@@ -42,8 +42,6 @@ namespace gz
       /// \param[in] _file Filename.
       /// \param[in] _line Line number.
       /// \param[in] _logLevel Log level.
-      /// \param[in] _fileInitialize True if the file logger needs to be
-      /// initialized or false otherwise.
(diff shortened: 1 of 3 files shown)


include/gz/common/Console.hh
```diff
@@ -46,8 +46,7 @@ namespace gz
       /// initialized or false otherwise.
```
caguero, line 46: We should remove the doxygen comment for `fileInitialize`.

caguero (changes requested):

include/gz/common/Console.hh
caguero: [a1488d5](https://github.com/gazebosim/gz-common/pull/627/commits/a1488d589d0a4d3b0d374f6d59e93cecbc6e3575)

caguero (approved):

PR #635, opened 2024-09-05
Fix severity level of gzlog

### LINKED ISSUE #632: gzlog prints msgs as [error]
## Environment
* OS Version: 24.04
* Source or binary build? source, gz-common6
  
  

## Description
* Expected behavior: In gz-common5, the `gzlog` msg does not get printed to the console but only gets logged to the `~/.gz/auto_default.log`
* Actual behavior: Use of `gzlog` prints msgs to console as `[error]` which can be misleading

## Steps to reproduce

1. `gz sim -v 4 shapes.sdf`
2. Add any gui plugin and see an error msg printed

## Output

Adding the `ApplyForceTorque` gui plugin prints out:

```
(2024-09-04 23:18:50.800) [error] [MainWindow.cc:268] [GUI] Add [ApplyForceTorque] via menu
```

azeey:
# 🦟 Bug fix

Fixes #632 

## Summary
The severity level of `gzlog` was incorrectly set to `err`. Changing this to `trace` solves two problems:
1. The log message is now a lot less alarming 
2. When running `gz sim -v4`, this message is not shown on the console. It will only be logged to file. It does show up on the console if you run with `-v5`. 
 
The second point is technically a behavior change since in Harmonic, regardless of the verbosity level, `gzlog` would never show up on the console. However, duplicating that behavior with the new Console implementation proved to be very complicated. So, I propose we change the meaning of `gzlog` to be equivalent to `gztrace`.

---

**Note**
We'll have to make new prereleases of all downstream libraries for this fix to propagate. I vote we just wait for the stable release.

## Checklist
- [ ] Added tests

### DIFF
--- a/graphics/src/ColladaLoader_TEST.cc
+++ b/graphics/src/ColladaLoader_TEST.cc
@@ -112,6 +112,7 @@ TEST_F(ColladaLoader, LoadZeroCount)
       common::testing::TestFile("data", "zero_count.dae"));
   ASSERT_TRUE(mesh);
 #ifndef _WIN32
+  common::Console::Root().RawLogger().flush();
   std::string log = LogContent();
 
   // Expect no errors about missing values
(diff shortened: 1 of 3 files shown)


iche033:
The test failures look related, probably due to behavior change?

iche033 (approved):

PR #654, opened 2024-12-06
Check normal presence before trying to read asMesh normals

qingyouzhao:
# 🦟 Bug fix

Fixes there isn't a bug number

## Summary
An assimp mesh could contain either
- the same number of vertices and normals
- only vertices

Existing code crashes when a mesh contains no normals

## Checklist
- [ ] Added tests
- [x] Consider updating Python bindings (if the library has them)

### DIFF
--- a/graphics/src/AssimpLoader.cc
+++ b/graphics/src/AssimpLoader.cc
@@ -700,6 +700,9 @@ SubMesh AssimpLoader::Implementation::CreateSubMesh(
     subMesh.AddIndex(face.mIndices[2]);
   }
   subMesh.SetMaterialIndex(_assimpMesh->mMaterialIndex);
+  if (subMesh.NormalCount() == 0u){
+    subMesh.RecalculateNormals();
+  }
   return subMesh;
 }
 
--- a/graphics/src/AssimpLoader.cc
+++ b/graphics/src/AssimpLoader.cc
@@ -667,18 +667,20 @@ SubMesh AssimpLoader::Implementation::CreateSubMesh(
   {
     // Add the vertex
     math::Vector3d vertex;
-    math::Vector3d normal;
     vertex.X(_assimpMesh->mVertices[vertexIdx].x);
     vertex.Y(_assimpMesh->mVertices[vertexIdx].y);
     vertex.Z(_assimpMesh->mVertices[vertexIdx].z);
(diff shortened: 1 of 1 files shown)


azeey (approved):
LGTM!

azeey:
I'm not sure if the rest of Gazebo handles the case where there are no normals properly. @iche033 do you know?

iche033:
> I'm not sure if the rest of Gazebo handles the case where there are no normals properly. @iche033 do you know?

gz-rendering should be able to [handle](https://github.com/gazebosim/gz-rendering/blob/9e874c0483d3c3cd34477baa59db8f7d9341a14b/ogre2/src/Ogre2MeshFactory.cc#L371) meshes without normals

However, I see that gz-physics' dartsim plugin [requires](https://github.com/gazebosim/gz-physics/blob/0fb5974b1ddc109d7f6c881cd5aafcd402cdfd5f/dartsim/src/CustomMeshShape.cc#L142-L151) normals to be present. 

To be safe I think we should add a check and auto generate normals if the assimp mesh does not contain normals, i.e. in the same function before `return subMesh;`  we can add something like:

```cpp
if (subMesh.NormalCount() == 0u)
  subMesh.RecalculateNormals();
```

qingyouzhao:
> > I'm not sure if the rest of Gazebo handles the case where there are no normals properly. @iche033 do you know?
> 
> gz-rendering should be able to [handle](https://github.com/gazebosim/gz-rendering/blob/9e874c0483d3c3cd34477baa59db8f7d9341a14b/ogre2/src/Ogre2MeshFactory.cc#L371) meshes without normals
> 
> However, I see that gz-physics' dartsim plugin [requires](https://github.com/gazebosim/gz-physics/blob/0fb5974b1ddc109d7f6c881cd5aafcd402cdfd5f/dartsim/src/CustomMeshShape.cc#L142-L151) normals to be present.
> 
> To be safe I think we should add a check and auto generate normals if the assimp mesh does not contain normals, i.e. in the same function before `return subMesh;` we can add something like:
> 
> ```c++
> if (subMesh.NormalCount() == 0u)
>   subMesh.RecalculateNormals();
> ```

@iche033 Added.

iche033:
@mergifyio backport gz-common5 gz-common6

PR #451, opened 2022-10-10
Fixed MeshManager Singleton

ahcorde:
Signed-off-by: ahcorde <ahcorde@gmail.com>

# 🦟 Bug fix

## Summary

Related with https://github.com/gazebosim/gz-sim/issues/1725

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #694, opened 2025-07-10
Fix for singleton in Profiler

ntfshard:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
Recurring problem with dangerous `SingletonT`:
* https://github.com/gazebosim/gz-common/issues/451
* https://github.com/gazebosim/gz-rendering/pull/734
* https://github.com/gazebosim/gz-rendering/pull/876
* guess `OptixRenderEngine` is still affected

In this PR I propose similar to MeshManager change (keep inheritance, add method which hides method from base class).
Also added comment to `SingletonT` (tbh, I'd rather remove usage of it across the product, deprecate and remove)

I target it to common5 to start discussion about better way of handling this problem. But just merging is still fine, I'm just user.

## Checklist
- [ ] Added tests
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)


profiler/src/Profiler.cc
```diff
@@ -115,3 +115,10 @@ bool Profiler::SetImplementation(std::unique_ptr<ProfilerImpl> _impl)
   gzdbg << "Gazebo profiling with: " << this->impl->Name() << std::endl;
   return true;
 }
+
+//////////////////////////////////////////////////
+Profiler *Profiler::Instance()
+{
+  static Profiler profiler;
+  return &profiler;
```
azeey, line 123:
```suggestion
  return SingletonT<Profiler>::Instance();
```

I think doing this might maintain ABI compatibility. Can you give it a try?

ntfshard: And about NeverDestroyed (very short version). IMHO it's a very dirty hack.
1. Valgrind will see indirect memory leak if object inside it allocates memory.
2. AFAIU it was invented for cases of library loading with dlopen and unloaded before termination. In this case code of destructor of static object is unloaded and attempt to call it will lead to crash on termination (I saw similar case with destructor of exception on plugin loading mechanism in one of project where I worked)
3. I guess it can be treated as UB. Can deep dive in standard to find more accurate interpretation, but it will take a lot of time

azeey: I think you're right that it'll probably work fine during initialization, but your approach would still be problematic during destruction since the class is not trivially destructible. We follow the Google C++ style guide and if you look at the "Common patterns" subsection under [Static and Global Variables](
https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables), you will see that the preferred approach is to do `static const auto& impl = *new T(args...);`, which is what `NeverDestroyed` does.

My preference would be that we would consistently use `NeverDestroyed` as a pattern for singletons as it's already used in other places in the code base and complies with the Google style guide.

azeey (commented):
My sincere appologies for taking so long to get back to this. Thank you so much for your patience.

After looking at this one more time, I agree that since the profiler is not enabled by default in our binary distributions, this change would be safe. However, I would ask if you could target this to `main` and we'll backport it to older branches once merged.

I'd also like to resolve one of my comments.

ntfshard:
Hi.
I'm happy that it's moving! I guess it should be re-created to main, not retargeted, I can do it.

About avoiding destructor. I believe it will be problem to enable https://github.com/google/perfetto profiler (runtime gui is not very convenient for single shot events or events with different runtime) due to it need explicit tear down operation. example: https://github.com/google/perfetto/blob/main/examples/sdk/example.cc

Main problem with static destructors I believe when we are loading and unloading library by dlopen mechanism. If they linked together and loading happens automatically by system it shouldn't be a problem.
Good approach for static objects initialization, I believe, is a explicit library API for initialization/teardown


upd: at least if it will be moved inside, we can consider update of current profiler, and it will be more convenient to see events from different part of system (I hope it will be more correct timings visualization, but I guess google's library should be better due to has much more users)

azeey:
> I'm happy that it's moving! I guess it should be re-created to main, not retargeted, I can do it.

You can rebase onto the `main` branch, force-push and retarget on the github UI. Creating a new PR is also fine. Whichever is easier.

>  If they linked together and loading happens automatically by system it shouldn't be a problem.

I don't think this is generally true. It depends on what the destructor is doing and if it tries to use other objects with static storage duration since destruction is not ordered across translation units.

If it's important for the Profiler to clean up properly, maybe we should have a separate API call for doing that and call it before terminating the program. This could just delete the statically initialized object, for example.


> upd: at least if it will be moved inside, we can consider update of current profiler, and it will be more convenient to see events from different part of system (I hope it will be more correct timings visualization, but I guess google's library should be better due to has much more users)

Not sure what you mean by moved inside. Can you elaborate?

ntfshard:
> > I'm happy that it's moving! I guess it should be re-created to main, not retargeted, I can do it.
> 
> You can rebase onto the `main` branch, force-push and retarget on the github UI. Creating a new PR is also fine. Whichever is easier.

Ok, will check it.

> 
> > If they linked together and loading happens automatically by system it shouldn't be a problem.
> 
> I don't think this is generally true. It depends on what the destructor is doing and if it tries to use other objects with static storage duration since destruction is not ordered across translation units.

Agree, but having one static object which accessing other static object from other part of library very concerning. It's very close to initialization order fiasco. In previous message I meant one level of static objects.

> 
> If it's important for the Profiler to clean up properly, maybe we should have a separate API call for doing that and call it before terminating the program. This could just delete the statically initialized object, for example.

But who will call it? It's general architecture problem. As I told before, explicit init/fint for library could help but it should be initially to avoid any confusion. In this PR I'd fix one simple thing. I still believe relying on static destructor for builded with profiler library is not a bad idea. But I won't insist.

> 
> > upd: at least if it will be moved inside, we can consider update of current profiler, and it will be more convenient to see events from different part of system (I hope it will be more correct timings visualization, but I guess google's library should be better due to has much more users)
> 
> Not sure what you mean by moved inside. Can you elaborate?

Moved inside exact translation unit. I guess I completely forgot what was the initial problem in code for this period time. Sorry for confusion

profiler/include/gz/common/Profiler.hh
```diff
@@ -101,11 +102,17 @@ namespace gz
...
+      /// \brief Get an instance of the singleton
+      public: static Profiler *Instance();
+
       /// \brief Pointer to the profiler implementation
-      private: ProfilerImpl *impl;
+      private: std::unique_ptr<ProfilerImpl> impl;
 
-      /// \brief Needed for SingletonT.
+      /// \brief Needed for SingletonT (outdated).
       private: friend class SingletonT<Profiler>;
```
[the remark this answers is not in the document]
azeey: Yeah, let's go ahead and remove this. Thanks!

azeey (approved):
LGTM! We can remove the SingletonT in this PR or a followup. It would also be nice to add another function to `Profiler` to explicitly teardown the profiler. But that can be done in a separate PR.

azeey (approved):
Yes, I'll go ahead and merge! Thanks for the patience @ntfshard!

PR #767, opened 2026-01-30
graphics: Optimize Texture Processing and Memory in AssimpLoader

azeey:
# 🦟 Bug fix

## Summary

This PR reduces system and GPU memory usage when loading meshes with shared textures (e.g., large GLB assets).

**Key Improvements:**
*   **75% RAM Saving**: Uses `L_INT8` for Metallic, Roughness, and Lightmaps instead of manual RGBA expansion.
*   **GPU Deduplication**: Implements canonical naming based on texture source (path/index) instead of material name, enabling `gz-rendering` to reuse GPU resources.
*   **Caching**: Adds image and split-map caching to `AssimpLoader` to eliminate redundant IO and pixel processing.
*   **Robustness**: Fixes a memory leak and improves `Image::ChannelData` to handle grayscale source formats.
*   **Observability**: Adds debug logging for texture cache performance.

Generated-By: Gemini 3.0
## Checklist
- [ ] Added tests
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Generated-by: Gemini 3.0

### DIFF
--- a/graphics/src/AssimpLoader.cc
+++ b/graphics/src/AssimpLoader.cc
@@ -85,12 +85,10 @@ class AssimpLoader::Implementation
   /// \param[in] _scene the assimp scene
   /// \param[in] _matIdx index of the material in the scene
   /// \param[in] _path path where the mesh is located
-  /// \param[in] _fileBaseName Base name of the mesh file.
   /// \return pointer to the converted common::Material
   public: MaterialPtr CreateMaterial(const aiScene *_scene,
                                      unsigned _matIdx,
-                                     const std::string &_path,
-                                     const std::string &_fileBaseName) const;
+                                     const std::string &_path) const;
 
   /// \brief Load a texture embedded in a mesh (i.e. for GLB format)
   /// into a gz::common::Image
@@ -366,8 +364,7 @@ void AssimpLoader::Implementation::RecursiveSkeletonCreate(const aiNode* _node,
 
 //////////////////////////////////////////////////
 MaterialPtr AssimpLoader::Implementation::CreateMaterial(
-    const aiScene *_scene, unsigned _matIdx, const std::string &_path,
-    const std::string &_fileBaseName) const
+    const aiScene *_scene, unsigned _matIdx, const std::string &_path) const
 {
   MaterialPtr mat = std::make_shared<Material>();
   aiColor4D color;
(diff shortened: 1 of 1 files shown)


graphics/src/Image.cc
```diff
@@ -707,6 +707,19 @@ FIBITMAP* Image::Implementation::SwapRedBlue(const unsigned int &_width,
 //////////////////////////////////////////////////
 std::vector<unsigned char> Image::ChannelData(Channel _channel) const
 {
+  if (!this->Valid())
+    return {};
+
+  // If the image is already 8-bit grayscale, return the data for RGB channels
```
iche033: since it's a single channel image, there're technically no RGB channels?
```suggestion
  // If the image is already 8-bit grayscale, return the the whole data buffer
```

graphics/src/AssimpLoader.cc
```diff
@@ -182,6 +198,27 @@ static std::string ToString(const aiString& str)
...
 }
 
+//////////////////////////////////////////////////
+std::string AssimpLoader::Implementation::FullTextureKey(
+    const std::string &_texturePath) const
+{
+  if (_texturePath.empty())
+    return "";
+
+  if (_texturePath[0] == '*')
```
iche033: I was wondering why the value of `texturePath.C_Str()` is something like`*0`. Looks like it's assimp internal string representation for embedded textures. Can you add a brief comment here about this check?

scpeters:
we should remember to forward-port this to `main`, though it will be hard to test until we have a Rotary collection set up

graphics/src/AssimpLoader.cc
```diff
@@ -430,25 +468,49 @@ MaterialPtr AssimpLoader::Implementation::CreateMaterial(
...
     }
-    if (roughTexture)
+    else
     {
-      pbr.SetRoughnessMap(
-          this->GenerateTextureName(_fileBaseName, _scene, assimpMat,
-          "Roughness"), roughTexture);
+      auto [texName, texData] = this->LoadTexture(_scene, texturePath,
+          this->GenerateTextureName(_fileBaseName, _scene, textureKey,
+          "MetallicRoughness"));
```
luca-della-vedova: I wonder, since for split `MetallicRoughness` maps we lookup in a separate hash map `splitMapCache`, would it make sense to add an argument to this texture that bypasses cache updates?
The idea is that we don't really care about caching the MetallicRoughness map, we only care about the split Metallic and Roughness map, and this could potentially help us save some additional memory by unloading the combined map after it's been split.
What do you think?

graphics/src/AssimpLoader.cc
azeey: Added comment in https://github.com/gazebosim/gz-common/pull/767/commits/c0854decee1add8712249068d0371606424559b3

graphics/src/AssimpLoader.cc
azeey: Makes sense. Thanks! https://github.com/gazebosim/gz-common/pull/767/commits/2edaef8814108732a1cba0b4d81547708b2e3c00

iche033 (approved):

PR #816, opened 2026-06-03
Profiler: shut down Remotery at exit to fix LeakSanitizer report

j-rivero:
## 🦟 Description

Fixes the last remaining ASAN failure in `UNIT_Profiler_Remotery_TEST` — a *nondeterministic* LeakSanitizer report of 336 bytes in 3 allocations from the Remotery worker thread. Replaces #815 (closed), whose approach turned out to be dead code.

### Root cause

`Profiler::Instance()` holds the profiler in `utils::NeverDestroyed<Profiler>`, so `RemoteryProfilerImpl`'s destructor never runs and **`rmt_DestroyGlobalInstance()` is never called**. The Remotery background worker thread therefore keeps running — and continuously allocating its own profiling samples (`Wakeup`/`ServerUpdate`/`ConsumeMessageQueue`, fresh allocations every loop iteration) — right up until process teardown.

LeakSanitizer pauses that still-running thread to scan memory and intermittently catches samples it has `malloc`'d but not yet linked into the (reachable) sample tree, reporting them as leaks. Measured failure rate locally: ~3–13% of runs, which is why CI failures looked intermittent. Instrumentation confirmed none of Remotery's destructors execute at exit, which is why the earlier #815 fix inside `SampleTree_Destructor` had no effect.

### Fix

Register an `atexit` handler in `RemoteryProfilerImpl` that destroys the Remotery instance: it joins the worker + sampler threads and releases all Remotery memory before LeakSanitizer's end-of-process check runs (atexit handlers registered after the sanitizer's own run before it). Remotery's public API already guards every entry point against a null global instance, so any profiling calls during later static destruction are safe no-ops. The vendored `Remotery.c` is untouched.

## 🧪 Test it

**Leak reproduction/elimination (ASAN build, `GZ_SANITIZER=Address`):**
- Before: `UNIT_Profiler_Remotery_TEST` failed intermittently (`336 byte(s) leaked in 3 allocation(s)`)
- After: **250/250 consecutive runs clean**; `UNIT_Profiler_Error_TEST` and `UNIT_Profiler_Disabled_TEST` also pass clean

**Functional verification that Remotery still works** (built `examples/profiler.cc` against the ASAN build):
- Profiler starts, server listens on port 17815
- WebSocket handshake succeeds (HTTP 101) and the server immediately streams profiling sample data to the connected client
- Multi-threaded profiling session (3 worker threads + main + connected viewer, ~75 s) runs without issue
- `SIGTERM` → clean exit with no hang (atexit joins the Remotery threads), port released, **zero** ASAN/LSan reports

`make codecheck` passes.

## Checklist
- [x] No ABI changes (implementation-private `.cc` change only)
- [x] All tests passing under ASAN
- [x] Functional check of the Remotery profiler (server, websocket, shutdown)

Generated-by: Claude Opus 4.8

### DIFF
--- a/profiler/src/RemoteryProfilerImpl.cc
+++ b/profiler/src/RemoteryProfilerImpl.cc
@@ -18,12 +18,43 @@
 #include "gz/common/config.hh"
 #include <gz/common/profiler/Export.hh>
 
+#include <cstdlib>
+
 #include "RemoteryProfilerImpl.hh"
 #include "gz/common/Console.hh"
 #include "gz/common/Util.hh"
 
 using namespace gz;
 using namespace common;
+
+namespace {
+  /// \brief The single active Remotery instance to shut down at process exit.
+  ///
+  /// The Profiler that owns the RemoteryProfilerImpl is a NeverDestroyed
+  /// singleton (see Profiler::Instance), so RemoteryProfilerImpl's destructor
+  /// is never run. Without an explicit shutdown the Remotery background worker
+  /// thread keeps running -- and continuously allocating its own profiling
+  /// samples -- right up until the process is torn down. LeakSanitizer pauses
+  /// that still-running thread to scan and intermittently catches samples it
+  /// has malloc'd but not yet linked into the (reachable) sample tree,
+  /// reporting them as leaks.
+  ///
+  /// Destroying the instance from an atexit handler joins the worker thread and
+  /// releases all Remotery memory before LeakSanitizer's end-of-process check
+  /// runs (atexit handlers registered after the sanitizer's own run before it),
+  /// which removes the nondeterministic leak reports.
+  Remotery *g_remoteryToShutdown = nullptr;
+
+  /// \brief atexit handler that tears down the active Remotery instance.
+  void shutdownRemotery()
+  {
+    if (g_remoteryToShutdown != nullptr)
+    {
+      rmt_DestroyGlobalInstance(g_remoteryToShutdown);
+      g_remoteryToShutdown = nullptr;
+    }
+  }
+}  // namespace
 
 GZ_COMMON_PROFILER_VISIBLE
 std::string rmtErrorToString(rmtError error) {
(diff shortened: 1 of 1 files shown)


azeey:
@ntfshard mind taking a look? I remember we had discussed the implications of using `gz::utils::NeverDestroyed` in the past.

ntfshard:
Hey. Yes, we discussed it C:

(it will be short version based on mine memory, it's quite late here already)

I guess main problem is that `atexit` approach is a basically inventing `static` objects.

Let me explain
I have a guess that _static_ `ctor`/`dtor` and `atexit` are rely on same `.init`/`.finit` sections of shared object (let it be Linux)
https://ftp.math.utah.edu/u/ma/hohn/linux/misc/elf/node3.html
Maybe `ctor` has a more interesting initialization (I saw one article, need to double-check)

So, if we are linked application with library, system loader will call code from this sections for us automatically before and after main then nobody shouldn't access it (or maybe wild dtor, but it will be symmetrical to static initialization fiasco case).

And if someone decide to try `dlopen` -- I guess it will be also called, but if someone had a pointer (code/data) to sections from unloaded library -- bad luck ☠️ 

If we really want to avoid static in all modules, we need some mechanism for all singletones in a project (hello to MeshManager).
I can imagine some parent, which track all childrens creation and calling shutdown for each of them when on a main tread we have termination sequence. Not the best idea, still too much cross-module interactions. But if we handle lifetime of this SingletonManager we can get a pseudo-static objects w/o static. (and if we didn't make untrack ability if child destroyed earlier -- very funny crashes ☠️ )
So yeah, no silver bullet, only trade-offs.

Will try to get some details tomorrow. But right now it feels like hand-made static

ntfshard:
I made a simple experiment to confirm a guess that `atexit` and `static` are actually siblings. And yes, it's the same (tech details below).

So, if you interested in my opinion, I still believe compiler-related `static` is better than `hand-made atexit static`. Surely `static` is not perfect and should be used wisely, but manual state handling has much more error potential.
Google wrote their guidelines with some specific reason and it was quite long time ago. I believe cpp runtimes became much more stable from 2011. And Gazebo is much more straightforward application than for example Chromium. 
In Chromium they seems have some event system which listen events and destroying on exit event. Or maybe I misunderstood their code.

**Tech details:**

In archive [sample.zip](https://github.com/user-attachments/files/28684942/sample.zip) a simple cpp program in 2 files. Compile it it with `g++ -c main.cpp lib.cpp` and run file in gdb.
With commands `b test::~test` `b atexit_handler` to make breakpoints and `r` to run we will be interrupted twice with stacks once in dtor, other in atexit_handler(`c` for continue). And rest of the stack is the same.
Screenshot 2026-06-08 010604

AFAIU here is a line in library: https://elixir.bootlin.com/glibc/glibc-2.39.9000/source/stdlib/exit.c#L108

j-rivero:
Thanks for the insights @ntfshard, very useful.

I don't think that we can solve all the issues you mentioned here but I changed the approach to use a RAII that handle the destruction in a more elegant way, leaving to the compiler the removal of the statics. Is this a better approach? Is there anything more that we can do to improve the situation?

azeey:
In https://github.com/gazebosim/gz-common/pull/694#pullrequestreview-3573725176, we discussed having a new API for explicitly calling to teardown the profiler. I would think that would be a better approach rather than introducing yet another static variable. For now, we can just ignore this test in the ASAN jobs.

ntfshard:
> In [#694 (review)](https://github.com/gazebosim/gz-common/pull/694#pullrequestreview-3573725176), we discussed having a new API for explicitly calling to teardown the profiler. I would think that would be a better approach rather than introducing yet another static variable. For now, we can just ignore this test in the ASAN jobs.

Or liberalization of the rule about `static` use. It's also the way.
Before https://github.com/gazebosim/gz-common/pull/694 (fix to avoid multiple instances) it was a regular static.

ntfshard:
Not sure how it should be organized, but for this new interface, maybe we need some place for design or discussions? I also concerned about maintainers' workload. It can be done in a different way(listen event, get a refcount object in scope...) and which is better and more suit design is quite discussable

azeey:
Starting a discussion on [Zulip](https://openrobotics.zulipchat.com/#narrow/channel/526040-Gazebo-General)[#Gazebo General](https://openrobotics.zulipchat.com/#narrow/channel/526040-Gazebo-General) would be a good start.

PR #817, opened 2026-06-06
Optimize texture image decoding and channel conversion

caguero:
🔨 Improvement

### Summary

This patch optimizes texture image decoding and channel conversion in `Image`, the render-thread hotspot from the [Gazebo profiling report](https://caguero.github.io/gz-profiling/2026-04-21/findings_report.pdf). This will impact loading time and potentially model spawn at runtime.

There are two changes:

(1) `Image::DataWithChannels` now converts channels in a single pass directly into the output buffer, dropping the redundant clone + intermediate + copy that `stbi__convert_format` forced (~7× faster conversion)

(2) new `Image::LoadAsRgba` / `SetFromCompressedDataAsRgba` decode straight to 8-bit RGBA, which `AssimpLoader` now uses for textures (SIMD-aligned faster decode, and `RGBAData()` becomes a passthrough). 

The output is byte-identical to the previous path, and the new methods should be ABI-safe. On a 1000-iteration, RTF=0 `jetty` run the two together cut wall-clock ~7%.

### Results

`jetty`, 1000 iterations, `<real_time_factor>0</real_time_factor>`, headless-rendering, 5 runs each:

| config | min (s) | mean (s) |
|---|---|---|
| baseline | 19.61 | 19.88 |
| + single-pass convert | 18.58 | 19.13 |
| + decode straight to RGBA | **18.24** | **18.51** |

### Test it

Build and run `UNIT_Image_TEST`, the `RGBADataChannelConversions`, `LoadAsRgba`, and `SetFromCompressedDataAsRgba` cases assert the new output is byte-identical to the previous decode-then-convert path for PNG/JPEG and all channel counts.

### Checklist

- [x] Added tests
- [ ] codecheck passed
- [ ] All tests passed
- [x] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Generated-by: Claude Opus 4.8

### DIFF
--- a/graphics/src/Image.cc
+++ b/graphics/src/Image.cc
@@ -185,17 +185,18 @@ int Image::Load(const std::string &_filename)
 
 //////////////////////////////////////////////////
 int Image::Load(const std::string &_filename,
-                std::optional<PixelFormatType> _type)
+                std::optional<PixelFormatType> _outputFormat)
 {
   // No target format requested: load in the source's native format.
-  if (!_type.has_value())
+  if (!_outputFormat.has_value())
     return this->Load(_filename);
 
-  if (_type.value() != RGBA_INT8)
+  if (_outputFormat.value() != RGBA_INT8)
   {
-    gzerr << "Unsupported target pixel format[" << _type.value() << "] for ["
-          << _filename << "]; only RGBA_INT8 (or std::nullopt for the native "
-          << "format) is supported.\n";
+    gzerr << "Unsupported target pixel format["
+          << _outputFormat.value() << "] for [" << _filename
+          << "]; only RGBA_INT8 (or std::nullopt for the native format) is "
+          << "supported.\n";
     return -1;
   }
 
@@ -401,22 +402,22 @@ void Image::SetFromCompressedData(unsigned char *_data,
 //////////////////////////////////////////////////
 void Image::SetFromCompressedData(const unsigned char *_data,
                                   unsigned int _size,
-                                  Image::PixelFormatType _format,
-                                  std::optional<PixelFormatType> _type)
+                                  Image::PixelFormatType _inputFormat,
+                                  std::optional<PixelFormatType> _outputFormat)
 {
   // No target format requested: decode to the source's native format. The
   // 3-argument overload does not modify _data (stb takes it as const), so the
   // const_cast is safe.
-  if (!_type.has_value())
+  if (!_outputFormat.has_value())
   {
     this->SetFromCompressedData(const_cast<unsigned char *>(_data), _size,
-        _format);
+        _inputFormat);
     return;
   }
 
-  if (_type.value() != RGBA_INT8)
+  if (_outputFormat.value() != RGBA_INT8)
   {
-    gzerr << "Unsupported target pixel format[" << _type.value()
+    gzerr << "Unsupported target pixel format[" << _outputFormat.value()
           << "]; only RGBA_INT8 (or std::nullopt for the native format) is "
           << "supported.\n";
     return;
@@ -426,9 +427,9 @@ void Image::SetFromCompressedData(const unsigned char *_data,
     stbi_image_free(this->dataPtr->bitmap);
   this->dataPtr->bitmap = nullptr;
 
-  if (_format != COMPRESSED_PNG && _format != COMPRESSED_JPEG)
+  if (_inputFormat != COMPRESSED_PNG && _inputFormat != COMPRESSED_JPEG)
   {
-    gzerr << "Unable to handle format[" << _format << "]\n";
+    gzerr << "Unable to handle format[" << _inputFormat << "]\n";
     return;
   }
 
--- a/graphics/src/Image.cc
+++ b/graphics/src/Image.cc
@@ -73,10 +73,10 @@ namespace gz
       public: std::string fullName;
 
       /// \brief Width of the image
-      public: int width;
+      public: int width{0};
 
       /// \brief Height of the image
-      public: int height;
+      public: int height{0};
 
       /// \brief the number of channels per pixel
       ///
@@ -85,10 +85,10 @@ namespace gz
       ///       2           grey, alpha
       ///       3           red, green, blue
       ///       4           red, green, blue, alpha
-      public: int channels;
+      public: int channels{0};
 
       /// \brief the number of bits per pixel
-      public: int bits_per_channel;
+      public: int bits_per_channel{0};
 
       /// \brief Converts bitmap data to the given number of channels
       public: std::vector<unsigned char> DataWithChannels(int out_channels)
@@ -200,7 +200,7 @@ int Image::LoadAsRgba(const std::string &_filename)
 
   if (this->dataPtr->bitmap)
     stbi_image_free(this->dataPtr->bitmap);
-  this->dataPtr->bitmap = NULL;
+  this->dataPtr->bitmap = nullptr;
 
   // Decode straight to 8-bit RGBA in a single pass (req_comp = 4). This is
   // faster than decoding to native channels and converting afterwards -- stb's
@@ -211,7 +211,7 @@ int Image::LoadAsRgba(const std::string &_filename)
   int h;
   int n;
   void *bitmap = stbi_load(this->dataPtr->fullName.c_str(), &w, &h, &n, 4);
-  if (bitmap == NULL)
+  if (bitmap == nullptr)
   {
     gzerr << "Failed to load file [" << this->dataPtr->fullName
           << "]: " << stbi_failure_reason() << std::endl;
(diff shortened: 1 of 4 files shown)


graphics/src/Image.cc
```diff
@@ -182,6 +182,50 @@ int Image::Load(const std::string &_filename)
...
+  if (!exists(this->dataPtr->fullName))
+  {
+    gzerr << "Unable to open image file[" << this->dataPtr->fullName
+          << "], check your GZ_RESOURCE_PATH settings.\n";
+    return -1;
+  }
+
+  if (this->dataPtr->bitmap)
+    stbi_image_free(this->dataPtr->bitmap);
+  this->dataPtr->bitmap = NULL;
```
iche033: good time to switch from `NULL` to `nullptr`?
```suggestion
  this->dataPtr->bitmap = nullptr;
```

There are other new code added in this PR that uses `NULL`

graphics/src/Image.cc
```diff
@@ -348,52 +427,143 @@ int Image::Pitch() const
...
-      gzerr << "Error converting image to " << out_channels << " channels\n";
-      return std::vector<unsigned char>();
-    }
-    data =
-        this->DataImpl(bitmap_rgb, this->width * this->height * out_channels *
-                                       this->bits_per_channel / 8);
-    STBI_FREE(bitmap_rgb);
+    case 2 * 8 + 1:
+      for (std::size_t p = 0; p < _npix; ++p, _src += 2, _dst += 1)
+      { _dst[0] = _src[0]; }
```
iche033: minor suggestion on coding style.
```suggestion
      {
        _dst[0] = _src[0];
      }
```
same for the lines below.

graphics/src/Image.cc
```diff
@@ -340,6 +384,41 @@ void Image::SetFromCompressedData(unsigned char *_data,
   this->dataPtr->channels = n;
 }
 
+//////////////////////////////////////////////////
+void Image::SetFromCompressedDataAsRgba(unsigned char *_data,
```
iche033: I think `_data` can be const?
```suggestion
void Image::SetFromCompressedDataAsRgba(const unsigned char *_data,
```

graphics/src/Image.cc
```diff
@@ -348,52 +427,143 @@ int Image::Pitch() const
...
+/// \param[out] _dst Destination buffer (_npix * _outCh samples).
+/// \param[in] _outCh Number of channels to write.
+/// \param[in] _npix Number of pixels to convert.
+/// \param[in] _aMax Opaque alpha to insert (255 for 8-bit, 0xffff for 16-bit).
+/// \return False if the channel combination is unsupported.
+template <typename T>
+bool ConvertChannels(const T *_src, int _inCh, T *_dst, int _outCh,
+    std::size_t _npix, T _aMax)
+{
+  switch (_inCh * 8 + _outCh)
```
iche033, line 499: can you provide a brief comment on this? I think what this does is that it attempts to encode the 2 in / out channels into a single unique integer so that can be handled in one switch statement?

graphics/src/Image.cc
```diff
@@ -348,52 +427,143 @@ int Image::Pitch() const
...
-  else
+  return true;
+}
+
+}  // namespace
+
+std::vector<unsigned char>
+Image::Implementation::DataWithChannels(int out_channels) const
+{
+  const size_t outSize = static_cast<size_t>(this->width) * this->height *
```
iche033, line 598: I just noticed that these variables (`this->width`, `this->height`, `this->bits_per_channel`, `this->channel`) could be used without being initialized. Can initialize them here: https://github.com/gazebosim/gz-common/blob/cfa5c24e6210fd55cb3c9d29e7737f76464c7b40/graphics/src/Image.cc#L76-L91

e.g.

```c++
public: int width{0};
public: int height{0};
public: int channels{0};
public: int bits_per_channel{0};
```

iche033: We can also add a check to see if width / height is 0 or if bitmap is nullptr, if so, return from this function before doing any work

graphics/src/Image.cc
caguero: Updated in [8f8c17f](https://github.com/gazebosim/gz-common/pull/817/commits/8f8c17f9977da2c2f33cc120dfd560cbab26470b)

graphics/src/Image.cc
caguero: Updated in [8f8c17f](https://github.com/gazebosim/gz-common/pull/817/commits/8f8c17f9977da2c2f33cc120dfd560cbab26470b)

graphics/src/Image.cc
caguero: Updated in [8f8c17f](https://github.com/gazebosim/gz-common/pull/817/commits/8f8c17f9977da2c2f33cc120dfd560cbab26470b)

graphics/src/Image.cc
caguero: Added in [8f8c17f](https://github.com/gazebosim/gz-common/pull/817/commits/8f8c17f9977da2c2f33cc120dfd560cbab26470b). That's exactly right.

graphics/src/Image.cc
caguero: See [8f8c17f](https://github.com/gazebosim/gz-common/pull/817/commits/8f8c17f9977da2c2f33cc120dfd560cbab26470b)

iche033 (approved):

luca-della-vedova (commented):
Nice perf improvement!
I guess the new API is to make sure this is trivial to backport?

I wonder if we could tune the API in a way that allows us (in rotary) to only have a single API for loading images going forward, for example I imagine the ideal API would look something like:

```
int Load(const std::string &_filename, const std::optional<ImagePixelFormatType> _type = std::nullopt);
```

Then users could call `Load(filename)` and it would just load as it does now, `Load(filename, PixelFormatType::RGBA_INT8)` to process an RGBA8, or any additional format we might need in the future.

Now my understanding is that adding a new parameter breaks ABI so this is not backportable, but I wonder if we could take a two step approach:

1) Add a `int Load(const std::string &_filename, const std::optional<ImagePixelFormatType> _type);` function. Since the type is not `std::nullopt` it will be easy for the compiler to know which overload to call. This is straightforward to backport.
2) In main only where we can break ABI, change it to have a `std::nullopt` default argument and merge both `Load` into a single function. This shouldn't break API thanks to the default parameter but should allow us to have a single implementation.

Does this sound reasonable?

caguero:
> Does this sound reasonable?

I think it's a good idea. I implemented only step 1 in 4ba5d9d. I guess we'll do the backport and then a separate PR with step 2. 

Is this what you had in mind?

luca-della-vedova (approved):
LGTM from an API standpoint and with green CI, just a minor nit

caguero:
> LGTM from an API standpoint and with green CI, just a minor nit

Code updated in [f612eab](https://github.com/gazebosim/gz-common/pull/817/commits/f612eab4bc3fae4015e1f5de426947e8df531418)

caguero:
@mergifyio backport gz-common7 gz-common6 gz-common5 ign-common4

caguero:
> I wonder if we could tune the API in a way that allows us (in rotary) to only have a single API for loading images going forward, for example I imagine the ideal API would look something like:
> 
> ```
> int Load(const std::string &_filename, const std::optional<ImagePixelFormatType> _type = std::nullopt);
> ```
> > int Load(const std::string &_filename, const std::optional<ImagePixelFormatType> _type = std::nullopt);
> ```
> 
> Then users could call `Load(filename)` and it would just load as it does now, `Load(filename, PixelFormatType::RGBA_INT8)` to process an RGBA8, or any additional format we might need in the future.
> 
> Then users could call `Load(filename)` and it would just load as it does now, `Load(filename, PixelFormatType::RGBA_INT8)` to process an RGBA8, or any additional format we might need in the future.
> 
> Now my understanding is that adding a new parameter breaks ABI so this is not backportable, but I wonder if we could take a two step approach:
> 
> 1. Add a `int Load(const std::string &_filename, const std::optional<ImagePixelFormatType> _type);` function. Since the type is not `std::nullopt` it will be easy for the compiler to know which overload to call. This is straightforward to backport.
> 2. In main only where we can break ABI, change it to have a `std::nullopt` default argument and merge both `Load` into a single function. This shouldn't break API thanks to the default parameter but should allow us to have a single implementation.
> 
> Does this sound reasonable?

I'm having second thoughts about this. If we converge into a single `Load()` function, users will always be able to use the single argument option, which isn't fully optimized. If we follow the classical approach with two functions (the old one with one parameter and the new one with two required parameters), we can deprecate the old function and eventually remove it, in favor of the fully optimized path.

What do people think?

luca-della-vedova:
> I'm having second thoughts about this. If we converge into a single `Load()` function, users will always be able to use the single argument option, which isn't fully optimized. If we follow the classical approach with two functions (the old one with one parameter and the new one with two required parameters), we can deprecate the old function and eventually remove it, in favor of the fully optimized path.
> 
> What do people think?

So users are forced to change their `Load(path)` into a `Load(path, std::nullopt)` if they want to maintain the old behavior and not have compile deprecation / warnings going forward?
I guess the deprecation warning could also help users think a bit whether they know what format they need and they can migrate into anything other than `std::nullopt` and have a performance gain?
I'm personally fairly neutral on this, both ways have pros and cons (the deprecation is a bit more churn for users but also potential for performance improvements).

PR #830, opened 2026-06-15
Improve COLLADA loader (supersedes #569) - Part 1/3

caguero:
**Supersedes:** #569 (the original "improved Collada loader" work).
Credit to @fredmarkus

This patch is the first of three PRs that take over #569. This one lands #569 on `main` and fixes the correctness regressions.

#569's problem was that it was inflating meshes.
Vertex/normal counts before vs after (correct count = the pre-#569 baseline):

| Mesh | pre-#569 (correct) | #569 as-is | This PR |
|---|---|---|---|
| sydney_regatta (17 MB) | 109,588 ✅ | **150,776 (+37.6%) ❌** | 109,588 ✅ |
| lrauv base (10.8 MB) | 363,199 ✅ | **363,624 ❌** | 363,199 ✅ |

## Backport Policy

    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [x] Added tests
- [x] `codecheck` passed
- [x] All tests passed

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit
message to match the summary while retaining `Signed-off-by` messages, and keep
the co-author trailer below so the original author is credited.

Co-authored-by: Frederik Markus <frederik.anilmarkus@gmail.com>


luca-della-vedova (commented):
Before this lands I'd prefer to have a benchmark to have a quantitative before / after test. The way to test this mentioned in #569 doesn't work because:

1) The mesh provided as a test case is a google drive link that doesn't work anymore
2) The process would still be very manual.

@cheriehu can you come up with a benchmark to test the performance of loading meshes in this library? The unit tests of the various loaders can be a starting point into what the API to load files is like, in a first iteration we could just use the existing the more complex / larger existing test meshes.
This would also help us make a more informed decision into whether moving to `assimp` as discussed in the PMC would impact load time performance in any way

caguero:
> Before this lands I'd prefer to have a benchmark to have a quantitative before / after test. The way to test this mentioned in #569 doesn't work because:
> 
> 1. The mesh provided as a test case is a google drive link that doesn't work anymore
> 2. The process would still be very manual.

This work is splitted in three PRs. I've included some benchmarking info in the PR 3/3 (#832).

cheriehu:
I added a benchmark here (#834) and can also replicate performance improvements for loading the collada mesh on #832 compared to main

test/data/xy_square_lines.dae
shameekganguly, line 1: Can we rename this to xy_square_lines.dae since it doesn't actually set up a 3D box?

test/data/box_texture.dae
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0" encoding="utf-8"?>
+<COLLADA xmlns="http://www.collada.org/2005/11/COLLADASchema" version="1.4.1">
+  <asset>
+    <unit name="meter" meter="1"/>
+    <up_axis>Z_UP</up_axis>
+  </asset>
+  <library_images>
+    <image id="tex0" name="tex0">
+      <init_from>box_texture.png</init_from>
```
shameekganguly: Are we missing this texture file? If it is not needed for the tests, can we remove this?

test/data/xy_triangle_texture.dae
shameekganguly, line 1: This doesn't seem to have a box, but rather a single triangle? Can we rename to xy_triangle_texture.dae?

graphics/src/ColladaLoader_TEST.cc
```diff
@@ -522,3 +522,101 @@ TEST_F(ColladaLoader, NoAnimName)
...
+  ASSERT_NE(nullptr, mat);
+
+  // The diffuse texture chain should have been resolved and stored.
+  EXPECT_FALSE(mat->TextureImage().empty());
+  EXPECT_NE(mat->TextureImage().find("box_texture.png"), std::string::npos);
+  delete mesh;
+}
+
+/////////////////////////////////////////////////
+// Loading a non-existent file must return null, not crash.
```
shameekganguly, line 552: For this test case and below, please document these behaviors for different input cases in a docstring for `ColladaLoader::Load`: https://github.com/gazebosim/gz-common/blob/main/graphics/include/gz/common/ColladaLoader.hh#L44

graphics/src/ColladaLoader_TEST.cc
```diff
@@ -522,3 +522,101 @@ TEST_F(ColladaLoader, NoAnimName)
   EXPECT_EQ(animName, "animation1");
   delete mesh;
 }
+
+/////////////////////////////////////////////////
+TEST_F(ColladaLoader, LoadLines)
+{
+  common::ColladaLoader loader;
+  common::Mesh *mesh = loader.Load(
```
shameekganguly: nit: here and elsewhere in this file, can we wrap the mesh in a std::unique_ptr instead of manually having to delete it at the end of each test case?

graphics/src/ColladaLoader.cc
```diff
@@ -2442,7 +2576,7 @@ void ColladaLoader::Implementation::LoadTriangles(
 
   // vertexIndexMap is a map of collada vertex index to Gazebo submesh vertex
   // indices, used for identifying vertices that can be shared.
-  std::map<unsigned int, std::vector<GeometryIndices> > vertexIndexMap;
+  std::unordered_map<unsigned int, std::vector<GeometryIndices>> vertexIndexMap;
 
   std::vector<unsigned int> values(offsetSize);
   std::vector<std::string> strs = split(pStr, " \t\r\n");
```
shameekganguly, line 2584: As another minor optimization, we can pre-allocate an upper bound for `vertexIndexMap` here:
```
if (offsetSize > 0)
  vertexIndexMap.reserve(strs.size() / offsetSize);
```

graphics/src/ColladaLoader.cc
shameekganguly, line 2215: As another minor optimization, we can pre-allocate an upper bound for `vertexIndexMap` here:
```
if (inputSize > 0)
  vertexIndexMap.reserve(strs.size() / inputSize);
```

shameekganguly (changes requested):
Suggested some minor perf, documentation and readability improvements.

test/data/xy_square_lines.dae
caguero: Renamed in 0d596e23.

test/data/box_texture.dae
caguero: The texture reference is needed by the LoadTextureMaterial test, but the file itself was missing. Added a small texture file in 0d596e23.

test/data/xy_triangle_texture.dae
caguero: Renamed in 0d596e23.

graphics/src/ColladaLoader_TEST.cc
caguero: Documented in 0d596e23.

graphics/src/ColladaLoader_TEST.cc
caguero: Modified in 0d596e23. The tests added in parts 2/3 were updated the same way in 901bd31c and 4c55d9c7.

graphics/src/ColladaLoader.cc
caguero: Added in 0d596e23.

graphics/src/ColladaLoader.cc
caguero: Added in 0d596e23. Note that in part 3/3 this code is replaced and the reserve is based on the position count instead.

caguero:
> Suggested some minor perf, documentation and readability improvements.

Thanks! Good to go

shameekganguly (approved):
Thanks, LGTM!

caguero:
@Mergifyio backport gz-common7 gz-common6 gz-common5 ign-common4

PR #831, opened 2026-06-16
More robust COLLADA loader - Part 2/3

caguero:
Built on top of #830.

This PR makes the COLLADA loader more robust and tolerant to malformed data.

## Backport Policy

    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

## Checklist
- [x] Added tests
- [x] `codecheck` passed
- [x] All tests passed

**Note to maintainers**: Remember to use **Squash-Merge** and retain
`Signed-off-by` messages.


luca-della-vedova (commented):
I'm lowkey worried this might conflict with the effort in https://github.com/gazebosim/gz-common/pull/845.

Specifically, we are trying to unify all the loading tests from being loader specific to being just at the MeshManager level, then we can run the same suite of tests with the flag to force usage of assimp set to true or false and test both code paths.
My worry here is how assimp behaves with all these new test assets, if it behaves the same great but if it doesn't should that be considered a blocker? How many of these assets are part of, or inspired from, real assets in the wild and how many are specially crafted to test the robustness of the `ColladaLoader`?
Also the tests are very much `ColladaLoader` specific, they test that the console logging messages contain specific messages. This just won't work when we use an external library where we cannot control logs, so the only type of testing we could do there is "load make sure it doesn't crash the system".

I don't have a clear recommendation, I guess ideally we would manage to get the loaders unification in first, then if this PR's new unit tests targeted that refactor we would be able to run them with both loaders and see if there is any difference. Alternatively we could manually merge the two and see how assimp behaves, or something else?

caguero:
> I'm lowkey worried this might conflict with the effort in #845.
> 
> Specifically, we are trying to unify all the loading tests from being loader specific to being just at the MeshManager level, then we can run the same suite of tests with the flag to force usage of assimp set to true or false and test both code paths. My worry here is how assimp behaves with all these new test assets, if it behaves the same great but if it doesn't should that be considered a blocker? How many of these assets are part of, or inspired from, real assets in the wild and how many are specially crafted to test the robustness of the `ColladaLoader`? Also the tests are very much `ColladaLoader` specific, they test that the console logging messages contain specific messages. This just won't work when we use an external library where we cannot control logs, so the only type of testing we could do there is "load make sure it doesn't crash the system".
> 
> I don't have a clear recommendation, I guess ideally we would manage to get the loaders unification in first, then if this PR's new unit tests targeted that refactor we would be able to run them with both loaders and see if there is any difference. Alternatively we could manually merge the two and see how assimp behaves, or something else?

We can definitely tweak the tests if assimp cannot handle any of the corner cases that we're covering here. In particular, the part where we're checking the error message could be relaxed for sure. Do you want to load the test assets and let me know which ones don't make assimp behave properly. That's exactly the spirit of the PR but we can tweak it to unblock #845.

luca-della-vedova:
I wonder if an alternate approach could have a _minimal_ version of #845 merged where we just create the test infrastructure to test a specific mesh with both loaders and perhaps one migrated test but defer migrating all the other tests (that is the bulk of the work) to later. Then we can update this PR to add the tests to the new MeshManager unit test file so they are immediately tested against both loader implementations and we don't need to test them manually / migrate them later, what do you think?

cheriehu:
> I wonder if an alternate approach could have a _minimal_ version of #845 merged where we just create the test infrastructure to test a specific mesh with both loaders and perhaps one migrated test but defer migrating all the other tests (that is the bulk of the work) to later.

Would this involve just merging the load box/cube tests for each of the file types first (without tests for texture/animation etc.)? From what I can tell there’s LoadBox (for Collada), LoadObjBox, LoadSTL, LoadGlTF2Box (for glb), and LoadFbxBox. 

Or maybe we could just merge all the tests that are not giving any problems (i.e. all tests that are identical between our custom loaders and AssimpLoader)

I'll create a new PR for this

luca-della-vedova:
> Would this involve just merging the load box/cube tests for each of the file types first (without tests for texture/animation etc.)? From what I can tell there’s LoadBox (for Collada), LoadObjBox, LoadSTL, LoadGlTF2Box (for glb), and LoadFbxBox.
> 
> Or maybe we could just merge all the tests that are not giving any problems (i.e. all tests that are identical between our custom loaders and AssimpLoader)
> 
> I'll create a new PR for this

Yea merging all the ones that are identical sounds good to me, then we leave the other ones as a followup. Thank you!

cheriehu:
The other PR with the more straightforward tests is here #862

luca-della-vedova:
Alright! Now that #862 is merged could you update this PR to add tests to the `MeshManager` instead testing both loader implementations? So we can see where we are at, maybe we won't actually need to change much other than the logger expectations.

caguero:
> Alright! Now that #862 is merged could you update this PR to add tests to the `MeshManager` instead testing both loader implementations? So we can see where we are at, maybe we won't actually need to change much other than the logger expectations.

Sounds good! I just updated the PR.

cheriehu:
I think the tests look ok to me. I’m wondering if after this PR is merged we should align the behavior for the cases where the AssimpLoader and ColladaLoader differ (ie LoadEmptyInitFrom and LoadMalformedPositionHugeCount)?

luca-della-vedova:
> I think the tests look ok to me. I’m wondering if after this PR is merged we should align the behavior for the cases where the AssimpLoader and ColladaLoader differ (ie LoadEmptyInitFrom and LoadMalformedPositionHugeCount)?

Thanks for reviewing! I'll leave it to @caguero to pitch in for this, IMHO if those type of assets exist in the wild we could consider it but if they are explicitly malformed assets to test that the loader is robust the behavior of "don't crash" is good enough

luca-della-vedova (approved):

caguero:
@mergifyio backport gz-common7 gz-common6 gz-common5 ign-common4

PR #833, opened 2026-06-16
Improve performance on STL loader (ASCII)

caguero:
This patch improves the performance of our custom STL loader parsing class.

`STLLoader::ReadAscii` called `SubMesh::IndexOfVertex()`, a linear scan over all vertices so far, once per vertex, making ASCII STL loading **O(n²)**. This replaces it with a `vertex → first-index` hash map (the same technique the ColladaLoader uses), giving an **O(n)** load.

The output is unchanged: every vertex/normal is still appended (so `VertexCount` is identical) and repeated vertices still reuse the index of their first occurrence.

**Only the ASCII path is affected**; the binary path was already O(n).

## Performance results

Synthetic ASCII STL files, full load time:

| Triangles (vertices) | Before | After | Speedup |
|---|---|---|---|
| 12k (36,000 v) | 412 ms | 36 ms | **11×** |
| 24k (72,000 v) | 1644 ms | 72 ms | **23×** |
| 48k (144,000 v) | 6298 ms | 186 ms | **34×** |

The "before" times quadruple as triangles double (O(n²)); the "after" times  scale linearly. Vertex/normal/index counts are identical before and after (e.g. `cube.stl` → 36 vertices in both), so the mesh output is unchanged.

## Checklist
- [ ] Added tests (behavior unchanged; covered by the existing `STLLoader_TEST`)
- [x] `codecheck` passed
- [x] All tests passed

**Note to maintainers**: Remember to use **Squash-Merge** and retain
`Signed-off-by` messages.


graphics/src/STLLoader.cc
```diff
@@ -145,7 +168,19 @@ bool STLLoader::ReadAscii(FILE *_filein, Mesh *_mesh)
...
+        if (vertexIndexIt != vertexIndex.end())
+        {
+          subMesh.AddIndex(vertexIndexIt->second);
+        }
+        else
+        {
+          unsigned int newIndex = subMesh.VertexCount() - 1;
+          vertexIndex[vertex] = newIndex;
+          subMesh.AddIndex(newIndex);
+        }
```
luca-della-vedova: First of all, I do agree that the previous implementation looks off but I'm also not sure this change is correct.
Specifically, what seems sketchy in the previous implementation is that we _always_ add a vertex, regardless whether it exists then we do a full linear scan when we could just... take the index of the last item in the vector since we just pushed to it? We could do what this PR is doing with one line of code change, no new hash maps / lookups / branching / dynamic allocations by just changing:

```
subMesh.AddVertex(vertex);
subMesh.AddIndex(subMesh(IndexOfVertex(vertex));
```

To

```
subMesh.AddVertex(vertex);
subMesh.AddIndex(subMesh.VertexCount() - 1);
```

Since again, we are just naively adding vertex at every iteration, which will be pushed to the end of the vector.
I tested this locally and it "preserves behavior" as in it passes CI.

Now the issue with "preserving behavior", it seems the unit test was actually written _after_ the loader was created, and I suspect it was written to make sure "the current implementation would pass". For this reason it expects a cube to have [36 vertices of which "24 unique, 12 shared"](https://github.com/gazebosim/gz-common/blob/758fa1241b1376c4bb87cb2654c472414e696027/graphics/src/STLLoader_TEST.cc#L45).
This "24 unique 12 shared" looks oddly familiar, the [same comment](https://github.com/gazebosim/gz-common/blob/758fa1241b1376c4bb87cb2654c472414e696027/graphics/src/ColladaLoader_TEST.cc#L42) is in the ColladaLoader, where the test actually expects to see 24 vertices (that also looks off, but something for another day), while the STL test actually expects 36 vertices.

But again in an implementation that has a hash map to reuse vertices that comment shouldn't really apply anymore. Why would a cube have more than 8 vertices if we are adding logic to reuse existing vertices?
I tried to indeed make sure that we reuse vertices by applying this diff and now it returns 8 vertices.

```patch
diff --git a/graphics/src/STLLoader.cc b/graphics/src/STLLoader.cc
index 479d3646..15172141 100644
--- a/graphics/src/STLLoader.cc
+++ b/graphics/src/STLLoader.cc
@@ -166,7 +166,6 @@ bool STLLoader::ReadAscii(FILE *_filein, Mesh *_mesh)
         vertex.Y(r2);
         vertex.Z(r3);
 
-        subMesh.AddVertex(vertex);
         subMesh.AddNormal(normal);
         // Reuse the first index for a repeated vertex, matching the previous
         // IndexOfVertex() behavior but without the linear scan.
@@ -177,6 +176,7 @@ bool STLLoader::ReadAscii(FILE *_filein, Mesh *_mesh)
         }
         else
         {
+          subMesh.AddVertex(vertex);
           unsigned int newIndex = subMesh.VertexCount() - 1;
           vertexIndex[vertex] = newIndex;
           subMesh.AddIndex(newIndex);
diff --git a/graphics/src/STLLoader_TEST.cc b/graphics/src/STLLoader_TEST.cc
index 15189896..d53e559f 100644
--- a/graphics/src/STLLoader_TEST.cc
+++ b/graphics/src/STLLoader_TEST.cc
@@ -42,8 +42,8 @@ TEST_F(STLLoaderTest, LoadSTL)
   EXPECT_STREQ("unknown", mesh->Name().c_str());
   EXPECT_EQ(math::Vector3d(20, 0, 20), mesh->Max());
   EXPECT_EQ(math::Vector3d(0, -20, 0), mesh->Min());
-  // 36 vertices, 24 unique, 12 shared.
-  EXPECT_EQ(36u, mesh->VertexCount());
+  // 8 unique vertices
+  EXPECT_EQ(8u, mesh->VertexCount());
   EXPECT_EQ(36u, mesh->NormalCount());
   EXPECT_EQ(36u, mesh->IndexCount());
   EXPECT_EQ(0u, mesh->TexCoordCount());
```

So overall I think we should either change this to a naive one-liner if we want to preserve behavior and improve performance (first proposal), or have a full shared vertex mechanism that will change behavior but drop the number of vertices (second proposal).

luca-della-vedova (changes requested):

graphics/src/STLLoader.cc
caguero: I think you're right. In fact that seems to be the strategy for parsing the binary STL as well. I'm leaning towards preserving behavior with the trivial update. [4ad7062](https://github.com/gazebosim/gz-common/pull/833/commits/4ad7062de21bac38c9e5b17b4377c14cb9663157)

luca-della-vedova (approved):
Nice, thanks for the cleanup into shared function + test update as well!

caguero:
@Mergifyio backport gz-common7 gz-common6 gz-common5 ign-common4

PR #844, opened 2026-07-10
Change AssimpLoader behaviour when dealing with textures

cheriehu:
# 🦟 Bug fix

#### 3. Change AssimpLoader behavior when loading `bump` and `map_Bump` fields in .mtl files 
They are loaded as `aiTextureType_HEIGHT` not `aiTextureType_NORMALS`

#### 4. Change AssimpLoader behavior when when loading `map_Ns` field in .mtl files
(edited)
It's loaded as `aiTextureType_SHININESS` which IS not read by our AssimpLoader before.
4a. Assign `aiTextureType_SPECULAR` to Pbr->SpecularMap

#### 5. Blender-specific behaviors not supported by assimp
Blender puts the roughness map in what Assimp reads as `aiTextureType_SHININESS`, and the metalness map into `refl` which Assimp doesn't read at all.
Added warnings if there is an OBJ loaded with PBR to recommend users to move to GLTF instead since their PBR might not be loaded by AssimpLoader in the way the expect. 

#### 6. Change AssimpLoader behavior when assigning texture names to texture maps
I've changed the AssimpLoader to use the `texturePath` if the texture is not embedded and the `textureName` if it is (which was the behaviour before).

For non-embedded textures the name assigned used to be (for example) something like `/$HOME/gz_ws/src/gz-common/test/data/gltf/PurpleCube_Diffuse.png_Diffuse`. So now the name of the texture should be something like `PurpleCube_Diffuse.png` instead.

## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [X] Added tests

Assisted-by: Gemini 3.1 Pro


luca-della-vedova (commented):
@cheriehu is this ready for review?

cheriehu:
> @cheriehu is this ready for review?

Sorry, not yet. There are some changes from #845 I wanted to put here and also more differences in the AssimpLoader behavior that I haven't fully resolved yet. 

I'll move the changes I've already made to this PR first.

luca-della-vedova (commented):
To have a balance between "endless PRs" and "one PR to rule them all" I wonder if we can just split this in two? Specifically between fairly straightforward and non controversial fixes:

1) A very straightforward + backportable ifdef removal.
2) A generally straightforward assimp default material check.
3) The gltf lowercase extension change.

On the other hand these should use a bit more thorough review:

1) The new maps parsed into normals / metalness for OBJ PBR (need to run a fresh grep through assimp to see if there is any risk of the wrong map being assigned).
2) The name changes for the texture keys (to make sure it doesn't break anything).
3) The warning with obj PBR (not sure we should backport a new warning).

cheriehu:
> To have a balance between "endless PRs" and "one PR to rule them all" I wonder if we can just split this in two? Specifically between fairly straightforward and non controversial fixes:
> 
> 1. A very straightforward + backportable ifdef removal.
> 2. A generally straightforward assimp default material check.
> 3. The gltf lowercase extension change.

Thanks for the suggestion, the more straightforward changes are now in #858

luca-della-vedova:
> > To have a balance between "endless PRs" and "one PR to rule them all" I wonder if we can just split this in two? Specifically between fairly straightforward and non controversial fixes:
> > 
> > 1. A very straightforward + backportable ifdef removal.
> > 2. A generally straightforward assimp default material check.
> > 3. The gltf lowercase extension change.
> 
> Thanks for the suggestion, the more straightforward changes are now in #858

Merged, if you merge main / address conflicts the diff should now be more manageable

graphics/src/AssimpLoader.cc
```diff
@@ -870,7 +889,27 @@ Mesh *AssimpLoader::Load(const std::string &_filename)
...
+    // Warn if file is an OBJ file with real PBR information
+    common::Pbr defaultPbr;
+    if (extension == "obj" && *mat->PbrMaterial() != defaultPbr)
+    {
+      gzwarn << "OBJ file with PBR materials detected. Note that the fields in "
+             << "the .mtl file may not be read as expected "
+             << "(eg. if you exported this file with Blender). "
+             << "Use GLTF for a more modern format that supports PBR."
+             << std::endl;
+    }
```
luca-della-vedova, line 946: Let's add the path of the mesh that generated this warning to help users debug this case

cheriehu:
not too sure if [757d010](https://github.com/gazebosim/gz-common/pull/844/commits/757d010963e42044850f95623aff634a6f4fb369) (consistently using`math::` instead of `gz::math::` in MeshManager_TEST.cc) belongs in this PR

graphics/src/MeshManager_TEST.cc
```diff
@@ -77,15 +77,15 @@ TEST_F(MeshManager, CreateExtrudedPolyline)
   // test extrusion of a path with two subpaths:
   // a smaller square inside a bigger square.
   // The smaller square should be treated as a hole inside the bigger square.
-  std::vector<std::vector<gz::math::Vector2d> > path;
-  std::vector<gz::math::Vector2d> subpath01;
+  std::vector<std::vector<math::Vector2d> > path;
+  std::vector<math::Vector2d> subpath01;
```
luca-della-vedova: Can we remove all the spurious changes in this file that are only about removing the `gz` namespace to make the diff a bit more manageable?

graphics/src/AssimpLoader_TEST.cc
```diff
@@ -224,28 +224,84 @@ TEST_F(AssimpLoader, LoadCylinderAnimatedFrom3dsMax)
 }
 
 /////////////////////////////////////////////////
-// Open a gltf mesh with an external texture
-TEST_F(AssimpLoader, LoadGlTF2BoxExternalTexture)
+// This tests opening an OBJ file that has PBR fields
+TEST_F(AssimpLoader, PBR)
```
luca-della-vedova: Since this will probably be as far as we get with PBR support in Assimp maybe we can just merge this test in `MeshManager` and guard the expectation behind and `if (forceAssimp)` statement?

graphics/src/MeshManager_TEST.cc
cheriehu: [937371a](https://github.com/gazebosim/gz-common/pull/844/commits/937371acff46cde5910cb1eb220431b7ae9f53e3)

graphics/src/AssimpLoader_TEST.cc
cheriehu: 812caf0

graphics/src/AssimpLoader.cc
cheriehu: [f7cc8f9](https://github.com/gazebosim/gz-common/pull/844/commits/f7cc8f98cbb237a565be0c1fce267543c693b0d1)

graphics/src/AssimpLoader.cc
```diff
@@ -583,8 +583,10 @@ MaterialPtr AssimpLoader::Implementation::CreateMaterial(
       }
     }
   }
-  ret = assimpMat->GetTexture(aiTextureType_NORMALS, 0, &texturePath);
-  if (ret == AI_SUCCESS)
+  if (assimpMat->GetTexture(
+          aiTextureType_NORMALS, 0, &texturePath) == AI_SUCCESS ||
+      assimpMat->GetTexture(
+          aiTextureType_HEIGHT, 0, &texturePath) == AI_SUCCESS)
```
luca-della-vedova, line 591: Let's add a comment here to say why we are doing this and why is it safe (from what I understand because this is what assimp does for Obj)

graphics/src/AssimpLoader.cc
```diff
@@ -600,6 +602,16 @@ MaterialPtr AssimpLoader::Implementation::CreateMaterial(
...
   }
+
+  ret = assimpMat->GetTexture(aiTextureType_SHININESS, 0, &texturePath);
+  if (ret == AI_SUCCESS)
+  {
+    std::string textureKey = this->FullTextureKey(texturePath.C_Str());
+    auto [texName, texData] = this->LoadTexture(
+      _scene, texturePath, this->GenerateTextureName(textureKey, "Shininess"));
+    pbr.SetSpecularMap(texName);
+  }
```
luca-della-vedova: I'm a bit conflicted here, it seems assimp defines a `aiTextureType_SPECULAR` as well, what makes it even more confusing is that it in obj it has both a [specularTexture](https://github.com/assimp/assimp/blob/ba8f4f456446e79594a6131c11674a559a7e07b1/code/AssetLib/Obj/ObjFileMtlImporter.cpp#L60) and a [specularityTexture](https://github.com/assimp/assimp/blob/ba8f4f456446e79594a6131c11674a559a7e07b1/code/AssetLib/Obj/ObjFileMtlImporter.cpp#L71).
Did you encounter any of these? I wonder if we should have an approach similar to the above where we parse both fields to try to get the specular map. Now of course the issue is that one of the if statements would end up being dead code if we don't actually ever have an asset with `aiTextureType_SPECULAR`.

Can you give a _quick_ look on whether it's straightforward to export a cube with the specular workflow instead of the metallic workflow as shown in the [khronos spec](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_specular/README.md)?
_quick_ look because it seems we don't really use specular maps much so if it's too much trouble we can just accept the lower coverage imho.

luca-della-vedova (approved):
One minor comment on documenting our behavior and one on expanding an if statement and _perhaps_ expanding coverage, with emphasis on the _perhaps_, non blocking for merging

graphics/src/AssimpLoader.cc
cheriehu: Yep ok. The assimp OBJ importer maps `map_bump` and `bump` to `aiTextureType_HEIGHT`. And `map_Kn` and `norm` to `aiTextureType_NORMALS`. Their exporter treats both `aiTextureType_HEIGHT` and `aiTextureType_NORMALS` [the same way (?)](https://github.com/assimp/assimp/blob/ba8f4f456446e79594a6131c11674a559a7e07b1/code/AssetLib/Obj/ObjExporter.cpp#L253-L257). 

[64ea134](https://github.com/gazebosim/gz-common/pull/844/commits/64ea1343182c89c7cc7f778ab58023985fbb4cca)

graphics/src/AssimpLoader.cc
cheriehu: > I'm a bit conflicted here, it seems assimp defines a `aiTextureType_SPECULAR` as well, what makes it even more confusing is that it in obj it has both a [specularTexture](https://github.com/assimp/assimp/blob/ba8f4f456446e79594a6131c11674a559a7e07b1/code/AssetLib/Obj/ObjFileMtlImporter.cpp#L60) and a [specularityTexture](https://github.com/assimp/assimp/blob/ba8f4f456446e79594a6131c11674a559a7e07b1/code/AssetLib/Obj/ObjFileMtlImporter.cpp#L71). Did you encounter any of these? I wonder if we should have an approach similar to the above where we parse both fields to try to get the specular map. Now of course the issue is that one of the if statements would end up being dead code if we don't actually ever have an asset with `aiTextureType_SPECULAR`.

Ah yea seems like `map_Ks` maps to `SpecularTexture` [which maps to `aiTextureType_SPECULAR`](https://github.com/assimp/assimp/blob/ba8f4f456446e79594a6131c11674a559a7e07b1/code/AssetLib/Obj/ObjFileImporter.cpp#L667-L674) and `map_Ns` maps to `SpecularityTexture` [which maps to `aiTextureType_SHININESS`](https://github.com/assimp/assimp/blob/ba8f4f456446e79594a6131c11674a559a7e07b1/code/AssetLib/Obj/ObjFileImporter.cpp#L728-L735) in assimp. We have `map_Ns` in the blender_pbr.mtl file which was set to be the RoughnessMap in the OBJLoader.

[Looking at tiny_obj_loader.h](https://github.com/gazebosim/gz-common/blob/4d7be5ad8c7df54debb72aee1f947f5b91df1fa3/graphics/src/tiny_obj_loader.h#L182-L183), it seems like `map_Ks` is the real specular map and `map_Ns` is supposed to be the specular highlight (??)

> Can you give a _quick_ look on whether it's straightforward to export a cube with the specular workflow instead of the metallic workflow as shown in the [khronos spec](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_specular/README.md)?

I couldn’t find a way to export with the specular workflow from blender (think it’s not supported yet). Managed to modify the PurpleCube.gltf and get assimp to recognize a texture as `aiTextureType_SPECULAR` (using the `KHR_materials_pbrSpecularGlossiness` extension) with help from Gemini.

I do think that assigning shininess to the specular map is probably not right. We can either remove that block or change it to:
```suggestion
  ret = assimpMat->GetTexture(aiTextureType_SPECULAR, 0, &texturePath);
  if (ret == AI_SUCCESS)
  {
    std::string textureKey = this->FullTextureKey(texturePath.C_Str());
    auto [texName, texData] = this->LoadTexture(
      _scene, texturePath, this->GenerateTextureName(textureKey, "Specular"));
    pbr.SetSpecularMap(texName);
  }
```

luca-della-vedova: This sounds reasonable, but I thought the `SHININESS` parsing was for the blender OBJ PBR, wasn't it?

cheriehu: Yea the field that the roughness map was shoved into is actually read as shininess in assimp. I think either way the behaviour will change? Either we recognize what's supposed to be the roughness map as shininess and then shove it into specularmap (which might not be what they expect) or I guess with this change it'll just not do anything with the  `map_Ns` field -- similar to what's happening with the `refl` field now.

graphics/src/MeshManager_TEST.cc
```diff
@@ -1010,6 +1108,41 @@ TEST_F(MeshManager, LoadGlTF2BoxWithJPEGTexture)
...
+
+  // Test that SpecularMap has data
+  auto materialId = mesh->SubMeshByIndex(0).lock()->GetMaterialIndex();
+  ASSERT_TRUE(materialId.has_value());
+  auto material = mesh->MaterialByIndex(materialId.value());
+  ASSERT_NE(material, nullptr);
+  auto pbr = material->PbrMaterial();
+  ASSERT_NE(pbr, nullptr);
+  EXPECT_NE(pbr->SpecularMap(), "");
+  mgr->RemoveAll();
```
luca-della-vedova: Nit, set the expectation to the filename itself instead of it not being an empty string?

More for a followup but I wonder if we can get rid of the `RemoveAll()` call in every test and just have it as part of the "teardown" for the test, since it's a fair bit of repeated code and prone to error if a user forgets to add it.

cheriehu: > Nit, set the expectation to the filename itself instead of it not being an empty string?

ah yes good point [c25e1b1](https://github.com/gazebosim/gz-common/pull/844/commits/c25e1b1066c62b751b4b339d501876979a2ce1fc)

> More for a followup but I wonder if we can get rid of the `RemoveAll()` call in every test and just have it as part of the "teardown" for the test, since it's a fair bit of repeated code and prone to error if a user forgets to add it.

removed in [a30db6d](https://github.com/gazebosim/gz-common/pull/844/commits/a30db6d4b41122f565acad22756a5fe144cc3311). confirmed from debug statements that teardown is called after every test.

luca-della-vedova (approved):
Thanks for iterating!

cheriehu:
thanks for the reviews 🙏

luca-della-vedova:

PR #543, opened 2023-10-13
Fix glTF / glb root node transform

iche033:
# 🦟 Bug fix

## Summary

There is a long standing issue in assimp on models having incorrect rotations: https://github.com/assimp/assimp/issues/849.

Currently, our assimp loader drops the root node transform (setting it to identity) to work around the issue. However, this caused other assets's rotation to be appear incorrect and inconsistent with other rendering / modeling tools. This PR adapts a another workaround posted in https://github.com/assimp/assimp/issues/849 to fix the root node transform for glTF / glb meshes.

Update as of [40c7bbb](https://github.com/gazebosim/gz-common/pull/543/commits/40c7bbbb0b3231f03888a6413e25a500511cc12c) - Changes should how only affect glTF / glb and kept the same behavior for other file formats like dae.

Here are tests with a couple of glb meshes in gazebo and other tools. The camera in Gazebo is set up to have the same orientation as the these tools (Y up, facing -X)

[WaterBottle](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/WaterBottle)

Gazebo Before (ignore the black seam at the front, it's clipped by camera's near clip plane):
WaterBottle_before

Gazebo After (pay attention to logo on the bottle):
WaterBottle_after

three.js editor:

WaterBottle_threejs_editor

godot:
WaterBottle_godot

[BarramundiFish](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BarramundiFish):

Gazebo Before:
BarramundiFish_before

Gazebo After:
BarramundiFish_after

three.js editor:
BarramundiFish_threejs_editor

godot:
BarramundiFish_godot

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #845, opened 2026-07-10
Merge remaining Collada/Assimp Loader tests to MeshManager_TEST

cheriehu:
# 🎉 New feature

## Summary
Combine remaining tests into MeshManager_TEST. Adds if/else block for tests with differing expectations.

This reduces duplicated tests between AssimpLoader and the other custom loaders. 
The test will run for both `GZ_MESH_FORCE_ASSIMP=true` and `GZ_MESH_FORCE_ASSIMP=false`. Assimp and the other mesh loaders will be used by MeshManager depending on what the environment variable is. 

#### 1. Change the scope of #ifndef _WIN32
The `#ifndef _WIN32` originally added in https://github.com/gazebosim/gz-common/commit/0b7709b86594f5f2c9c781d4425fb2a53c5ca97 to disable the MeshManager polyline tests for Windows. I changed the scope of this back to just the polyline tests.

#### 2. Merge remaining ColladaLoader and AssimpLoader tests into MeshManager_TEST
I'm addressing differing expectations between tests by using `if / else` blocks. Most of the differing expectations have to do with the different `VertexCount`/`TexCoordCount`/`NormalCount`/`IndexCount` between AssimpLoader and custom loaders. 

Additionally, there are also tests that have an expectation of specific error messages being printed to the terminal. These messages are not aligned between loaders. Tests of malformed meshes are also handled differently between loaders.

~Something that might be worth noting is that for the `LoadLines` test, ColladaLoader sets the `SubMeshPrimitiveType` to `common::SubMesh::LINES`, while AssimpLoader sets it to `common::SubMesh::TRIANGLES`.~ Fixed in this PR.

#### 3. Fix box_multiple_inst_controllers.dae test mesh
I believe this mesh was broken before (would not open in Blender as well)

## Backport Policy

- [X] This is safe to backport to the following versions:
    - [X] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] This **should not** be backported
- [ ] I am not sure
- [ ] Other (fill in yourself)

## Test it

1. After building workspace
2. `cd build/gz-common`
3. `ctest -R "MeshManager" --output-on-failure`

## Checklist
- [X] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the feature
- [X] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
[...]

PR #858, opened 2026-07-28
Minor AssimpLoader changes

cheriehu:
# 🦟 Bug fix

More minor fixes/changes to AssimpLoader (from #844)

## Summary

1. Remove ifdefs 
2. Do not create a material if it was just a default by assimp
Assimp creates a material for every mesh, but some meshes do not have any material information. So tests where there is an expectation for 0 materials will fail. I added logic to check if the name of the assimp material has the default AI_DEFAULT_MATERIAL_NAME name and to skip creating a material if it is to restore the old behavior.
3. Fix rotation for gltf
Referring to the images in #543 where this logic was introduced, the [BarramundiFish](https://github.com/KhronosGroup/glTF-Sample-Models/tree/main/2.0/BarramundiFish) is supposed to face in the positive z direction/left, which it does after this fix. 

Before fix
image

After fix
image

## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [ ] Added tests

Assisted-by: Gemini 3.1 Pro

### DIFF
--- a/graphics/src/AssimpLoader.cc
+++ b/graphics/src/AssimpLoader.cc
@@ -35,9 +35,7 @@
 #include "gz/common/SystemPaths.hh"
 #include "gz/common/Util.hh"
 
-#ifndef GZ_ASSIMP_PRE_5_2_0
-  #include <assimp/GltfMaterial.h>    // GLTF specific material properties
-#endif
+#include <assimp/GltfMaterial.h>    // GLTF specific material properties
 #include <assimp/Importer.hpp>      // C++ importer interface
 #include <assimp/postprocess.h>     // Post processing flags
 #include <assimp/scene.h>           // Output data structure
@@ -407,7 +405,6 @@ MaterialPtr AssimpLoader::Implementation::CreateMaterial(
     mat->SetBlendFactors(opacity, 1.0 - opacity);
   }
 
-#ifndef GZ_ASSIMP_PRE_5_1_0
   // basic support for transmission - currently just overrides opacity
   // \todo(iche033) The transmission factor can be used with volume
   // material extension to simulate effects like refraction
@@ -419,7 +416,6 @@ MaterialPtr AssimpLoader::Implementation::CreateMaterial(
   {
     mat->SetTransparency(transmission);
   }
-#endif
 
   // TODO(luca) more than one texture, Gazebo assumes UV index 0
   Pbr pbr;
@@ -435,7 +431,6 @@ MaterialPtr AssimpLoader::Implementation::CreateMaterial(
     auto [texName, texData] = this->LoadTexture(
         _scene, texturePath, this->GenerateTextureName(textureKey, "Diffuse"));
     mat->SetTextureImage(texName, texData);
-#ifndef GZ_ASSIMP_PRE_5_2_0
     // Now set the alpha from texture, if enabled, only supported in GLTF
     aiString alphaMode;
     auto paramRet = assimpMat->Get(AI_MATKEY_GLTF_ALPHAMODE, alphaMode);
@@ -452,9 +447,7 @@ MaterialPtr AssimpLoader::Implementation::CreateMaterial(
         mat->SetAlphaFromTexture(true, alphaCutoff, twoSided);
       }
     }
-#endif
   }
-#ifndef GZ_ASSIMP_PRE_5_2_0
   // Edge case for GLTF, Metal and Rough texture are embedded in a
   // MetallicRoughness texture with metalness in B and roughness in G
   // Open, preprocess and split into metal and roughness map
@@ -583,7 +576,6 @@ MaterialPtr AssimpLoader::Implementation::CreateMaterial(
       }
     }
   }
-#endif
   ret = assimpMat->GetTexture(aiTextureType_NORMALS, 0, &texturePath);
   if (ret == AI_SUCCESS)
   {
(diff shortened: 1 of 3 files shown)


graphics/src/AssimpLoader.cc
```diff
@@ -852,14 +840,23 @@ Mesh *AssimpLoader::Load(const std::string &_filename)
...
   for (unsigned _matIdx = 0; _matIdx < scene->mNumMaterials; ++_matIdx)
   {
+    aiString matName;
+    if (scene->mNumMaterials == 1 &&
+      AI_SUCCESS == scene->mMaterials[_matIdx]->Get(AI_MATKEY_NAME, matName) &&
+      std::string(matName.C_Str()) == AI_DEFAULT_MATERIAL_NAME)
+    {
+      // If there's only 1 material and it's Assimp's default, skip adding it.
+      continue;
+    }
```
luca-della-vedova: This looks a tiny bit risky.
Specifically, we are only checking the name for the default check, the name seems to be `DefaultMaterial` [here](https://github.com/assimp/assimp/blob/a939f2479513a4d92060f2348909886b67d548aa/include/assimp/material.h#L60), it's not impossible to think that there is an asset out there with that name. (I would have been less worried if it was `ASSIMP_DEFAULT_MATERIAL` or something along those lines).
Also, it seems for obj (which I guess is what we usually test here) the library [assigns the default material name](https://github.com/assimp/assimp/blob/a939f2479513a4d92060f2348909886b67d548aa/code/AssetLib/Obj/ObjFileMtlImporter.cpp#L467) but continues on to parse.
I wonder if instead we can check that the material has some other default property, for example I can see that material has a default constructor and [a member](https://github.com/assimp/assimp/blob/a939f2479513a4d92060f2348909886b67d548aa/include/assimp/material.h#L972) that contain the number of properties the material has, I wonder if those can be used instead for a default check? I.e. a material with no properties and a default name is a default material?

In that case we should probably refactor into a separate dataPtr function, something like:

```
bool IsDefaultMaterial([...])
```

Then finally we could put this check in the `CreateMaterial` function and update it so it returns a `nullptr` if it is a default material instead?

luca-della-vedova (commented):
Not touched in this PR but I _think_ you can remove the `(void) specularDefine` since it was to silence a compile warning in an ifdef path.

graphics/src/AssimpLoader.cc
cheriehu: Ah ok, I did a quick print out of the value of `mNumProperties` for the materials created while running AssimpLoader_TEST.cc. It seems. like `mNumProperties`=2 when there isn't originally a material but assimp creates a material. 

So maybe we can check for `mNumProperties == 2` in addition to the name being equal to `ASSIMP_DEFAULT_MATERIAL`

According to Gemini, here is where the two default properties are added (https://github.com/assimp/assimp/blob/a939f2479513a4d92060f2348909886b67d548aa/code/Common/ScenePreprocessor.cpp#L74-L100)

graphics/src/AssimpLoader.cc
luca-della-vedova, line 88: Let's update the docs of this function with the new behavior (nullptr if default material)

graphics/src/AssimpLoader.cc
```diff
@@ -369,9 +371,12 @@ MaterialPtr AssimpLoader::Implementation::CreateMaterial(
   MaterialPtr mat = std::make_shared<Material>();
   aiColor4D color;
   bool specularDefine = false;
-  // gcc is complaining about this variable not being used.
-  (void) specularDefine;
   auto& assimpMat = _scene->mMaterials[_matIdx];
+  bool defaultMaterial = IsDefaultMaterial(assimpMat);
+  if (defaultMaterial)
```
luca-della-vedova: Nit do we need a new variable for this or can we just put the function call in the if statement?

graphics/src/AssimpLoader.cc
```diff
@@ -804,6 +801,23 @@ SubMesh AssimpLoader::Implementation::CreateSubMesh(
   return subMesh;
 }
 
+//////////////////////////////////////////////////
+bool AssimpLoader::Implementation::IsDefaultMaterial(
+    const aiMaterial* _assimpMat) const
+{
+  aiString matName;
+  aiColor3D clr;
+  _assimpMat->Get(AI_MATKEY_COLOR_DIFFUSE, clr);
```
luca-della-vedova: Did you mean to put a check for `clr`? I.e. whether the fetch was successful, whether the color was 0.6 0.6 0.6, it seems right now it is unused

cheriehu: Ah, that was a mistake, I'll remove this. I wasn't intending on checking the color. The default color for STL doesn't seem to be 0.6, 0.6, 0.6

luca-della-vedova (approved):
Thanks for the quick iteration!

luca-della-vedova:
@mergifyio backport gz-common5 gz-common6

luca-della-vedova:

cheriehu:
oh no I realised the default `mNumProperties` for STL is 4 for some reason, so the MaterialCount check fails for STL (there wasn't the test for LoadSTL in the AssimpLoader_TEST before)

we might need to either remove that check or go back to the original behaviour of just creating a material for every mesh for now?

luca-della-vedova:
I see! That's OK, this would have inevitably surfaced when migrating the STL test.
Maybe we can do that after the straightforward tests are migrated from #845? I could imagine a workflow where:

* Because of this issue, the straightforward migrations from #845 will _not_ include the `LoadSTL` test.
* We get the straightforward ones in first.
* We have a new PR with the STL material fix + add the LoadSTL test to `MeshManager` tests, so we can validate that it doesn't break anything and it passes the new test.

Does this make sense?

cheriehu:
Yep makes sense, thanks!

PR #862, opened 2026-07-29
Merge mesh loader tests with common expectations into MeshManager_TEST

cheriehu:

# 🎉 New feature

## Summary
Merge tests that have common expectations between Assimp and custom loaders into the MeshManager test. (subset of #845)

## Backport Policy

- [X] This is safe to backport to the following versions:
    - [X] Jetty
    - [X] Ionic
    - [X] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Test it

1. Build the gz workspace
2. `./build/gz-common/bin/UNIT_MeshManager_TEST` and verify all tests pass

## Tests that have yet to be ported
- all the ones with differing expectations removed in 5579b87f1da684e1c356b594c230061d105bf919
- ColladaLoader tests added in #830 

## Checklist
- [ ] Added a screen capture or video to the PR description that demonstrates the feature
- [ ] Added tests

Assisted-by: Gemini 3


luca-della-vedova (commented):
Any reason not to remove the original implementations as well?

cheriehu:
> Any reason not to remove the original implementations as well?

ah yea you're right that was the point of merging the tests

graphics/src/MeshManager_TEST.cc
```diff
@@ -27,10 +30,42 @@
...
 #ifndef _WIN32
-class MeshManager : public common::testing::AutoLogFixture { };
+class MeshManager : public common::testing::AutoLogFixture,
+                    public testing::WithParamInterface<bool> {
+  protected: void SetUp() override
+  {
+    common::testing::AutoLogFixture::SetUp();
+
+    forceAssimpEnv = this->GetParam();
+    if (forceAssimpEnv)
```
luca-della-vedova: Nit, should use `this->` to follow coding convention

graphics/src/MeshManager_TEST.cc
```diff
@@ -27,10 +30,42 @@
...
+  protected: bool forceAssimpEnv = false;
+};
+
+INSTANTIATE_TEST_SUITE_P(
+    ForceAssimpScenarios,
+    MeshManager,
+    testing::Bool());
 
 /////////////////////////////////////////////////
-TEST_F(MeshManager, CreateExtrudedPolyline)
```
luca-della-vedova, line 33: We shouldn't need to change these to be tested with both parameters unless they actually load a file, same for `CreateMesh`, `Remove`, and generally quite a few of the pre-existing tests here.
Since they don't load a file testing them with both combinations of the parameter won't really change anything

luca-della-vedova (commented):
A minor nit and a more substantial suggestion to revert changes to pre-existing `MeshManager` tests

luca-della-vedova:
@osrf-jenkins run tests

graphics/src/MeshManager_TEST.cc
```diff
@@ -458,4 +496,642 @@ TEST_F(MeshManager, CreateMesh)
...
+
+  const common::Mesh *mesh = mgr->Load(meshFilename);
+
+  EXPECT_TRUE(mesh != nullptr);
+  mgr->RemoveAll();
+}
+
+/////////////////////////////////////////////////
+// Open a non existing file
+TEST_P(MeshManagerLoad, NonExistingMesh)
```
luca-della-vedova: I just realized, we probably don't need to load glbs with both combinations of the parameter as well, since glb can only be loaded by the assimploader either way

graphics/src/MeshManager_TEST.cc
```diff
@@ -458,4 +496,642 @@ TEST_F(MeshManager, CreateMesh)
...
+                  124.0f / 255.0f);
+
+  // Bug in assimp 5.0.x that doesn't parse coordinate sets properly
+  // \todo(iche033) Lightmaps are disabled for glb meshes
+  // due to upstream bug
+  // EXPECT_EQ(pbr->LightMapTexCoordSet(), 1);
+
+  // \todo(iche033) Lightmaps are disabled for glb meshes
+  // due to upstream bug
+  // EXPECT_NE(pbr->LightMapData(), nullptr);
```
luca-della-vedova: Is this still an issue?

graphics/src/MeshManager_TEST.cc
cheriehu: Hm there was one time the test running twice on a glb file uncovered a bug for me while I was modifying AssimpLoader because the second time it ran it used a cached texture instead of loading one from scratch. Would it be worth running each of the test cases twice [(something like this?)](https://github.com/gazebosim/gz-sim/blob/32ac5467fa2bbb49c95eac468c7c61ae7abfed16/src/Server_TEST.cc#L1386) to check for this issue?

graphics/src/MeshManager_TEST.cc
cheriehu: ah doesn't seem to be an issue anymore

graphics/src/MeshManager_TEST.cc
luca-della-vedova: Interesting, other than issues with bringup / teardown (i.e. making sure that `MeshManager` has the meshes fully cleared, since it's a global static) do you have any idea how that issue might have happened?
I'm not in principle opposed to the idea, just would like to know what happened. If the only global static we have is `MeshManager` then running it twice would be pretty much just checking that our bringup / teardown logic is valid. If any of the other loaders or dependencies is also a global static I can see why we would need to repeat the test.

cheriehu: I think it came up because mgr->RemoveAll() didn’t clear the textures that were in AssimpLoader’s [`imageCache`](https://github.com/gazebosim/gz-common/blob/d7c3fe44134872bc9068bdb68461149ec0e1556d/graphics/src/AssimpLoader.cc#L63). There might be similar issues with caches in the other loaders.

Is that something we want to fix or is that a feature?

luca-della-vedova: Ah I see, I am not very sure what the best course of action would be here to be honest.
I suppose a few options could be:

* Make the `AssimpLoader` clear its caches when initializing a new mesh load, similar to what the ColladaLoader does: https://github.com/gazebosim/gz-common/blob/d7c3fe44134872bc9068bdb68461149ec0e1556d/graphics/src/ColladaLoader.cc#L570-L578. The only drawback I can think of is a potential loss of performance if meshes reference textures from other meshes, since those would previously be in the cache but now the cache would be cleared.
* Make the `RemoveAll` method also clear the `AssimpLoader` caches, possibly through a new virtual `Loader::Clear()` API, or destroying + recreating the loader. This would fix tests but doesn't sound too clean.
* Of course, do nothing.

I'm wondering how much the "textures referenced from other meshes" use case happens and whether we could go for option 1. My hunch also is that since the cache keeps a `shared_ptr` to an `Image`, it would prevent the image from being deallocated when the mesh is destroyed (its reference count would never reach 0 just because it's in a cache), so if we want to keep the cache and decide not to reset it we should consider downgrading it to a `weak_ptr` to avoid that.
Still, I don't want to block this so we can tackle this in a followup since it's somewhat of a pre-existing issue. In either case I think I would rather have a single unit test that isolates the issue by trying to load a mesh twice rather than repeating the whole test suite twice.

luca-della-vedova (approved):

graphics/src/MeshManager_TEST.cc
cheriehu: ok! makes sense to use a single test to test the issue

luca-della-vedova:

PR #873, opened 2026-08-10
Modify AssimpLoader behaviour for hierarchical nodes

cheriehu:
# 🦟 Bug fix

## Summary

The following changes help to align the expectations between AssimpLoader and ColladaLoader for the `LoadBoxWithHierarchicalNodes` test.

1. Change AssimpLoader to parse hierarchical nodes in the same order that ColladaLoader does it (children first then parent) 
2. Make AssimpLoader use the `name` field in the collada file for the submesh `Name()` by setting the `AI_CONFIG_IMPORT_COLLADA_USE_COLLADA_NAMES` flag. 
3. Assign a child node with no name the name of the first ancestor node that has a name, or if none exist, then `unnamed_submesh_XXXX`. 
4. Fix bug where ColladaLoader would get stuck in while loop if none of the node's ancestors have a name and added tests for this case.

## Testing it
1. Build gazebo workspace
2. `./build/gz-common/bin/UNIT_MeshManager_TEST` 

## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [X] I am not sure

## Checklist
- [X] Added tests

Assisted-by: Gemini 3.1 Pro


graphics/src/AssimpLoader.cc
```diff
@@ -201,6 +202,27 @@ static std::string ToString(const aiString& str)
...
+  if (_node->mMetaData)
+  {
+    aiString colladaId;
+    if (_node->mMetaData->Get(AI_METADATA_COLLADA_ID, colladaId))
+    {
+      // If we have a value for the node ID, update it
+      nodeID = ToString(colladaId);
+    }
+  }
+  return nodeID;
```
luca-della-vedova: This can be refactored to avoid the `nodeID` variable altogether and just have line 220 be a `return ToString(colladaId)`, and line 223 be a `return ToString(_node->mName)`.

graphics/src/ColladaLoader.cc
```diff
@@ -729,6 +729,7 @@ void ColladaLoader::Implementation::LoadNode(
         nodeName = name;
         break;
       }
+      parent = dynamic_cast<tinyxml2::XMLElement *>(parent->Parent());
```
luca-della-vedova, line 732: Wow was this an endless while loop in the previous implementation 🤯 ?

graphics/src/AssimpLoader.cc
```diff
@@ -248,12 +270,47 @@ void AssimpLoader::Implementation::RecursiveCreate(const aiScene* _scene,
...
   // Visit this node, add the submesh
   for (unsigned meshIdx = 0; meshIdx < _node->mNumMeshes; ++meshIdx)
   {
     auto assimpMeshIdx = _node->mMeshes[meshIdx];
     auto& assimpMesh = _scene->mMeshes[assimpMeshIdx];
     auto nodeName = ToString(_node->mName);
+
+    // if node had no name originally and was assigned a default by
+    // assimp, replace it with the name of first ancestor node that has a name
+    if (nodeName.find("$ColladaAutoName$") == 0)
```
luca-della-vedova, line 293: Can this (and following) `find` just be an equality comparison? It's a bit more readable and by right this would return true if the name was `$ColladaAutoName$<anythingelse`. Very obscure but still `==` is a bit more readable.

cheriehu: > this would return true if the name was `$ColladaAutoName$<anythingelse`

Assimp labels a node with no name with `$ColladaAutoName$_<count from 0>` so I think this should be expected?

https://github.com/assimp/assimp/blob/bb7140123a51466c7690dea8723a349816aaf732/code/AssetLib/Collada/ColladaLoader.cpp#L1807

graphics/src/ColladaLoader.cc
cheriehu: I distinctly remember that happening but when I tried commenting this line out and running the test again, I realised the current test doesn't actually get stuck in an infinite while loop. 

I updated the test in [d5afc2d](https://github.com/gazebosim/gz-common/pull/873/commits/d5afc2ddaece314d984a9123b70af9d42897148a) so now it'll get stuck if I remove this line.

graphics/src/AssimpLoader.cc
cheriehu: Thanks for the suggestion! ad1f9c6

graphics/src/AssimpLoader.cc
luca-della-vedova: Ahhhh yes thanks for clarifying

graphics/src/AssimpLoader.cc
```diff
@@ -248,12 +269,47 @@ void AssimpLoader::Implementation::RecursiveCreate(const aiScene* _scene,
...
+    {
+      const aiNode *parent = _node->mParent;
+      nodeName = "";
+      while (parent && parent->mParent)
```
luca-della-vedova: Would this while loop end prematurely? It seems it needs both the `parent` to be valid and `parent` to have a parent, so it wouldn't traverse all the way to the root, but only to the first child of a root node?

luca-della-vedova (commented):
One minor edge boundary condition in the while loop, otherwise LGTM

graphics/src/AssimpLoader.cc
cheriehu: Mm yes, I added the second check because if it was just `while (parent)`, the test case I added (where neither the node nor the parent node has a name) would fail with this:
```
/Users/cheriehu/rotary_gz11_ws/src/gz-common/graphics/src/MeshManager_TEST.cc:846: Failure
Expected equality of these values:
  "unnamed_submesh_0"
  mesh->SubMeshByIndex(5).lock()->Name()
    Which is: "Scene"
[  FAILED  ] ForceAssimpScenarios/MeshManagerLoad.LoadBoxWithHierarchicalNodes/1, where GetParam() = true (0 ms)
```

since technically the `Scene` is the ancestor of all the nodes. 

The ColladaLoader deals with this by [checking that the parent node is of type "node"](https://github.com/gazebosim/gz-common/blob/cc9bf4423e292edc4e7c4be26e256f850f8e79b0/graphics/src/ColladaLoader.cc#L699) but there isn't an equivalent way to check the type of a node in assimp as far as I can tell. 

An alternative is to check that the `parent != scene->mRootNode`?

luca-della-vedova: Interesting, yea the `parent != scene->mRootNode` is a bit clearer

luca-della-vedova (approved):

luca-della-vedova:

PR #874, opened 2026-08-12
Build STB with static linkage

luca-della-vedova:
# 🦟 Bug fix

Fixes duplicated symbols issue with STB when compiled into a statically linked library / binary.

## Summary

This especially happens when compiling with Bazel, STB has macros that can make it use `static` instead of `extern` to make it have internal linkage.

## Backport Policy
I think STB is only on main and Harmonic.
    - [x] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [ ] Added tests
- [x] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Assisted-by: Gemini


shameekganguly (approved):

iche033 (approved):

luca-della-vedova:
I had to go somewhat 180 on this so requesting a re-review!

The approach did de-duplicate the dependency but did _not_ fix duplicate symbol linking issues, since STB is still a header only library and is included "as is" in code that includes it.
When compiling a statically linked target (i.e. through bazel) there will still be duplicate symbols.
The solution seems to be instead to use the `_STATIC` STB macros to mark the functions as having static linkage. I changed the PR to _only_ do that and it seems to fix bazel builds.
The choice of whether to have our own STB or fetch one from Bazel is more of an architecture choice but doesn't actually affect this so I reverted it.

luca-della-vedova:
This however now results in unused function warnings that breaks CI on Ubuntu. This should be a spurious warning since we obviously don't use all the functions that STB has and compiling it with internal linkage will guarantee that no downstream users can see them, hence the compiler can flag them as unused.
I suppose I can just add a new macro to SuppressWarning as I did for switch with no default statement in https://github.com/gazebosim/gz-utils/pull/217 and suppress it?

graphics/src/Image.cc
```diff
@@ -15,7 +15,10 @@
  *
  */
 #include <algorithm>
+#define STB_IMAGE_STATIC
```
shameekganguly, line 20: nit: consider documenting here that the STB_XXX_STATIC macros are defined to force internal linkage and avoid downstream duplicate symbol linker issues.

shameekganguly (approved):
Thanks, this makes sense!

graphics/src/Image.cc
luca-della-vedova: Done in https://github.com/gazebosim/gz-common/pull/874/commits/3f9883db4dcb6e0d7bd0327afb505e59fd59ccde, note that we will need https://github.com/gazebosim/gz-utils/pull/226 merged to get CI green otherwise the suppression macro won't exist.

luca-della-vedova:

PR #891, opened 2026-08-21
Fix skeleton root node naming for AssimpLoader

cheriehu:
# 🦟 Bug fix

## Summary

This PR fixes the MergeBoxWithDoubleSkeleton test in AssimpLoader to make its behavior match the ColladaLoader behaviour. 

1. Previously, the skeleton hierarchy output by AssimpLoader had the `Scene` as the root. With 4867c071c1c9ece50a1cd662a8d05009bc895282, the loader will recursively traverse the tree to find the lowest common ancestor of all the bones and use that as the root node. I
2. When printing out the hierarchy of the skeleton in the `box_with_double_skeleton.dae` file, I realised that ColladaLoader uses `sid` as the skeleton node name, and if that's not available it falls back to `name`, then `id` ([here](https://github.com/gazebosim/gz-common/blob/252093cfe6da79018a6074a7b48a1630f72fb724/graphics/src/ColladaLoader.cc#L1383-L1389)). [5963946](https://github.com/gazebosim/gz-common/pull/891/commits/5963946ffdc49fff53e1ab144da6de08312afcd1) changes AssimpLoader to match this behaviour (only for skeleton nodes) using a `GetColladaSkeletonNodeName` function.
3. When there is a Collada file with multiple roots, ColladaLoader got stuck in a while loop. The fix is here 5912d9fbef9de332e6ee810db37648c5c10190c2. 
4. Merged MergeBoxWithDoubleSkeleton test into MeshManager_TEST and removed duplicates.
5. Also merged CheckNonRootDisplacement test into MeshManager_TEST. Removed the expectation for the root node to be a nullptr because with the new changes, the root node is not a nullptr anymore.

I verified that the animated collada meshes still work 

https://github.com/user-attachments/assets/569157c9-6d66-4b07-a03e-62be7b952424

### Edit: changes for GLB animations
[267a68f](https://github.com/gazebosim/gz-common/pull/891/commits/267a68f8532ccd26042aa9a4f1408046a8ff2c31) fixes transforms for GLB meshes with animations. 

Before fix:
 

https://github.com/user-attachments/assets/5d39d4ff-0831-4847-a893-b5f55e680792

After fix: 

https://github.com/user-attachments/assets/7d55f655-bace-43ce-a8ec-f60b69aeb516

## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [X] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [X] Added tests

Assisted-by: Gemini 3.1 Pro


graphics/src/AssimpLoader.cc
```diff
@@ -209,19 +209,48 @@ static std::string GetColladaNodeID(const aiNode* _node)
...
+    _node->mMetaData->Get(AI_METADATA_COLLADA_SID, colladaSid))
+  {
+    return ToString(colladaSid);
+  }
+
+  std::string nodeName = ToString(_node->mName);
+  if (nodeName.find("$ColladaAutoName$") != 0)
+  {
+    return nodeName;
+  }
```
luca-della-vedova: With this we are reaching three instances of `string.find("$ColladaAutoName")`, it's enough that we should refactor it into a separate function and reuse it.

graphics/src/AssimpLoader.cc
```diff
@@ -209,19 +209,48 @@ static std::string GetColladaNodeID(const aiNode* _node)
...
+  // ColladaLoader prioritizes `sid` field for skeleton node name,
+  // then `name`, then `id`. Make AssimpLoader match this behaviour
+  aiString colladaSid;
+  if (_node->mMetaData &&
+    _node->mMetaData->Get(AI_METADATA_COLLADA_SID, colladaSid))
+  {
+    return ToString(colladaSid);
+  }
+
+  std::string nodeName = ToString(_node->mName);
```
luca-della-vedova: Nit, `const` since we don't expect to change it

graphics/src/AssimpLoader.cc
```diff
@@ -323,14 +352,19 @@ void AssimpLoader::Implementation::RecursiveCreate(const aiScene* _scene,
       for (unsigned boneIdx = 0; boneIdx < assimpMesh->mNumBones; ++boneIdx)
       {
         auto& bone = assimpMesh->mBones[boneIdx];
-        auto boneNodeName = ToString(bone->mName);
+        auto node = bone->mNode;
+        if (!node)
+        {
+          node = _scene->mRootNode->FindNode(bone->mName);
+        }
+        auto boneNodeName = GetColladaSkeletonNodeName(node);
```
luca-della-vedova: Another `const` (and generally throughout, if you expect to not change a variable set it to `const`).
This code is not super `const` correct so no need to go out of scope and fix the whole thing but if we are editing lines might as well address them.

graphics/src/AssimpLoader.cc
```diff
@@ -998,15 +1035,50 @@ Mesh *AssimpLoader::Load(const std::string &_filename)
...
+          // this parent node is their LCA
+          else
+          {
+            lca = _node;
+            break;
+          }
+        }
+      }
+      return lca;
+    };
```
luca-della-vedova: Not a super big fan of this `std::function` approach / function name. Maybe we should just follow the convention in the remaining part of the code where we declare a function so we can also add `doxygen` for it to make it clearer. The abbreviation is also not explained (I suppose it's something about a `Common Ancestor`)? So I would suggest either document the abbreviation in the doxygen or expand the name when declaring this as a function.

graphics/src/AssimpLoader.cc
```diff
@@ -998,15 +1035,50 @@ Mesh *AssimpLoader::Load(const std::string &_filename)
   {
     std::unordered_set<std::string> boneNames;
     this->dataPtr->RecursiveStoreBoneNames(scene, rootNode, boneNames);
+    std::function<const aiNode*(const aiNode*)> findLCA =
+      [&](const aiNode* _node) -> const aiNode*
+    {
+      if (!_node) return nullptr;
+      if (boneNames.find(GetColladaSkeletonNodeName(_node)) != boneNames.end())
```
luca-della-vedova: This naming is slightly tricky, it implies that we are only looking for a `Collada` node name but this loader is multi-format so this code is expected to work regardless of whether we are loading Collada, gltf or anything else.
I understand how the function would actually work regardless of the format (the metadata lookups would fail in non collada format and it would just default to the node name) but it might not be super immediate to people that first see this function.
I think we should update this in a way that makes it clear that this loader works regardless of the format, but I don't have a good idea on how to do it on top of my head, a few imperfect / incomplete ideas, probably the right one is some sort of combination of these and / or something else:
* Rename the function so it's clear that it works for every format.
* Have a check for the extension before calling the function.
* Pass the file extension to the function and use it in the computation of the node name.
* Gate the function call by extension and only call it for collada.

graphics/src/AssimpLoader.cc
cheriehu: moved to a separate function here [d8698db](https://github.com/gazebosim/gz-common/pull/891/commits/d8698db255a1bdb6dc80b1cdc4d6e3c2099b6307). 

> The abbreviation is also not explained (I suppose it's something about a `Common Ancestor`)

yep it's supposed to be lowest common ancestor, didn't realize it wasn't explained anywhere

graphics/src/AssimpLoader.cc
cheriehu: Hmm yes there are a number of collada-specific behaviours we've been introducing. 

For this function, renaming it and then passing in an extension to the function might make sense? But it would mean we need to introduce a new `_extension` variable to the `RecursiveCreate`, `RecursiveSkeletonCreate`, `RecursiveStoreBoneNames` functions where this function is used (?). 

The `GetColladaNodeId` function and the new function to refactor `string.find("$ColladaAutoName”)` will also be collada-specific - we could add the extension variable to the both of them and remove the reference to Collada in the function name?
e.g. for the skeleton node name function
```cpp
static std::string GetSkeletonNodeName(const aiNode* _node, std::string &_extension)
{
  if (!_node)
    return "";

  if (_extension == "dae")
  {
    // Collada-specific logic
  }
  else
  {
    return ToString(_node->mName)
  }
}
```

There’s also the thing of using a parent/ancestor name for a node with no name field which also seems collada-specific and can also be gated by the file extension. (https://github.com/gazebosim/gz-common/blob/11cf5d0eb3aac4da627d38dac3fe51843af0776f/graphics/src/AssimpLoader.cc#L291-L313)

luca-della-vedova: Hum I don't say this with a lot of pride but it seems there is a member variable for the `dataPtr` that contains the filename, `currentMeshPath` so I guess if instead of making it a `static` function it was a member of the `dataPtr` it could access the filename (hence the extension)

cheriehu: ahh makes sense! and i guess we can also make finding the extension a function

https://github.com/gazebosim/gz-common/blob/11cf5d0eb3aac4da627d38dac3fe51843af0776f/graphics/src/AssimpLoader.cc#L961-L969

graphics/src/AssimpLoader.cc
```diff
@@ -202,26 +241,87 @@ static std::string ToString(const aiString& str)
...
+{
+  if (_extension == "dae")
+  {
+    return _name.find("$ColladaAutoName$") == 0;
+  }
+  return false;
+}
+
 //////////////////////////////////////////////////
 // Utility function to get Collada nodeID from assimp node object
```
luca-della-vedova: Can remove this comment since it's now incorrect and the documentation is in the function declaration

graphics/src/AssimpLoader.cc
```diff
@@ -202,26 +241,87 @@ static std::string ToString(const aiString& str)
...
-  if (_node->mMetaData)
+  if (_extension == "dae" && _node->mMetaData)
   {
     aiString colladaId;
     if (_node->mMetaData->Get(AI_METADATA_COLLADA_ID, colladaId))
     {
-      // If we have a value for the node ID, update it
       return ToString(colladaId);
     }
+    return ToString(_node->mName);
```
luca-della-vedova: This is not necessary right? We would just fall through to the return statement immediately below

graphics/src/AssimpLoader.cc
```diff
@@ -202,26 +241,87 @@ static std::string ToString(const aiString& str)
...
+
+    aiString colladaId;
+    if (_node->mMetaData &&
+        _node->mMetaData->Get(AI_METADATA_COLLADA_ID, colladaId))
+    {
+      return ToString(colladaId);
+    }
+  }
+
+  return nodeName;
```
luca-della-vedova: This looks exactly like what the `GetNodeID` function is doing, can we just call  that function here?

graphics/src/AssimpLoader.cc
```diff
@@ -202,26 +241,87 @@ static std::string ToString(const aiString& str)
...
+}
+
+//////////////////////////////////////////////////
+std::string AssimpLoader::Implementation::GetFileExtension() const
+{
+  std::string filename = this->currentMeshPath;
```
luca-della-vedova: Can this be a `const auto&` to avoid a copy?

graphics/src/AssimpLoader.cc
```diff
@@ -202,26 +241,87 @@ static std::string ToString(const aiString& str)
...
+  std::string extension;
+  std::size_t extIdx = filename.rfind(".");
```
luca-della-vedova: `const`, + apply to all variables that are assigned a value at creation time and never changed (i.e. in `LCA` function)

graphics/src/AssimpLoader.cc
```diff
@@ -998,15 +1138,23 @@ Mesh *AssimpLoader::Load(const std::string &_filename)
...
     std::unordered_set<std::string> boneNames;
     this->dataPtr->RecursiveStoreBoneNames(scene, rootNode, boneNames);
+
+    const aiNode* lcaNode = this->dataPtr->FindLowestCommonAncestor(rootNode,
+                                                                    boneNames);
+    const aiNode* skelRoot = lcaNode ? lcaNode : rootNode;
+
+    const std::string rootName = this->dataPtr->GetSkeletonNodeName(
+        skelRoot, extension);
+    const std::string rootID = this->dataPtr->GetNodeID(skelRoot, extension);
```
luca-della-vedova: I'm starting to wonder whether we should gate this whole skeleton creation section and only do it if the mesh has animations. It seems otherwise we might be doing a fair bit of work for no reason.
Although there might be a reason why there is no if statement at the beginning of this block

luca-della-vedova (approved):
Thanks for iterating! Looks good and nice find on the inverse bind transform issue!

luca-della-vedova:

Issue #363, opened 2022-06-01
Support for GLTF MetallicRoughness maps


luca-della-vedova:
## Desired behavior
The GLTF2 format uses [MetallicRoughness](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#metallic-roughness-material) maps for PBR, where Green is roughness and Blue is metalness. Common exporters (i.e. blender) also pack separate textures into a single one following the specification.
We should be able to accept MetallicRoughness maps in Gazebo.

## Alternatives considered
I can imagine two different approaches for this:
* Add MetallicRoughness texture support to `gz::common::Material`, this would be as trivial as adding another property and getters / setters, then it's up to the users of the library to know what to do with it and how to handle edge cases (i.e. if both Metalness map and MetallicRoughness map are set?) Also from my understanding OGRE2 doesn't currently support MetallicRoughness maps so we would need to unpack them manually at the `gz-rendering` level, or add support to OGRE2.
* Preprocess the MetallicRoughness and split into two textures at load time: This could work well within the gz-common scope since it does image processing already, the main issue would be how to save and use the textures downstream. We would need to either save them to a temporary file (to reuse the current Material Textures API that requires having a filename) or we would need to have APIs to change textures from filenames to in-memory blocks.

## Implementation suggestion
My instinct is to go for the most complex approach of preprocessing the textures and having APIs to allow users to set textures as binary blobs in memory rather than filenames, but this is quite a complex task that would need considerable API change. The undefined case in approach 1 is a bit worrying and, while temp files would work great, it feels a bit hacky.
That being said though, the first approach with OGRE2 support at the shader level does offer the possibility of being the most optimal from a performance point of view, since we wouldn't need to do any sort of preprocessing, so it's still tradeoffs and choices.
Open to suggestions!

luca-della-vedova:
Implemented in the assimp test branch (`luca/assimp_sandbox`). The second approach was used and the MetallicRoughness map was split into a metal and a roughness map to be used in the existing pipelines

PR #906, opened 2026-08-26
Do lazy texture loading for non-embedded textures in AssimpLoader

### LINKED ISSUE #839: AssimpLoader handles textures differently than Collada/OBJ mesh loaders
I was trying to profile the gz-sim-gui-client while loading the [sydney_regatta](https://app.gazebosim.org/OpenRobotics/fuel/models/sydney_regatta) model and noticed a difference in how AssimpLoader and ColladaLoader handle texture images.

AssimpLoader calls [this setTextureImage function](https://github.com/gazebosim/gz-common/blob/2fc551b2a100d250ebb75e562ecfd53885d6f283/graphics/src/Material.cc#L144), setting the texture data to an `Image` type variable directly whereas ColladaLoader and OBJLoader calls [this function](https://github.com/gazebosim/gz-common/blob/2fc551b2a100d250ebb75e562ecfd53885d6f283/graphics/src/Material.cc#L158-L159) that just sets the filepath to the texture. 

From these flamegraphs (on mac) it seems that the AssimpLoader spends longer to load the meshes (~950ms) than the ColladaLoader (~239ms) but most of the time is spent in gz::common::Image::Load() to load the texture. 

GZ_MESH_FORCE_ASSIMP=true
Image

GZ_MESH_FORCE_ASSIMP=false
Image

When benchmarking the `MeshManager::Load function` using a modified version of this benchmark (https://github.com/gazebosim/gz-common/blob/main/test/benchmark/MeshManager.cc), the AssimpLoader is still faster than the ColladaLoader for the sydney regatta world.
```
default collada
BM_MeshManager/sydney_regatta/1/iterations:100        232 ms          232 ms          100

export GZ_MESH_FORCE_ASSIMP=true
BM_MeshManager/sydney_regatta/1/iterations:100       56.4 ms         56.2 ms          100
```

Since we are transitioning to AssimpLoader being the default ([discourse post](https://discourse.openrobotics.org/t/gazebo-pmc-meeting-minutes-2026-06-15/55498)) for future releases, this difference in behaviour might be worth noting.

Credit to @luca-della-vedova for spotting the difference in function calls

cheriehu:
# 🦟 Bug fix

Fixes #839

## Summary

As described in #839, the AssimpLoader currently loads the full texture image in the loader itself, whereas the other loaders just set the texture of the image to the filepath, and let the renderer load the texture image.

This PR makes AssimpLoader match the behaviour of the existing loaders to do "lazy loading" (i.e. just setting the filepath) for non-embedded textures. 

## Visual testing
Tested with collada, obj, gltf with external texture, and glb meshes and textures still appear

different-meshes-lazy

## Benchmark 
Comparing the loading times for the sydney_regatta.dae (mesh with a lot of textures) file again, I can see that the time taken for loading the mesh drops significantly since the full images are no longer being loaded by AssimpLoader.
```
GZ_MESH_FORCE_ASSIMP=true on main
--------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
BM_MeshManager/sydney_regatta/1       1031 ms         1000 ms            1
```

```
GZ_MESH_FORCE_ASSIMP=true on this PR
--------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
BM_MeshManager/sydney_regatta/1       48.9 ms         48.8 ms           13
```

## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [X] I am not sure

## Checklist
- [X] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [ ] Added tests
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

### DIFF
--- a/graphics/src/AssimpLoader.cc
+++ b/graphics/src/AssimpLoader.cc
@@ -117,14 +117,15 @@ class AssimpLoader::Implementation
   /// \param[in] _texturePath the path where the texture is located
   /// \param[in] _textureName the name of the texture
   /// \param[in] _shouldCache Whether the image should be cached
+  /// \param[in] _loadFullTexture Whether to load the full texture into memory
   /// \return a pair containing the name of the texture and a pointer to the
   /// image data, if the texture was loaded in memory
   public: std::pair<std::string, ImagePtr>
           LoadTexture(const aiScene* _scene,
                       const aiString& _texturePath,
                       const std::string& _textureName,
-                      const bool _shouldCache = true,
-                      const bool _loadFullTexture = true) const;
+                      bool _shouldCache = true,
+                      bool _loadFullTexture = true) const;
 
   /// \brief Function to split a gltf metallicroughness map into
   /// a metalness and roughness map
@@ -847,8 +848,8 @@ std::pair<std::string, ImagePtr> AssimpLoader::Implementation::LoadTexture(
     const aiScene* _scene,
     const aiString& _texturePath,
     const std::string& _textureName,
-    const bool _shouldCache,
-    const bool _loadFullTexture) const
+    bool _shouldCache,
+    bool _loadFullTexture) const
 {
   std::pair<std::string, ImagePtr> ret;
   std::string textureKey = this->FullTextureKey(_texturePath.C_Str());
--- a/graphics/src/AssimpLoader.cc
+++ b/graphics/src/AssimpLoader.cc
@@ -728,23 +728,6 @@ std::pair<std::string, ImagePtr> AssimpLoader::Implementation::LoadTexture(
   }
   else
   {
-    // Load external texture from disk
-    if (common::exists(textureKey))
-    {
-      gzdbg << "Loading external texture [" << textureKey << "]" << std::endl;
-      // Textures are uploaded to the GPU as RGBA; decode straight to RGBA in a
-      // single pass to avoid a later channel conversion.
-      ret.second = std::make_shared<Image>();
-      ret.second->Load(textureKey, Image::PixelFormatType::RGBA_INT8);
-      if (_shouldCache)
-      {
-        this->imageCache[textureKey] = ret.second;
-      }
-    }
-    else
-    {
-      gzerr << "External texture [" << textureKey << "] not found" << std::endl;
-    }
     ret.first = ToString(_texturePath);
   }
   return ret;
--- a/graphics/src/AssimpLoader.cc
+++ b/graphics/src/AssimpLoader.cc
@@ -631,14 +631,12 @@ MaterialPtr AssimpLoader::Implementation::CreateMaterial(
   {
     std::string textureKey = this->FullTextureKey(texturePath.C_Str());
 
-    // Check if the texture is embedded or not
-    auto embeddedTexture = _scene->GetEmbeddedTexture(texturePath.C_Str());
     auto [texName, texData] = this->LoadTexture(
         _scene, texturePath, this->GenerateTextureName(textureKey, "Diffuse"),
         true, false);
     // If texture is not embedded, just set the texture
     // image to the path to the parent folder of the texture.
-    if (embeddedTexture)
+    if (texData)
       mat->SetTextureImage(texName, texData);
     else
       mat->SetTextureImage(texName, _path);
--- a/graphics/src/AssimpLoader.cc
+++ b/graphics/src/AssimpLoader.cc
@@ -444,47 +444,50 @@ void AssimpLoader::Implementation::RecursiveCreate(const aiScene* _scene,
     {
       // TODO(luca) merging skeletons here
       auto skeleton = _mesh->MeshSkeleton();
-      // TODO(luca) Append to existing skeleton if multiple submeshes?
-      skeleton->SetNumVertAttached(subMesh.VertexCount());
-      // Now add the bone weights
-      for (unsigned boneIdx = 0; boneIdx < assimpMesh->mNumBones; ++boneIdx)
+      if (skeleton)
       {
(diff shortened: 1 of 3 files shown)


luca-della-vedova (commented):
I was going to say "did you look at metallic roughness maps" since I think it's not possible to do lazy loading there until I realized that this PR only does diffuse maps (texture images).
For curiosity did you look into PBR maps at all? (This can also be done in a followup)

cheriehu:
> For curiosity did you look into PBR maps at all? (This can also be done in a followup)

Ohh do you mean doing lazy texture loading for PBR materials? I didn't look into that. My thought process for this change was just

1. Ctrl-F to see where `SetTextureImage` is called and change the input to a filepath for non-embedded textures
2. Delete the time-consuming step of loading full textures in the `LoadTexture` function for non-embedded textures

Will take a look at PBR!

luca-della-vedova:
Given PBR (especially with the metallic roughness map edge case) is a bit more involved, should we get this merged first to cover the 99% of the use cases of materials that only have a diffuse texture?

cheriehu:
Hmmm this current implementation will break the behaviour for meshes with external metallicroughness maps since none of the textures will be loaded (thanks for pointing that case out). Is that acceptable? 

I was considering adding a boolean input after `_shouldCache` in the `LoadTexture` function that's like `_loadFullTexture` that defaults to true and we can set to false if we're sure it's ok. But not sure if that'll be a bit confusing/if it's ok to change the LoadTexture function

luca-della-vedova:
Ah right I thought it was reverted with the last change but not really since `LoadTexture` just wouldn't load textures anymore.
Nope it's not acceptable :sweat_smile: found an old issue discussing [metallic roughness maps](https://github.com/gazebosim/gz-common/issues/363), but it actually seems metallic roughness maps are now supported in OGRE so we might not need to do the split anymore.
Regardless, changing the API of an internal implementation function is not the end of the world so it should be OK

cheriehu:
I added the additional boolean input for `_loadFullTexture` in [e610524](https://github.com/gazebosim/gz-common/pull/906/commits/e6105246617b9adff227212613259763c58a2fcc) so we can lazy-load diffuse textures. From some testing on PBR materials (e.g. from https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/CompareNormal), it feels like lazy loading for PBR materials doesn't work?

For example, for the CompareNormal model, it's supposed to look like this (no lazy loading):
normal works

But if I try to do lazy loading, it looks like this
normal doesnt work

luca-della-vedova:
I think the reason is that paths as returned by `assimp` are relative to the mesh file, while downstream libraries such as `gz-rendering` require a full path to access the image.
The overload for `SetTextureImage` takes both the texture path and the resource path and joins them to find the full texture path https://github.com/gazebosim/gz-common/blob/6d22a54c0dc9aaed41b1e9ee3b3ae2c181f388cf/graphics/src/Material.cc#L158-L181

By contrast, the PBR only sets the map and doesn't do any path joining https://github.com/gazebosim/gz-common/blob/6d22a54c0dc9aaed41b1e9ee3b3ae2c181f388cf/graphics/src/Pbr.cc#L235

If we want it to work for PBR we would need to do manual path joining before setting the map data, so downstream libraries will have a full path to access.

cheriehu:
> I think the reason is that paths as returned by `assimp` are relative to the mesh file, while downstream libraries such as `gz-rendering` require a full path to access the image.

Thank you, this was the issue! Updated both AssimpLoader and OBJLoader to set the full file path for PBR texture maps. 

Testing with a slightly modified version of https://app.gazebosim.org/iche033/fuel/models/Cube%20PBR%20OBJ, I can tell that there's a difference in texture before and after applying the change to the OBJLoader.
Before, cube is completely smooth:
image

After, there's some texture on the cube so light doesn't reflect like it does on a completely smooth surface:
image


Also, [ebd4bbc](https://github.com/gazebosim/gz-common/pull/906/commits/ebd4bbcbc661e821e11a7de663f57b517da6d1c1) adds a check for if a mesh has a skeleton before trying to access it. I think this issue surfaced because I added this if block in #891 

https://github.com/gazebosim/gz-common/blob/6d22a54c0dc9aaed41b1e9ee3b3ae2c181f388cf/graphics/src/AssimpLoader.cc#L1132-L1140

graphics/src/AssimpLoader.cc
```diff
@@ -624,9 +632,17 @@ MaterialPtr AssimpLoader::Implementation::CreateMaterial(
     std::string textureKey = this->FullTextureKey(texturePath.C_Str());
 
     // Check if the texture is embedded or not
+    auto embeddedTexture = _scene->GetEmbeddedTexture(texturePath.C_Str());
```
luca-della-vedova: Do we need this check or can we just check the texData pointer for `nullptr`?

luca-della-vedova (approved):
Just a minor nit but generally LGTM

graphics/src/AssimpLoader.cc
cheriehu: good point! [676575f](https://github.com/gazebosim/gz-common/pull/906/commits/676575f28438146a4f19c96ee4e43beba9101e20)

luca-della-vedova:
@osrf-jenkins run tests

luca-della-vedova:

